PR #1, opened 2015-11-22
Command line script

moyogo:
This adds a command line script `glyphs2ufo`.


n7s:
s/distutils.core/setuptools/ in setup.py?

anthrotype:
True, entry_points are setuptools only.

moyogo:
@n7s Thanks!

PR #9, opened 2016-01-20
Warn about unrecognized weights in get_weight_code.

dougfelt:
- Also support more weights, e.g. SemiLight, DemiLight, DemiBold,
  Regular, empty string (to handle cases where 'Italic' was removed
  leaving no weight-- defaults to Regular in this case).

### DIFF
--- a/Lib/glyphs2ufo/torf.py
+++ b/Lib/glyphs2ufo/torf.py
@@ -427,9 +427,8 @@ def get_weight_code(style_name):
         'Black': 900
     }.get(style_name, None)
     if not weight_code:
-      print('WARNING: Unrecognized style name "%s"' % style_name,
-            file=sys.stderr)
-      weight_code = 400
+        warn('Unrecognized style name "%s"' % style_name)
+        weight_code = 400
     return weight_code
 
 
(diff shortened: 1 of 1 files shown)


Lib/glyphs2ufo/torf.py
```diff
@@ -411,15 +412,25 @@ def build_postscript_name(family_name, style_name):
...
+        'DemiBold': 600,
         'SemiBold': 600,
         'Bold': 700,
         'ExtraBold': 800,
         'Black': 900
-    }.get(style_name, 400)
+    }.get(style_name, None)
+    if not weight_code:
+      print('WARNING: Unrecognized style name "%s"' % style_name,
+            file=sys.stderr)
```
jamesgk: Could you make printing to sys.stderr the behavior of warn() (defined at the bottom of torf.py), then use warn here?

Lib/glyphs2ufo/torf.py
```diff
@@ -411,15 +412,25 @@ def build_postscript_name(family_name, style_name):
...
         'Medium': 500,
+        'DemiBold': 600,
         'SemiBold': 600,
         'Bold': 700,
         'ExtraBold': 800,
         'Black': 900
-    }.get(style_name, 400)
+    }.get(style_name, None)
+    if not weight_code:
+      print('WARNING: Unrecognized style name "%s"' % style_name,
```
jamesgk: Indentation

jamesgk:
Thanks! LGTM

PR #13, opened 2016-02-04
Handle style attributes more carefully

jamesgk:


davelab6:
I'm not sure this is a good idea; could you explain more about your rationale? :)

jamesgk:
> I'm not sure this is a good idea; could you explain more about your rationale? :)

I'll defer to Doug on this.

Anyhow, I just realized this PR is actually incomplete because the styleMapStyleName (mapping to name table entry 2) might now contain the width (handled [here](https://github.com/googlei18n/glyphs2ufo/blob/0b8f585ffbc566c5f6085808f8812b70724bbe31/Lib/glyphs2ufo/torf.py#L304)). Fix incoming soon.

dougfelt:
James and I talked, and he is going to do whatever the other tools do.

For Noto, assuming this ends up in the font subfamily field in the name table, we'd put only 'Regular', 'Bold', 'Italic', and 'Bold Italic' in the subfamily, in fact would only put 'Bold' there if the family has only regular and bold weights.  We'll post-process the generated files to make it do what Noto does, if it doesn't already.

With Noto (with some exceptions for CJK), we break components of the font name ('look' e.g. sans/serif/naskh/nastaliq, script, script variant, mono, display, weight, width, slope)  into the preferred family and preferred subfamily.  Conceptually the family is the subset of Noto fonts with the same look and cmap, and the components that are the same across those fonts go into the preferred family, the rest go into the preferred subfamily.  We then base the remaining name table fields on this.  For the original family and subfamily, if there are only bold and regular weights, we put the weight into the original subfamily along with italic, otherwise we put all weights (including regular) into the family, leaving only 'Italic' and 'Regular' in the subfamily (this is how the Adobe Source Han Sans fonts handle the 7 weights-- each weight becomes its own family).  All other variations are put into the family, e.g. 'Condensed', 'Display', 'Mono' etc.  We also support the WWS fields.  For this we put components relating to weight, width, and slope into the subfamily, and push other components into the family (and if either is redundant with the preferred family/subfamily, we leave it empty).

jamesgk:
This is now ready for review.

> James and I talked, and he is going to do whatever the other tools do.

Well, I tried to do what Glyphs does, but it actually just throws everything in subfamily for TTFs and seems to drop a lot of style info for UFOs. So instead I went with the kind of straightforward / lazy approach here.

schriftgestalt:
I explained the problems with the style names in another thread. I hope I do understand the problem. I try to explain it a bit more:

Their are three different cases (as far as I can see) regarding font and style names. In all cases the preferred family name comes from the family name in the font. The style linking family name (name ID 1, win) needs to get everything form style name except Regular, Italic and Bold. 
Now, where to get the style name from:
- one master one instance: use the style name of the instance. 
- multiple master export instances: again use the style name from the instances. 
- multiple master export masters: two options:
  - just use the master name (weight+width+custom)
  - check if there is an instance matching the master exactly. Then use its style, otherwise fall back to option 1

For all cases where the style name of the instance it used, everything can be controlled by the user by adding custom parameters. You there shouldn't need any processing of (preferred) family and style name (and style mapping family and style is quite clear what has to go where).

davelab6:
Thanks for the details Doug!

> For Noto, assuming this ends up in the font subfamily field in the name table, we'd put only 'Regular', 'Bold', 'Italic', and 'Bold Italic' in the subfamily, in fact would only put 'Bold' there if the family has only regular and bold weights. We'll post-process the generated files to make it do what Noto does, if it doesn't already.

Yes, that sounds right to me. I've documented what I believe are correct values in this [Family Naming Google Docs Spreadsheet](https://docs.google.com/spreadsheets/d/1ckHigO7kRxbm9ZGVQwJ6QJG_HjV_l_IRWJ_xeWnTSBg/edit#gid=0), which is mostly based on research by @twardoch. 

> With Noto (with some exceptions for CJK), we break components of the font name ('look' e.g. sans/serif/naskh/nastaliq, script, script variant, mono, display, weight, width, slope) into the preferred family and preferred subfamily. 

By "preferred [sub]family" I guess you mean the name table items 16 and CFF.FamilyName, and name table item 17. However, I don't think this makes sense; I would expect these to be family names everywhere - eg, name table item 1.

> Conceptually the family is the subset of Noto fonts with the same look and cmap, and the components that are the same across those fonts go into the preferred family, the rest go into the preferred subfamily. 

I think this sounds right, but then, "Put width in style name instead of family name" doesn't match this, because a set of Noto fonts with the same category (sans) and cmap (thai) with the normal widths look different to the set of sans thai fonts with the condensed widths - so width should be in family name.

This is how Roboto and Open Sans are set up today, and this is required for using the fonts as web fonts; while CSS _does_ have the font-stretch property, not many browsers support it, and the Google Fonts API doesn't either.

> We then base the remaining name table fields on this. For the original family and subfamily, if there are only bold and regular weights, we put the weight into the original subfamily along with italic, 

Yes, sounds good

> otherwise we put all weights (including regular) into the family, leaving only 'Italic' and 'Regular' in the subfamily (this is how the Adobe Source Han Sans fonts handle the 7 weights-- each weight becomes its own family)

I think this is not good, because 'style linking' the Bold and Bold Italic (usWeightClass 700) is assumed by rich text environments - ie, that those have the same family name (name table item 1) as the regular and italic, with a subfamily of "Bold" or "Bold Italic", and some other metadata set in head table macStyle bit 0 and OS/2 table fsSelection bit5. Android in particular is very sensitive to those values being set correctly. 

If the Bold and Bold Italic are not style linked, rich text environments will autobold the Regular and Italic, not select a family named "$thisFamily Bold", and each environment uses its own autobold method, producing reflow across environments - such as Chrome vs Firefox vs MS Word. 

> All other variations are put into the family, e.g. 'Condensed', 'Display', 'Mono' etc. 

Yes :) 

> We also support the WWS fields. For this we put components relating to weight, width, and slope into the subfamily, and push other components into the family

I am familiar with the WWS name table items in so far as they exist in the standard, but I'm not really familiar with their wider context. Which rich text environments use them?

> (and if either is redundant with the preferred family/subfamily, we leave it empty).

Yes, that sounds good

Lib/glyphs2ufo/torf.py
```diff
@@ -42,6 +43,25 @@
...
+    'SemiLight': 350,
+    'DemiLight': 350,
+    '': 400,
+    'Regular': 400,
+    'Medium': 500,
+    'DemiBold': 600,
+    'SemiBold': 600,
+    'Bold': 700,
+    'ExtraBold': 800,
+    'Black': 900}
```
davelab6: In Google Fonts API, the Thin maps to CSS font-weight 100, and the fonts I manage with a "Thin" style also have a usWeightClass of 250

In Google Fonts API, the Light maps to CSS font-weight 300, and the fonts I manage with a "Light" style have a usWeightClass of 300. 

There is also an "ExtraLight" style, that maps to CSS font-weight 200, and usWeightClass 275.

SemiLight, DemiLight, DemiBold don't exist in the Fonts API. 

ExtraBold and Black are in the API and use the values here in CSS and usWeightClass.

dougfelt:
@davelab6 
> Yes, that sounds right to me. I've documented what I believe are correct values in this [Family Naming Google Docs Spreadsheet](https://docs.google.com/spreadsheets/d/1ckHigO7kRxbm9ZGVQwJ6QJG_HjV_l_IRWJ_xeWnTSBg/edit#gid=0), which is mostly based on research by @twardoch. 

Thanks for the link.

> With Noto (with some exceptions for CJK), we break components of the font name ('look' e.g. sans/serif/naskh/nastaliq, script, script variant, mono, display, weight, width, slope) into the preferred family and preferred subfamily. 

> By "preferred [sub]family" I guess you mean the name table items 16 and CFF.FamilyName, and name table item 17. However, I don't think this makes sense; I would expect these to be family names everywhere - eg, name table item 1.

Or 'typographic [sub]family', name table items 16/17.

There wouldn't be a different family name field if there weren't a need for a different name, no?  It seems to me that the whole point is, for the standard family/subfamily, anything that can't go into the subfamily has to go into the family, whether you're prefer to have it there or not.  The preferred/typographic family is where you put the family name as you'd really like it.

> Conceptually the family is the subset of Noto fonts with the same look and cmap, and the components that are the same across those fonts go into the preferred family, the rest go into the preferred subfamily. 

> I think this sounds right, but then, "Put width in style name instead of family name" doesn't match this, because a set of Noto fonts with the same category (sans) and cmap (thai) with the normal widths look different to the set of sans thai fonts with the condensed widths - so width should be in family name.

By 'look' I'm just picking a shorthand to refer to a typographic style category like 'sans' or 'serif', not to look more generally.  Everything affects the look, including weight and slant, after all.  I can just as accurately say "normal weights look different to the set of sans thai fonts with the regular weights."  I don't think this is a convincing counter argument.

> This is how Roboto and Open Sans are set up today, and this is required for using the fonts as web fonts; while CSS _does_ have the font-stretch property, not many browsers support it, and the Google Fonts API doesn't either.

So the Adobe Source Han Sans fonts can't be used as web fonts as they are?  Because they put the weight into the family name, always.

>> otherwise we put all weights (including regular) into the family, leaving only 'Italic' and 'Regular' in the subfamily (this is how the Adobe Source Han Sans fonts handle the 7 weights-- each weight becomes its own family)

> I think this is not good, because 'style linking' the Bold and Bold Italic (usWeightClass 700) is assumed by rich text environments - ie, that those have the same family name (name table item 1) as the regular and italic, with a subfamily of "Bold" or "Bold Italic", and some other metadata set in head table macStyle bit 0 and OS/2 table fsSelection bit5. Android in particular is very sensitive to those values being set correctly. 

We could make android, in particular, smarter about this. 

> If the Bold and Bold Italic are not style linked, rich text environments will autobold the Regular and Italic, not select a family named "$thisFamily Bold", and each environment uses its own autobold method, producing reflow across environments - such as Chrome vs Firefox vs MS Word. 

So, for these (broken) environments, it sounds like we need to put bold, and only bold, into the name table subfamily (field 2).

These environments are broken because they do not deal with weights other than bold and regular.  What, for example, happens when a user has text in a light weight and clicks the 'bold' button?  'Light' is in the family name, there is no other family with 'Light' in the family name that has 'Bold' in the subfamily, so they autobold the light weight?  They do this even when the preferred or wws family name is populated, and provides a family name grouping that can define a collection of fonts that do in fact provide the appropriate weight within the same family?  Or are you saying that these environments know about the preferred/wws family names, find 'Light' in those names and use them as the display name, but still use the name table 1 family to create family groups?  That still seems broken.

> > We also support the WWS fields. For this we put components relating to weight, width, and slope into the subfamily, and push other components into the family

> I am familiar with the WWS name table items in so far as they exist in the standard, but I'm not really familiar with their wider context. Which rich text environments use them?

I don't have extensive information about various environments, but I see no reason to not populate them since it's straightforward to do so.

jamesgk:
OK, it seems this discussion has outgrown the intent of the PR a bit. But I'd like to merge/close this ASAP, so what is the consensus on just putting width in the subfamily? What I like about it is that Glyphs does it, so I think glyphs2ufo should too. In terms of semantic correctness, I think it's an arbitrary decision (one could justify width as a family or style property).

But probably the most important concern is practical:

> > This is how Roboto and Open Sans are set up today, and this is required for using the fonts as web fonts; while CSS does have the font-stretch property, not many browsers support it, and the Google Fonts API doesn't either.

> So the Adobe Source Han Sans fonts can't be used as web fonts as they are? Because they put the weight into the family name, always.

Not sure what to do about this. Can we at least merge this PR for now (and post-process output as needed to work as webfonts)?

schriftgestalt:
> putting width in the subfamily?
This could easily be default as it can be overwritten with custom parameters. Or do I miss something?

davelab6:
@jamesgk Sorry my delay, what this code does and what Doug describe seem at odds to me. I'll write a full reply today.

dougfelt:
@jamesgk, sorry, thought I'd replied.  Do whatever you and Dave think is necessary for roboto.  Noto can do whatever's appropriate for it in some post-processing step.

twardoch:
The general practice within an "extended family" is this: 

Related fonts with usWeightClass 400 and 700 get the same nameID 1 (Styling Group Name) and the nameIDs 2 (Styling Link Names) "Regular" and "Bold" appropriately. For italics, it's "Italic" and "Bold Italic". For those fonts, the nameID 1 should be the same as the Typographic Family Name, which normally goes into nameID 16, but if 1 and 16 would be the same, 16 is not written. 

All other usWeightClass fonts get a different nameID 1, and for nameID 2, they get "Regular" or "Italic". 

The above requirements are because of Windows GDI. The menu naming for newer apps is done in nameID 16 and 17 (Typographic Family and Style Name). 

If the extended family differs in both Width and Weight, then some vendors put the Width at the end of 16, while only Weight and Slope go into 17. Other vendors some put Width into 17, where the older tradition was to put Width after Weight ("Black Condensed"), but some newer members put Width before Weight ("Cond Black"). There is no fixed rule or convention. The overall idea is that the subfamily (style) list should not be overloaded with too many styles. So if the family is huge (e.g. 5 widths x 7 weights) then most likely Width would be put into nameID 16. If there are only two widths, it may make sense to put Width into nameID 17. 

Not sure how this was done with Noto previously, but if we stick to the newer Google Fonts / CSS recommendations, I'd say that only Weight and Slope should go into nameID 17 (and that there should be no more than 9 different weights in such a Typographic Family), while all other distinctions (Width, Sans/Serif, primary script or charset, primary weight group identifiers for families with >9 weights) should go into nameID 16. In other words, there would be fonts with nameID 17 "Noto Sans UI" and other fonts with nameID 17 "Noto Sans UI Cond". 

On the other hand, this may get a bit complicated with GX variation fonts. In Mac OS X, the nameIDs specified in the 'fvar' table show up in the "styles" list. To have a GX font show up as multiple "families" in the menus in OS X would most likely require a TTC with separate 'name' and 'fvar' tables for each subfont.

twardoch:
@dougfelt BTW, the `name` table has **nothing** to do with web fonts! In CSS, the web developer makes up their own naming scheme. In EOT, there was an odd requirement that nameID 4 had to include nameID at the beginning, but with WOFF/WOFF2, I'm not aware of any such restrictions. You can put whatever you want into the `name` table, it's not read by the browser. The CSS `font-family` descriptor and the other descriptors within `@font-face` are what the browser picks. You can make up "font families" in CSS via `@font-face` where the fonts are actually from completely different families. 

Of course it's possible that the Google Fonts API works so that it generates the "font-family" and other descriptors in the CSS it serves from the `name` table contained in the font files, but that's purely an implementation decision of the Google Fonts API, not something that is inherent to CSS web fonts per se.

davelab6:
>>> With Noto (with some exceptions for CJK), we break components of the font name ('look' e.g. sans/serif/naskh/nastaliq, script, script variant, mono, display, weight, width, slope) into the preferred family and preferred subfamily.
>>
>> By "preferred [sub]family" I guess you mean the name table items 16 and CFF.FamilyName, and name table item 17. However, I don't think this makes sense; I would expect these to be family names everywhere - eg, name table item 1.
>
> Or 'typographic [sub]family', name table items 16/17.
>
> There wouldn't be a different family name field if there weren't a need for a different name, no? It seems to me that the whole point is, for the standard family/subfamily, anything that can't go into the subfamily has to go into the family, whether you're prefer to have it there or not. The preferred/typographic family is where you put the family name as you'd really like it.

The family/subfamily is constrained by the Windows GDI API (used by MS Word); it only allows for a family to have 4 subfamilies (Regular, Italic, Bold, Bold Italic.)

For Adobe CS, as far as I know - not tested - the preferred/typographic family/subfamily is unconstrained, and via name IDs 16 and 17 you can have like 1,000 subfamilies in a family with all kinds of names that are unrelated to other OpenType metadata (such as usWeightClass, usWidthClass, panose, etc.) 

For web fonts, the preferred/typographic family/subfamily is constrained by CSS, and further by the Google Fonts API:

CSSv3 allows for a family to have 9 subfamilies for `font-weight` (100 to 900) multipled by 9 subfamilies for `font-stretch` (1 to 9, aka width) multiplied by 3 subfamilies for `font-variant` (normal, italic, oblique) for a total of 243 possible subfamilies. 

Google Fonts API allows for a family to have 9 subfamilies for weight (mapping to CSS font-weight 100-900) multipled by 2 subfamilies for `font-variant` (normal, italic) for a total of 18 possible subfamilies. This reflects the real-world status of CSSv3 support, as sadly not all browsers support font-stretch or font-variant and haven't prioritised doing so at all, so it will be 3-5 years until they do so.

So the title of this PR, "Put width in style name instead of family name," gives me cause for concern, because web browsers can not easily use such a family; those with partial CSS3 support require width to be in family name, not style name.

@twardoch how do the 'monster' families like FF Good split fonts across the typography-family/subfamily divide?

>>> Conceptually the family is the subset of Noto fonts with the same look and cmap, and the components that are the same across those fonts go into the preferred family, the rest go into the preferred subfamily.
>>
>> I think this sounds right, but then, "Put width in style name instead of family name" doesn't match this, because a set of Noto fonts with the same category (sans) and cmap (thai) with the normal widths look different to the set of sans thai fonts with the condensed widths - so width should be in family name.
>
> By 'look' I'm just picking a shorthand to refer to a typographic style category like 'sans' or 'serif', not to look more generally. Everything affects the look, including weight and slant, after all. I can just as accurately say "normal weights look different to the set of sans thai fonts with the regular weights." I don't think this is a convincing counter argument.

Okay, ya got me :) 

I can agree that **conceptually** a set of fonts with the same cmap are natrually grouped together - but **practically** they ought to not have the same family name due to the real-world constraints of various platform APIs.

>> This is how Roboto and Open Sans are set up today, and this is required for using the fonts as web fonts; while CSS does have the font-stretch property, not many browsers support it, and the Google Fonts API doesn't either.
>
> So the Adobe Source Han Sans fonts can't be used as web fonts as they are? Because they put the weight into the family name, always.

I am not fully informed of what the Google Fonts API engineering team believes is required to onboard CJK fonts, but if this was not a CJK font, then I would require that the weight not be included in the family name for Regular, Italic, Bold and Bold Italic fonts, and other fonts not have their weight in the subfamily but only Regular and Italic; and to included the family and subfamily in the typography family and typographic subfamily; such that the fonts when installed behave normally in GDI (eg the bold and italic styling buttons on toolbars call on real fonts, not autobold/autoslant versions of Regular) and in OT-aware applications.

>>> otherwise we put all weights (including regular) into the family, leaving only 'Italic' and 'Regular' in the subfamily (this is how the Adobe Source Han Sans fonts handle the 7 weights-- each weight becomes its own family)
>>
>>I think this is not good, because 'style linking' the Bold and Bold Italic (usWeightClass 700) is assumed by rich text environments - ie, that those have the same family name (name table item 1) as the regular and italic, with a subfamily of "Bold" or "Bold Italic", and some other metadata set in head table macStyle bit 0 and OS/2 table fsSelection bit5. Android in particular is very sensitive to those values being set correctly.
>
>We could make android, in particular, smarter about this.

I don't see how; Android is working perfectly correctly, adhering to the specification, and I don't see that making platforms accept off-spec fonts is good.

>> If the Bold and Bold Italic are not style linked, rich text environments will autobold the Regular and Italic, not select a family named "$thisFamily Bold", and each environment uses its own autobold method, producing reflow across environments - such as Chrome vs Firefox vs MS Word.
>
>So, for these (broken) environments, it sounds like we need to put bold, and only bold, into the name table subfamily (field 2).

Yes, given that there are no italic styles.

>These environments are broken because they do not deal with weights other than bold and regular. What, for example, happens when a user has text in a light weight and clicks the 'bold' button? 'Light' is in the family name, there is no other family with 'Light' in the family name that has 'Bold' in the subfamily, so they autobold the light weight?

Yes. 

> They do this even when the preferred or wws family name is populated, and provides a family name grouping that can define a collection of fonts that do in fact provide the appropriate weight within the same family? 

They ignore the preferred or wws family name IDS, because they are only aware of the name IDs that were defined/specified at the time their API was defined/specified.

> Or are you saying that these environments know about the preferred/wws family names, find 'Light' in those names and use them as the display name, but still use the name table 1 family to create family groups? That still seems broken.

I'm not aware of any environments that do anything like that; if a text environment is aware of name IDs 16 and 17 it uses them as direct replacements for IDs 1 and 4 (which is why they are sometimes referred to as the 'preferred' names.)

>>> We also support the WWS fields. For this we put components relating to weight, width, and slope into the subfamily, and push other components into the family
>>
>> I am familiar with the WWS name table items in so far as they exist in the standard, but I'm not really familiar with their wider context. Which rich text environments use them?
>
> I don't have extensive information about various environments, but I see no reason to not populate them since it's straightforward to do so

Fair enough; I don't populate them because I don't know what the real-world side effects might be. 

> What I like about it is that Glyphs does it, so I think glyphs2ufo should too. 

I'm confused; given the comment by @schriftgestalt above, it seems Glyphs can do it either way. I would start by assuming that Georg has not considered the constraints imposed by web browsers, since his main user segment are Adobe app users.

However, perhaps dealing with that constraint ought to be done by passing an argument to fontmake, or including metadata in a `.glyphs` file or `.glyphsProject` file, or post-processing the SFNT files with fontTools.

dougfelt:
>  but if this was not a CJK font, then I would require that the weight not be included in the family name for Regular, Italic, Bold and Bold Italic fonts,

What does it being a CJK font have to do with it?  I don't follow.

> I don't see how; Android is working perfectly correctly, adhering to the specification, and I don't see that making platforms accept off-spec fonts is good.

Please point me to the specification.

jamesgk:
OK, I'm just going to leave width in the family name for now. But I still want to merge the other stuff in this pull request. Has this code been reviewed?

davelab6:
3013fa6 looks good to me

davelab6:
>> but if this was not a CJK font, then I would require that the weight not be included in the family name for Regular, Italic, Bold and Bold Italic fonts,
>
> What does it being a CJK font have to do with it? I don't follow.

I can't onboard CJK fonts to the Google Fonts API at the moment, because their filesize is too large. I don't fully understand the details, I believe the DK and Garret are working with Xiangye to plan for dealing with that and whatever else is keeping this off my plate :)

>> I don't see how; Android is working perfectly correctly, adhering to the specification, and I don't see that making platforms accept off-spec fonts is good.
>
> Please point me to the specification.

https://www.microsoft.com/typography/otspec/head.htm has the macStyle specification. 

https://www.microsoft.com/typography/otspec/os2.htm#fss has the OS/2 fsSelection specfication.

jamesgk:
I'm merging this, but feel free to continue the discussion.

twardoch:
FontFont changed their opinion on this. One of Dziedzic’s superfamilies put all widths and weights under one name ID 16 (family) while another put the width into family and only weight and slope was in subfamily. Clan was done one way, Good the other way, but I’m not sure which was which. But I think FontFont’s more recent practice was to put width into family. Adobe on the other hand tends to have just one family and all other distinguishers (width, weight, optical size) go to subfamily. 

I’m more inclined to put width into family. If we ever release Lato Condensed, it’ll have “Lato Cond” as family, and subfamily will only contain weight and slope, i.e. there will be no more than 18 subfamilies (9 weights plus italics) in each family. 

The main reason for this is that while most desktop apps sort families alphabetically, different desktop apps sort subfamilies differently — some alphabetically, some by usWeightClass first, some by usWidthClass first. It’s super confusing for the end-users and it keeps changing. It’s very tedious to locate the right font in the menu for Adobe-style families that can have 40-50 subfamilies, because you never know how the app will sort it.

twardoch:
Also this has to do with typographic practice: within one line of text or sentence or paragraph, you're likely to mix fonts from one family that have different weight and slope, but you're unlikely to ever get good results when mixing different widths. 

In typographic practice, widths are usually kept consistent within the same level in text hierarchy. You may use different widths for different heading levels. 

Occasionally, you'll mix different widths in tables — but even then, in different columns. 

Switching weight and slope is usually a “within paragraph” operation. Switching width mostly happens for different paragraphs or hierarchy levels. So it’s actually confusing as hell to have different widths appear under one family.

davelab6:
what  ｄｏ  you mean, no  ｍｉｘｉｎｇ  of widths? :dancer: 

> If we ever release Lato Condensed, it’ll have `Lato Cond` as family

Why abbreviated, why not `Lato Condensed`?

dougfelt:
@davelab6, thanks for the reply.

> https://www.microsoft.com/typography/otspec/head.htm has the macStyle specification.
> https://www.microsoft.com/typography/otspec/os2.htm#fss has the OS/2 fsSelection specification.

I'm aware of these specs.  They're not pertinent to what we're discussing, which is what belongs in a preferred family/subfamily, what belongs in a wws family/subfamily, how font linking is done for purposes of choosing a font for bolding when 'bold' is applied to a font, and similar issues.  In particular I don't see how they indicate that using preferred or wws family to identify a suitable set of fonts to examine to find an appropriate font for bolding is in any way 'off-spec'.  This seems like a perfectly appropriate thing to do, regardless of whether various 90's APIs do it.

Adam's information about how applications sort subfamilies differently is interesting. We need to determine how much we care about this.  Given that our driving platforms are Android and the web, it's not clear to me.  If I understand Adam correctly the web (css at any rate) is completely indifferent to the name table data in a font, so the primary concern is Android.  And apparently there is no spec for sorting subfamilies for Android to violate, since applications seem to feel free to order them in UIs in any way they think fit.

My main driving reason for populating the preferred family/subfamily as I do is to surface the fact that across a family of Noto fonts users can expect the same text to be supported in the same way.  It will be possible in Noto to switch between a Display and non-display version of a font and expect that the same text will display in each-- same cmap, same conjuncts, same language support.  Whether your text can be displayed in a font is a pretty fundamental aspect of font selection.

twardoch:
@davelab6 Most of Łukasz’s fonts use 4-letter words as part of the family name, owing to Erik Spiekermann’s postulate that they’re the best. Erik’s fonts are Meta, Info, Unit, Real, and Łukasz’s are Clan, Good, More, Mach, Pitu, Eggo and Lato (with Ringo being an exception). I like to keep this if possible even for secondary identfiers in the family name — no technical reason, just taste. I prefer “Lato Cond” and “Lato Wide” to “Lato Condensed” and “Lato Extended” — especially that we might perhaps also make something like “Lato Slab Cond Pro” in future, and then having “Condensed” spelled out would be odd. :)

davelab6:
I suspect that your taste is trumped by library consistency, but I guess we
should cross the bridge when we come to it; it may be that browsers and the
API supports font-stretch by then, and then this is moot.

PR #17, opened 2016-02-26
Improve instance custom param handling

jamesgk:
This allows more custom parameters to be loaded into instance UFOs, using the same code that loads them for masters.

Part of https://github.com/googlei18n/fontmake/issues/20 in that some of these custom parameters are used for filtering glyph output.


dougfelt:
LGTM

PR #18, opened 2016-03-08
UFO3 kerning names and conflict removal

jamesgk:
Part of https://github.com/googlei18n/fontmake/issues/13


dougfelt:
LGTM, as far as I understand it.  But see question above about unused values in dicts.

PR #26, opened 2016-03-30
[interpolation] Designspace gen to separate func

### LINKED ISSUE #22: Export piece of code that writes .designspace file
For building variation fonts, I don't want to interpolate instances, but want the .designspace file.  The code at the beginning of glyphs2ufo.interpolate.interpolate() is short, but non-trivial.  Please move to a separate function.  Thanks.

jamesgk:
Should close #22


behdad:
Thanks James.

PR #29, opened 2016-04-01
Use defcon

jamesgk:
The biggest thing this does is replaces robofab with defcon, but it also tries to improve py23 compatibility and makes a couple of small fixes.

cc @behdad


dougfelt:
LGTM

behdad:
lgtm as well.

PR #31, opened 2016-04-11
Use instance family names

jamesgk:
Related to https://github.com/googlei18n/fontmake/issues/62. Names were not being applied to output properly before, and instances were getting mixed up.


behdad:
lgtm.  Thanks.

PR #34, opened 2016-04-13
Change more OpenType custom param keys

jamesgk:
A couple more custom parameter keys need to be changed so that they'll
be processed as OpenType table entries.

### DIFF
--- a/Lib/glyphs2ufo/builder.py
+++ b/Lib/glyphs2ufo/builder.py
@@ -308,8 +308,9 @@ def set_custom_params(ufo, parsed=None, data=None, misc_keys=(), non_info=()):
 
         opentype_attr_prefix_pairs = (
             ('hhea', 'Hhea'), ('description', 'NameDescription'),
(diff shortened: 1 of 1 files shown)


Lib/glyphs2ufo/builder.py
```diff
@@ -308,8 +308,9 @@ def set_custom_params(ufo, parsed=None, data=None, misc_keys=(), non_info=()):
 
         opentype_attr_prefix_pairs = (
             ('hhea', 'Hhea'), ('description', 'NameDescription'),
-            ('panose', 'OS2Panose'), ('typo', 'OS2Typo'), ('win', 'OS2Win'),
-            ('vendorID', 'OS2VendorID'), ('fsType', 'OS2Type'))
+            ('license', 'NameLicense'), ('panose', 'OS2Panose'),
```
dougfelt: We've been missing licenseUrl too.

jamesgk: Yep, these are just prefixes. Anything starting with "license" will be changed.

dougfelt:
LGTM

PR #35, opened 2016-04-14
Put width in style name instead of family name

jamesgk:
We decided we want this after all.

cc @behdad

### DIFF
--- a/Lib/glyphs2ufo/builder.py
+++ b/Lib/glyphs2ufo/builder.py
@@ -24,8 +24,7 @@ from defcon import Font
 
 __all__ = [
     'to_ufos', 'clear_data', 'set_redundant_data', 'set_custom_params',
-    'build_family_name', 'build_style_name', 'build_ufo_path',
-    'write_ufo', 'GLYPHS_PREFIX'
+    'build_style_name', 'build_ufo_path', 'write_ufo', 'GLYPHS_PREFIX'
 ]
 
 
@@ -196,8 +195,8 @@ def generate_base_fonts(data, italic):
     for master in data['fontMaster']:
         ufo = Font()
 
-        ufo.info.familyName = build_family_name(family_name, master, 'width')
-        ufo.info.styleName = build_style_name(master, 'weight', italic)
+        ufo.info.familyName = family_name
+        ufo.info.styleName = build_style_name(master, 'width', 'weight', italic)
 
         ufo.info.openTypeHeadCreated = date_created
         ufo.info.unitsPerEm = units_per_em
@@ -253,11 +252,11 @@ def set_redundant_data(ufo):
     family_name, style_name = ufo.info.familyName, ufo.info.styleName
 
     width_match = re.search('(%s)' % '|'.join(filter(None, WIDTH_CODES.keys())),
-                            family_name)
+                            style_name)
     width = width_match.group(0) if width_match else ''
     ufo.info.openTypeOS2WidthClass = WIDTH_CODES[width]
 
-    weight = style_name.replace('Italic', '').strip()
+    weight = style_name.replace('Italic', '').replace(width, '').strip()
     weight_code = WEIGHT_CODES.get(weight, None)
     if not weight_code:
         warn('Unrecognized weight "%s"' % weight)
@@ -273,10 +272,11 @@ def set_redundant_data(ufo):
         ufo.info.styleMapStyleName = style_name.lower()
         ufo.info.styleMapFamilyName = family_name
     else:
-        ufo.info.styleMapStyleName = (
-            'italic' if 'Italic' in style_name else 'regular')
-        ufo.info.styleMapFamilyName = (
-            '%s %s' % (family_name, weight)).strip()
+        ufo.info.styleMapStyleName = ' '.join(s for s in (
+            'bold' if weight == 'Bold' else '',
+            'italic' if 'Italic' in style_name else '') if s) or 'regular'
+        ufo.info.styleMapFamilyName = ' '.join(s for s in (
+            family_name, width, weight if weight != 'Bold' else '') if s)
     ufo.info.openTypeNamePreferredFamilyName = family_name
     ufo.info.openTypeNamePreferredSubfamilyName = style_name
 
@@ -434,21 +434,17 @@ def set_master_user_data(ufo, user_data):
         ufo.lib[GLYPHS_PREFIX + 'fontMaster.userData'] = user_data
 
 
-def build_family_name(base_family, data, width_key):
-    """Build family name from base name and width string in data."""
-
-    width = data.pop(width_key, '')
-    return ('%s %s' % (base_family, width)).strip()
-
-
-def build_style_name(data, weight_key, italic):
-    """Build style name from weight string in data and whether it's italic."""
+def build_style_name(data, width_key, weight_key, italic):
+    """Build style name from width and weight strings in data, and whether the
+    style is italic.
+    """
 
     italic = 'Italic' if italic else ''
+    width = data.pop(width_key, '')
     weight = data.pop(weight_key, 'Regular')
-    if italic and weight == 'Regular':
+    if (italic or width) and weight == 'Regular':
         weight = ''
-    return ('%s %s' % (weight, italic)).strip()
+    return ' '.join(s for s in (width, weight, italic) if s)
 
 
 def to_ufo_time(datetime_obj):
--- a/Lib/glyphs2ufo/builder_test.py
+++ b/Lib/glyphs2ufo/builder_test.py
@@ -19,7 +19,7 @@ import unittest
 from defcon import Font
 
 from glyphs2ufo import builder
-from glyphs2ufo.builder import set_redundant_data, build_family_name, build_style_name
+from glyphs2ufo.builder import set_redundant_data, build_style_name
 
 
 _warnings = []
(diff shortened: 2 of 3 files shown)


dougfelt:
LGTM

davelab6:
Please could you explain why?

jamesgk:
Deferring to Behdad this time

behdad:
> Please could you explain why?

Why not?  We want WWS-compliant style names.

behdad:
And please keep discussion points short, otherwise it will be TL;DR'ed.

davelab6:
> We want WWS-compliant style names.

I can not onboard families named this way into the Fonts API.

behdad:
> I can not onboard families named this way into the Fonts API.

Then give me a short explanation of why, and what you require.  Short is key.

I reviewed your naming spreadsheet and you seem to be doing the same thing.

behdad:
Also, if Google Fonts has different requirements, those can be addressed at onboarding time.  Moreover, webfonts completely ignore the name table, so I have no clue what you are talking about.

davelab6:
Maybe I misunderstand the title of this PR and the code. Please could you tell me an example list of font filenames produced after this commit?

behdad:
NotoSansDevanagariUI-SemiCondensedSemiBold.ttf

On Thu, Apr 14, 2016 at 3:05 PM, Dave Crossland <notifications@github.com>
wrote:

> Maybe I misunderstand the title of this PR and the code. Please could you
> tell me an example list of font filenames produced after this commit?
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly or view it on GitHub
> <https://github.com/googlei18n/glyphs2ufo/pull/35#issuecomment-210174539>
>



-- 
behdad
http://behdad.org/

davelab6:
> if Google Fonts has different requirements, those can be addressed at onboarding time

Okay, if you're happy with that, I have no objections; I think it is confusing for the majority of users who will get fonts from the www.google.com/fonts directory's download zips that will be different to "the same" fonts gotten from elsewhere; but this is not all that important and if you can live with it, so can I :) 

cc @twardoch and @schriftgestalt since they were involved in https://github.com/googlei18n/glyphs2ufo/pull/13 and might have more insights

behdad:
I still do not understand your concern.

On Thu, Apr 14, 2016 at 3:15 PM, Dave Crossland <notifications@github.com>
wrote:

> if Google Fonts has different requirements, those can be addressed at
> onboarding time
>
> Okay, if you're happy with that, I have no objections; I think it is
> confusing for the majority of users who will get fonts from the
> www.google.com/fonts directory's download zips that will be different to
> "the same" fonts gotten from elsewhere; but this is not all that important
> and if you can live with it, so can I :)
>
> cc @twardoch <https://github.com/twardoch> and @schriftgestalt
> <https://github.com/schriftgestalt> since they were involved in #13
> <https://github.com/googlei18n/glyphs2ufo/pull/13> and might have more
> insights
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly or view it on GitHub
> <https://github.com/googlei18n/glyphs2ufo/pull/35#issuecomment-210178021>
>



-- 
behdad
http://behdad.org/

davelab6:
My concern is that various APIs don't support width variation in their model, only weight (either reg/bold or 1 to 9 weights) and roman/italic. 

The principle of progressive enhancement suggests to me that the fonts should work in primitive models and then add metadata for richer models - especially since these primitive APIs have massive usage. 

Without progressive enhancement, then putting width in style name instead of family name requires people who want to use the fonts in the primitive APIs to fork the fonts. 

Maybe I'm misunderstanding and my concern is unwarranted. Perhaps you could explain what all the nameIDs will be for `NotoSansDevanagariUI-SemiCondensedSemiBold.ttf`?

behdad:
> My concern is that various APIs don't support width variation in their model, only weight (either reg/bold or 1 to 9 weights) and roman/italic.

Citation needed.

> The principle of progressive enhancement suggests to me that the fonts should work in primitive models and then add metadata for richer models - especially since these primitive APIs have massive usage.

I'm not interested in supporting 95, neither I am in ideologies.  Concrete examples please.

> Without progressive enhancement, then putting width in style name instead of family name requires people who want to use the fonts in the primitive APIs to fork the fonts.

That's fine.  We don't support everyone.

> Maybe I'm misunderstanding and my concern is unwarranted. Perhaps you could explain what all the nameIDs will be for NotoSansDevanagariUI-SemiCondensedSemiBold.ttf?

nameID1 = Noto Sans Devanagari UI SemiCondensed SemiBold
nameID2 = Regular
nameID16 = Noto Sans Devanagari UI
nameID17 = SemiCondensed SemiBold
os/2 fsSelection bit 8 = 1 (meaning nameID 21/22 (WWS) are equivalent to nameID 16/17)

davelab6:
The anti-tofu conspiracy continues!! ;)

> We don't support everyone.

Like I said, fair enough

behdad:
Good one :)

On Thu, Apr 14, 2016 at 5:46 PM, Dave Crossland <notifications@github.com>
wrote:

> The anti-tofu conspiracy continues!! ;)
>
> We don't support everyone.
>
> Like I said, fair enough
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly or view it on GitHub
> <https://github.com/googlei18n/glyphs2ufo/pull/35#issuecomment-210226305>
>



-- 
behdad
http://behdad.org/

impallari:
FWIW, my preference will be:
nameID16 = Noto Sans Devanagari UI SemiCondensed
nameID17 = SemiBold

See [Libre Clarendon](https://github.com/impallari/Libre-Clarendon) if you want to test another multi-axis glyph file.

In a perfect world, you will want something like:
nameID16 = Noto Sans Devanagari UI
nameID17a = SemiCondensed
nameID17b = SemiBold
Since those 2 attributes means different things

dougfelt:
This just pushes WWS naming into the preferred fields.  The two attributes
mean different things, as does italic, but they can all be simultaneously
in the same field and it's up to the client to sort out how they group.
See e.g. https://www.microsoft.com/typography/otspec/name.htm under "Name
Ids" field 22: Examples of name ID 22: “Semibold Italic”, “Bold Condensed”.

behdad:
> In a perfect world, you will want something like:
> nameID16 = Noto Sans Devanagari UI
> nameID17a = SemiCondensed
> nameID17b = SemiBold

What does this mean?  Not following.

twardoch:
I agree with adopting the WWS naming as per Behdad’s decision. I thought long about this. If the Google Fonts API has limitations, it should build their own naming per name.NID[16] plus OS/2.usWidthValue and OS/2.usWeightValue (and the italic field), and map these to the appropriate CSS properties.

davelab6:
Please explain your long thinking and why you agree :)

twardoch:
1. Having three separate naming systems (GDI, Typographic and WWS) is stupid. We know that we need the GDI one, so I think using WWS-compatible naming within the Typographic naming makes most sense.
2. If as many families as possible have WWS-compatible naming, and actively use the OS/2 Width and Weight (plus the fsStyle for italic), then font fallback works better. If you change fonts, or you need fallback for unsupported characters, WWS naming will give you sensible matches if all or most fonts adhere to the WWS-conpatible naming. 
3. CSS3 font-stretch is not supported in browsers, but should be. Principally, CSS font-stretch, font-weight and font-style are WWS-compatible, so clean mapping is possible. Right now, some families don't put Width into TSN (Typographic subfamily name) but many do, including Adobe's. 
4. With the GX Variations on the horizon (at some point, hopefully), it's even more logocal to use WWS. At least width and weight axes will be supported by many families. If you offload Width to TFN (Typographic family name) today, it'll be much harder to get future compatibility with GX. 
5. With or without GX, better font selection UIs can implement "+/–" types of navigation along width and weight axes, and can do so even when changing families. Even Google Fonts might at some choose to implement a better UI. 
6. I could go on but that'd make it TL;DR, so I'll stop. :)

twardoch:
Ps. 7. AFAIK, Behdad's model of using WWS-compatible naming in TFN+TSN does **not** break any existing desktop or OS APIs. In fact, this naming is perfectly compatible with, for example, Adobe's Myriad Pro family. It also does not break any webfont usage directly as the name table isn't used there at all. 

It may break the current "Google Fonts onboarder", but I think it's easier for Google Fonts to tweak their onboarder tool than to introduce cause compatibility problems to users and future responsive font technologies. 

As long as there are very few opensource  families that use WWS naming with Width in TSN, browser makers will be complacent about supporting CSS font-stretch. But when Noto-able families emerge, then finally browser makers may make the effort to add support for font-stretch.

davelab6:
CSS font-stretch support is only weak on mobile, according to http://caniuse.com/#feat=css-font-stretch

twardoch:
True, but my point is that you don't have to rely on font-stretch for webfonts. 

The user-facing naming and grouping that webfont hosting services like Google Fonts, MyFonts, Fonts.com or Adobe Typekit expose to users does not need to slavishly follow the TFN/TSN "name" table stuff. It can use the GDI naming or even completely custom naming. 

I know that webfont vendors like Google Fonts are keen to make its life easier (e.g. by suggesting to remake Work Sans to have 9 weights instead of 10 because that's what CSS font-weight allows), and to some extent, it's fine. But it shouldn't go too far. 

In particular, IMO, the webfont hosting services shouldn't cement the browser vendors' shortsighted laziness in implementing web standards. :) Yes, these services operate "today", but they can adapt. Yet naming in desktop fonts tends to have more longevity. Much longer than the length of this post. ;)

PR #36, opened 2016-04-14
[interpolation] Use name field from instance data

jamesgk:
This is instead of the weight/widthClass fields, which we used before
simply for convenience but are not used by Glyphs itself. We do still
want to split the width and weight into two separate words for the
style name, so this pulls out a shared style parsing method.


dougfelt:
LGTM

PR #38, opened 2016-04-20
Propagate anchors from components

### LINKED ISSUE #33: Arabic mark positioning
The file [NotoSansArabic-MM.glyphs](https://github.com/googlei18n/noto-source/blob/master/src/NotoSansArabic-MM.glyphs) from noto-source outputs the following TTF in glyphs:

[NotoSansArabic-Regular.ttf.zip](https://github.com/googlei18n/glyphs2ufo/files/215753/NotoSansArabic-Regular.ttf.zip)

With this TTF it is possible to see that mark U+064B (arabic fathatan) has a positioning rule with base U+FEC0 (arabic letter dad medial form):

```
$ hb-shape NotoSansArabic-Regular.ttf "ﻀً"
[uni064B=0@490,114+0|uniFEC0=0+949]
```

This is despite the fact that in Glyphs' UFO output, glyph U+FEC0 does not have any anchors:

[dad-ar.medi.glif.zip](https://github.com/googlei18n/glyphs2ufo/files/215755/dad-ar.medi.glif.zip)

@schriftgestalt, how is the mark positioning rule determined in this case? Does the output UFO contain the necessary information to produce a correct TTF?

jamesgk:
Should close #33 

cc @behdad

### DIFF
--- a/Lib/glyphs2ufo/builder.py
+++ b/Lib/glyphs2ufo/builder.py
@@ -21,6 +21,7 @@ import shutil
 import sys
 
 from defcon import Font
+from fontTools.misc.transform import Transform
 
 __all__ = [
     'to_ufos', 'clear_data', 'set_redundant_data', 'set_custom_params',
@@ -678,13 +679,8 @@ def propagate_anchors(ufo):
     def get_anchor(glyph, name):
         return next((a for a in glyph.anchors if a.name == name), None)
 
-    # defcon does not allow simply copying an anchor from one glyph to another
-    def append_anchor(glyph, anchor):
-        anchor_dict = {'x': anchor.x, 'y': anchor.y, 'name': anchor.name}
-        glyph.appendAnchor(glyph.anchorClass(anchorDict=anchor_dict))
-
     for parent in ufo:
-        added_here = set()
+        added_here = {}
 
         # don't propagate anchors for mark glyphs
         if any(a.name.startswith('_') for a in parent.anchors):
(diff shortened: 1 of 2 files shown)


Lib/glyphs2ufo/builder.py
```diff
@@ -672,6 +673,46 @@ def add_glyph_to_groups(kerning_groups, glyph_data):
...
+        added_here = {}
+
+        # don't propagate anchors for mark glyphs
+        if any(a.name.startswith('_') for a in parent.anchors):
+            continue
+
+        # try to get anchors from base (first) components
+        glyph = parent
+        transformation = Transform()
+        while glyph.components:
```
dougfelt: I don't see how this loop terminates.

jamesgk: When `glyph` is re-assigned, it's set to the existing `glyph`'s first component (and presumably we won't have an infinite chain of components, i.e. eventually we'll get to a glyph with no components).

dougfelt: Oh, duh, you're walking up the chain, glyph keeps changing.  Sorry for being obtuse.

jamesgk:
PTAL. Added one more commit in response to discussion at https://github.com/googlei18n/ufo2ft/pull/34

dougfelt:
LGTM

behdad:
Reading https://glyphsapp.com/tutorials/advanced-diacritics-multiple-anchors looks like underscore has a special meaning, not `_1` or `_2`.  @schriftgestalt can you please clarify where it is stored which alternative anchor a mark is attached to?  I think that's what we need to implement here.

schriftgestalt:
The _1, _2 anchors are usually used on ligatures to generate the liga2mark feature.

jamesgk:
> The _1, _2 anchors are usually used on ligatures to generate the liga2mark feature.

Maybe this is obvious to other people, but I still don't understand how one of these anchors is chosen when attaching a mark. To follow the example in https://github.com/googlei18n/glyphs2ufo/issues/33, we can do:

```
$ hb-shape NotoSansArabic-Regular.ttf "ﻼً"
[uni064B=0@-44,95+0|uniFEFC=0+599]
$ hb-shape NotoSansArabic-Regular.ttf "عقلًا"
[uni064B=2@260,257+0|uniFEFC=2+599|uniFED8=1+495|uniFECB=0+503]
```

U+FEFC (lam_alef-ar.fina) has anchors `top_1 {422, 714}` and `top_2 {118, 552}`, while U+064B (fathatan-ar) has `_top {162, 457}`. So it seems that in the first case the mark is attached to top_2 while in the second it's attached to top_1. Thus this PR is incomplete, but I don't personally know how to fix it.

schriftgestalt:
Are we speaking about the anchors in precomposed glyphs? Or about the mark feature?

the `lam_alef-ar.fina` use a `pos ligature` statement. You can check the feature code in the Temp folder in `~/Library/Application Support/Glyphs/Temp/FontName/features.fea`

behdad:
> Are we speaking about the anchors in precomposed glyphs? Or about the mark feature?

I think the question is, where do you store which of top_x, top_y, top_z a particular top-mark attaches to?

jamesgk:
Well it's now clear to me how to use the _1/2/3 anchors (I wasn't even aware of the `pos ligature` statement... thank you Georg!). I guess it's still an open question how other anchors with underscores in their names are used. Unless, maybe they're not used in mark features at all?

schriftgestalt:
Anchir names with undercores+numbers are not used for mark positioning other than for ligatures. And all anchors that have a character other than a-zA-Z0-9._- should be ignored for the mark features. They are then only used to align components.

PR #39, opened 2016-04-26
don't import defcon/mutatorMath in global scope

anthrotype:
glyphs2ufo should not require defcon or mutatorMath if a user just wants to parse a .glyphs file using `glyphs2ufo.glyphslib.load` or `loads` functions, but does not want to also dump it to ufo.

### DIFF
--- a/Lib/glyphs2ufo/builder.py
+++ b/Lib/glyphs2ufo/builder.py
@@ -20,8 +20,6 @@ import re
 import shutil
 import sys
 
-from defcon import Font
-
 __all__ = [
     'to_ufos', 'clear_data', 'set_redundant_data', 'set_custom_params',
     'build_style_name', 'build_ufo_path', 'write_ufo', 'GLYPHS_PREFIX'
@@ -178,6 +176,7 @@ def clear_data(data):
 
 def generate_base_fonts(data, italic):
     """Generate a list of UFOs with metadata loaded from .glyphs data."""
+    from defcon import Font
 
     date_created = to_ufo_time(data.pop('date'))
     family_name = data.pop('familyName')
--- a/Lib/glyphs2ufo/interpolation.py
+++ b/Lib/glyphs2ufo/interpolation.py
@@ -17,10 +17,6 @@ from __future__ import print_function, division, absolute_import
 
 import os
 
-from defcon import Font
-from mutatorMath.ufo import build
-from mutatorMath.ufo.document import DesignSpaceDocumentWriter
-
 from glyphs2ufo.builder import set_redundant_data, set_custom_params,\
     clear_data, build_style_name, write_ufo, build_ufo_path, GLYPHS_PREFIX
 
@@ -37,6 +33,8 @@ def interpolate(ufos, master_dir, out_dir, instance_data,
     """Create MutatorMath designspace and generate instances.
     Returns instance UFOs, or unused instance data if debug is True.
     """
+    from defcon import Font
+    from mutatorMath.ufo import build
 
     designspace_path, instance_files = build_designspace(
         ufos, master_dir, out_dir, instance_data, italic)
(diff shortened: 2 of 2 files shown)


behdad:
There's a slight penalty to local imports.  But I agree that's a cleaner approach when the penalty is negligible.

behdad:
lgtm

anthrotype:
It's mainly to keep dependencies to a minimum if one is only interested in the parser (e.g. for the Fontlab scripts to import/export glyphs files: https://github.com/schriftgestalt/Glyphs-Scripts/pull/25)

Until today, I was actually not aware of the existence of a ASCII-text property list format, which is what Glyphs.app uses and glyphs2ufo parses. Wikipedia says it come from the glorious NeXTSTEP days, before Apple made the more common XML-flavored plist format. It's a pity that he built-in plistlib module only supports XML or binary plist, but also this "old-style" plist Glyphs uses. Still, glyphs2ufo does a pretty good job at parsing.

behdad:
> It's mainly to keep dependencies to a minimum if one is only interested in the parser (e.g. for the Fontlab scripts to import/export glyphs files: schriftgestalt/Glyphs-Scripts#25)

Sure.

> Until today, I was actually not aware of the existence of a ASCII-text property list format, which is what Glyphs.app uses and glyphs2ufo parses. Wikipedia says it come from the glorious NeXTSTEP days, before Apple made the more common XML-flavored plist format. It's a pity that he built-in plistlib module only supports XML or binary plist, but also this "old-style" plist Glyphs uses. Still, glyphs2ufo does a pretty good job at parsing.

Yeah, initially I was extending plistlib to add the ASCII format and contribute it back to Python.  But the ASCII version does not support types and many other things.  Moreover, Glyphs's use of it does not really conform to the Apple format.  For example, Glyphs does not quote tokens starting with period.  Anyway, after some back and force with @schriftgestalt I decided that this is really a Glyphs format now, but a generic one.

schriftgestalt:
Just to make it clear. I choose the text-plist format because it saves quite a bit space. 
And the files are still parseable by Apples implementation (NSDictionary, Property List Editor).

dougfelt:
It would be good to know how strict the parsing is.  I was trying to
regenerate files we'd parsed and found some data that looks numeric is
quoted while other numeric is not.  Also some of the lists are indented
while others are not. Matching the input exactly looked like I'd need to
add special cases for particular sections, so I shelved it.  If parsing is
lenient w.r.t quoting and indentation then I might try again.  It's useful
for swatting files.

schriftgestalt:
Can you give some examples about the inconsistencies? I cleaned some up recently when working on the .glyphs file saving part.

behdad:
@dougfelt The major issue I found is that in the ASCII plist, whether something is quoted or not does NOT specify it's type.  Ie, integer vs string.

I think my remaining ask from @schriftgestalt was to always quote hex numbers, and preferably prefix them with 0x.  Currently what might look like a decimal number may in fact be a hex unicode reference, so no automatic type detection should happen.

dougfelt:
We have files in which under instances/customParameters
typoAscender/typoDescender/typoLineGap/winAscent/winDescent... etc are all
quoted.  However under fontMaster the ascender/capHeight/descender are all
unquoted.

glyphOrder, unicodeRange, panose values are indented, but other lists are
not.  These lists all seem to be dictionary values, so perhaps that's the
distinction?

I'm not sure what version was used to generate these files.

dougfelt:
Yes, it looks like the 'unicode' values are all unquoted 4-digit hex
values, but some of them look like decimal, e.g. 2010 hyphen.  It would be
nice to be able to interpret the value without examining the key (or
position in the hierarchy) to find out if it is hex.

jamesgk:
Hmm. It seems like we're pretty overdue for renaming this library "glyphsLib" and moving the glyphslib module to \__init__. Is it too late to do so?

schriftgestalt:
> We have files in which under instances/customParameters
typoAscender/typoDescender/typoLineGap/winAscent/winDescent... etc are all
quoted.  However under fontMaster the ascender/capHeight/descender are all
unquoted.

Thanks for finding this. I fixed it.

> glyphOrder, unicodeRange, panose values are indented, but other lists are
not.  These lists all seem to be dictionary values, so perhaps that's the
distinction?

That should already be fixed in the latest betas. 

> Hmm. It seems like we're pretty overdue for renaming this library "glyphsLib" and moving the glyphslib module to __init__. Is it too late to do so?

Better late than never ;)

PR #40, opened 2016-04-27
Rename to glyphsLib

jamesgk:
cc @schriftgestalt @behdad


anthrotype:
thanks James, it looks good to me :+1:

anthrotype:
after this we shall also rename the git repository name. GitHub is clever enough to redirect from the old url to the new one.

setup.py
```diff
@@ -17,12 +17,12 @@
...
 setup(
-    name='glyphs2ufo',
+    name='glyphsLib',
     version='1.0',
-    packages=['glyphs2ufo'],
+    packages=['glyphsLib'],
     entry_points={
         "console_scripts": [
-            "glyphs2ufo = glyphs2ufo.__main__:main"
+            "glyphsLib = glyphsLib.__main__:main"
```
anthrotype: @jamesgk Wouldn't it be better to keep the console script as `glyphs2ufo`?

jamesgk: That makes sense. Could we just do `glyphs2ufo = glyphsLib.__main__:main`? Or, rename `__main__.py` -> `glyphs2ufo.py`?

On the other hand, there has been interest in adding UFO->Glyphs functionality to this library, which would probably also be an option in the console script. Not sure when that would happen though.

anthrotype: `glyphs2ufo = glyphsLib.__main__:main` sounds good

anthrotype: I think `ufo2glyphs` could be a separate script. The name "glyphsLib" sounds more appropriate for the library rather than for it's executable.

jamesgk: Sounds good

behdad: > rename __main__.py -> glyphs2ufo.py?

I think that might be a good idea.  But you can do separately.

PR #42, opened 2016-04-27
Add production names to glif libs

jamesgk:
For now, this only adds production names which are present in the
Glyphs source.

Part of https://github.com/googlei18n/glyphsLib/issues/12


dougfelt:
LGTM

PR #43, opened 2016-05-02
Remove old instance_ufo output before writing

jamesgk:
For reason described in comments. When a UFO is overwritten, old glyphs are never removed. So we have to do this manually.

Should fix any issue for which removing old output was the solution. cc @marekjez86

### DIFF
--- a/Lib/glyphsLib/interpolation.py
+++ b/Lib/glyphsLib/interpolation.py
@@ -16,6 +16,7 @@
 from __future__ import print_function, division, absolute_import
 
 import os
+import shutil
 
 from glyphsLib.builder import set_redundant_data, set_custom_params,\
     clear_data, build_style_name, write_ufo, build_ufo_path, GLYPHS_PREFIX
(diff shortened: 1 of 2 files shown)


dougfelt:
Yes, this tripped me up too.  I was manually deleting the files associated with a target since I wasn't wiping the entire build trees to get around it.  This will help.

LGTM

Lib/glyphsLib/interpolation.py
```diff
@@ -40,6 +41,10 @@ def interpolate(ufos, master_dir, out_dir, instance_data,
         ufos, master_dir, out_dir, instance_data, italic)
 
     print('>>> Building instances')
+    # make sure old UFO data is removed (may contain deleted glyphs)
+    for path, _ in instance_files:
+        if os.path.exists(path):
+            shutil.rmtree(path)
```
behdad: Umm.  This makes me a bit dizzy.  Can you please only do that if path ends with '.ufo'?

PR #45, opened 2016-05-04
Metadata fixes

jamesgk:
Adds a couple more fields to UFO output, and uses a new heuristic to determine when fonts are italic (I'm not sure that anyone was using the old `italic` argument besides fontmake anyways).


dougfelt:
LGTM

PR #46, opened 2016-05-12
Add code for writing typed/untyped data to glyphs files.

dougfelt:
This changes the parsing code to handle octal/unicode escapes as well
as quotes.  This means that unicode values are returned from the parser,
and may be passed to the writer.

The casting code no longer normalizes curly quotes, they are maintained in
the data.  If they need to normalized clients will need to do that conversion.
This allows for better round-tripping of the original glyphs data.

Note that it is not possible to fully round-trip the glyphs source.  In
particular leading spaces are not preserved.  Quoting of strings is also
not fully preserved, instead the determination of whether strings need to
be quoted is done based on the contents of the string.  Glyphs seems to
quote and indent based on the position of the element in the hierarchy and
not on the content.  This does not appear to impact Glyphs' ability to
load the rewritten source.

Analysis of the casting code revealed that the dictionaries describing the
type structure were never modified and so did not need to be created/copied,
so the former functions returning them were changed to constants.

The original casting functions were converted to read methods of a collection
of objects, to which corresponding 'write' methods were added, and the
loop that iterated over the structure was changed to call a 'convert' method
with a parameter indicating whether to read or write.

In some cases the original casting functions were cloning parts of the
data structure they were operating on, and some of these were changed to
modify the data in place like the bulk of the code.  This was not done
in all cases, however-- in particular map is used in a number of functions
and it created a new list (which is then substituted for the original).

### DIFF
--- a/Lib/glyphsLib/casting.py
+++ b/Lib/glyphsLib/casting.py
@@ -20,7 +20,8 @@ import json
 import re
 
 __all__ = [
-    'cast_data'
+    'cast_data',
+    'uncast_data'
 ]
 
 
@@ -53,7 +54,8 @@ CUSTOM_FLOAT_PARAMS = frozenset((
 
 CUSTOM_TRUTHY_PARAMS = frozenset((
     'isFixedPitch', 'postscriptForceBold', 'postscriptIsFixedPitch',
-    "Don't use Production Names", 'DisableAllAutomaticBehaviour'))
+    u'Don\u2019t use Production Names', 'DisableAllAutomaticBehaviour'))
+
 
 CUSTOM_INTLIST_PARAMS = frozenset((
     'fsType', 'openTypeOS2CodePageRanges', 'openTypeOS2FamilyClass',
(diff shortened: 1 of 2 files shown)


Lib/glyphsLib/casting.py
```diff
@@ -53,312 +54,469 @@
...
+    def read(self, src):
+        """Cast a node from a string with format X Y TYPE [SMOOTH]."""
+        g = self._regex.match(src).groups()
+        return [num.read(g[0]), num.read(g[1]), g[2].lower(), bool(g[3])]
 
-def glyphs_datetime(string):
-    """Parse a datetime object from a string."""
+    def write(self, val):
+        assert isinstance(val, list) and len(val) == 4
+        v2 = val[2]
```
jamesgk: Let's call this `node_type` instead of `v2`. May also want to unpack groups in read() as
`x, y, node_type, smooth = self._regex.match(src).groups()`

Lib/glyphsLib/casting.py
```diff
@@ -53,312 +54,469 @@
...
-    return datetime_obj + offset
 
+class RWIntList(RWGlyphs):
+    """Read/write a list of ints."""
 
-def kerning(kerning_data):
-    """Cast the values in kerning data to ints."""
+    def read(self, src):
+        assert isinstance(src, list)
+        return map(int, src)
```
jamesgk: In hindsight, we shouldn't use map like this because it doesn't work the same way in python3. So let's either do `list(map(...))` or (as I'd personally prefer) just use a list comprehension.

Lib/glyphsLib/casting.py
```diff
@@ -53,312 +54,469 @@
...
+        assert isinstance(src, list)
+        return map(node.read, src)
 
-def normalized(string):
-    """Replace escaped characters with their intended characters.
-    Unescapes curved quotes to straight quotes, so that we can definitely
-    include this casted data in feature syntax.
-    """
+    def write(self, val):
+        return map(node.write, val)
```
jamesgk: Not asserting val is a list?

Lib/glyphsLib/parser.py
```diff
@@ -114,17 +114,113 @@ def _parse_list(self, text, i):
...
+
+class Writer(object):
+    """Write parsed data back to flat file.  Normalizes quoting
+    and indentation."""
+
+    # ints and floats are unquoted, as are strings of letters/digits and
+    # underscore with period or leading forward slash.  Everything else
+    # is quoted.
+    _sym_re = re.compile(
+        r'^(?:-?[0-9]+\.?[0-9]*|[_a-zA-Z0-9/\.][_a-zA-Z0-9\.]*)$')
```
jamesgk: Do we need to worry about floats without digits before the dot? Glyphs will read those fine, though I don't know if it will ever generate them.

Lib/glyphsLib/parser.py
```diff
@@ -114,17 +114,113 @@ def _parse_list(self, text, i):
...
+
+    def _write_dict(self, data):
+        if self.reorder:
+            keys = sorted(data.keys())
+        else:
+            keys = data.keys()
+        self.curindent += self.indent
+        pad = ' ' * self.curindent
+        out = self.out
+        out.write('{\n')
```
jamesgk: We need `out.write(' ' * self.curindent)` before this line too if I'm not mistaken (and to increment curindent after this line instead of before).

Lib/glyphsLib/parser.py
```diff
@@ -114,17 +114,113 @@ def _parse_list(self, text, i):
...
+        self.curindent -= self.indent
+        out.write(' ' * self.curindent)
+        out.write('}')
+
+    def _write_list(self, data):
+        first = True
+        self.curindent += self.indent
+        pad = ' ' * self.curindent
+        out = self.out
+        out.write('(')
```
jamesgk: Same as last comment

Lib/glyphsLib/parser.py
```diff
@@ -114,17 +114,113 @@ def _parse_list(self, text, i):
...
+    def _write_atom(self, data):
+      data = Writer._escape_re.sub(Writer._escape_fn, data)
+      out = self.out
+      if Writer._sym_re.match(data):
+          out.write(data)
+          return
+      data = data.replace('"', '\\"')
+      out.write('"')
+      out.write(data)
+      out.write('"')
```
jamesgk: How about just `out.write('"%s"' % data)`

jamesgk:
LGTM in general. But, we may need to change builder.py before merging as I suspect UFO exporting is broken by not normalizing feature text.

jamesgk:
> we may need to change builder.py before merging as I suspect UFO exporting is broken by not normalizing feature text.

Never mind, the normalization has just been generalized in the parser, right? So disregard that.

Lib/glyphsLib/parser.py
dougfelt: Nope.  All types are written starting on the current line and at the current position, including lists and dicts.  The opening brace, for example, needs to follow the 'keyword = ' of the enclosing dict.

Lib/glyphsLib/casting.py
dougfelt: ok

Lib/glyphsLib/casting.py
dougfelt: Switched to call a function that applies the fn to the list in place, instead of using map.

Lib/glyphsLib/casting.py
dougfelt: The check is now in the new utility function that modifies the list in place, which will fix this.

Lib/glyphsLib/parser.py
dougfelt: I don't think we encounter these, but I modified the regex.  A bit cumbersome since we don't want to allow '-.' as a degenerate float.

Lib/glyphsLib/parser.py
dougfelt: I prefer to just write each piece to the underlying buffer rather than build an intermediate copy and then write that.  We've already copied data twice, kind of didn't want to do it again.  I'm already tempted to move the quote escaping into the _escape_re/fn...

jamesgk: > I'm already tempted to move the quote escaping into the _escape_re/fn...

That would make sense to me. No strong preference though.

Lib/glyphsLib/casting.py
```diff
@@ -76,6 +82,8 @@ def write(self, val):
         raise NotImplementedError('%s write' % type(self).__name__)
 
 
+
+
```
jamesgk: Unintentional new lines?

dougfelt:
Ok, fixed those. PTAL.

jamesgk:
LGTM

Lib/glyphsLib/parser.py
schriftgestalt: I removed all indents from the .glyphs files in the latest version. There where some indentations in 'userdata' parts, but that should be fixed now.

Issue #21, opened 2016-03-27
weightValue and widthValue should be num, NOT int


behdad:
IMO these are continuous numbers, so should NOT be integers.  @schriftgestalt can you please confirm?

schriftgestalt:
Do you mean the master coordinates? They are floats internally.

PR #47, opened 2016-05-17
Change interpolation{Weight,Width,Custom} from int to num

mashabow:
Glyphs seems to accept floats for instances' interpolation values, so they should be `num`, like #21.


mashabow:
Our company (Type Project Inc.) signed the Corporate CLA yesterday. Thanks.

PR #48, opened 2016-05-18
Normalize custom parameter names

jamesgk:
(Turns out this *is* necessary, I just hadn't done enough testing)

### DIFF
--- a/Lib/glyphsLib/builder.py
+++ b/Lib/glyphsLib/builder.py
@@ -336,8 +336,8 @@ def set_custom_params(ufo, parsed=None, data=None, misc_keys=(), non_info=()):
 
 def normalize_custom_param_name(name):
     """Replace curved quotes with straight quotes in a custom parameter name.
-    We may want to do this for all keys in Glyphs data, but for now this is the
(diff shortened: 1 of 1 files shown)


schriftgestalt:
Can you send me the file where this happened? And where did the file come from?

jamesgk:
For example: https://raw.githubusercontent.com/googlei18n/noto-source/master/src/NotoSansHebrew-MM.glyphs. The "Don't use Production Names" parameters use a curved apostrophe in Don't.

schriftgestalt:
Please report this directly that I can fix it.

schriftgestalt:
And it should only happen for custom parameters as that is the only place where the user can type.

jamesgk:
> Please report this directly that I can fix it.

Sure, where should we report bugs directly to Glyphs?

> And it should only happen for custom parameters as that is the only place where the user can type.

Good to know, I'll update the comment.

schriftgestalt:
> Sure, where should we report bugs directly to Glyphs?

The easiest is to send me an email.

> only place where the user can type.

I changed the configuration of the textfield and textview that it should not happen any more.

PR #49, opened 2016-05-19
Try to source metadata from regular master

jamesgk:
This tries to use the most regular master (either the master with
style Regular, or a style common between masters) before falling back
on just using the first master.

Also fixes a bug in which the first master's style was always included
in the designspace path.

Part of https://github.com/googlei18n/fontmake/issues/73

### DIFF
--- a/Lib/glyphsLib/interpolation.py
+++ b/Lib/glyphsLib/interpolation.py
@@ -86,33 +86,39 @@ def build_designspace(masters, master_dir, out_dir, instance_data):
 def add_masters_to_writer(writer, ufos):
     """Add master UFOs to a MutatorMath document writer.
 
-    Returns the masters' base family and style names, as determined by taking
-    the intersection of their individual family/style names. This is used for
+    Returns the masters' family name and shared style names. These are used for
     naming instances and the designspace path.
     """
 
-    base_family = ''
-    base_style = ''
-    specify_info_source = True
+    master_data = []
+    base_family = None
+    base_style = None
 
     for font in ufos:
         family, style = font.info.familyName, font.info.styleName
-        if family in base_family or not base_family:
+        if base_family is None:
             base_family = family
-        elif base_family not in family:
-            raise ValueError('Inconsistent family names for masters')
-        if style in base_style or not base_style:
-            base_style = style
+        assert family == base_family, 'Masters must all have same family name'
+        if base_style is None:
+            base_style = style.split()
+        base_style = [s for s in style.split() if s in base_style]
+        master_data.append((font.path, family, style, {
+            s: font.lib.get(GLYPHS_PREFIX + s + 'Value', DEFAULT_LOC)
+            for s in ('weight', 'width', 'custom')}))
+
+    base_style = ' '.join(base_style)
+    info_source = 0
+    for i, (path, family, style, location) in enumerate(master_data):
+        if family == base_family and style == (base_style or 'Regular'):
+            info_source = i
+            break
+
+    for i, (path, family, style, location) in enumerate(master_data):
+        is_base = (i == info_source)
         writer.addSource(
-            path=font.path,
-            name='%s %s' % (family, style),
-            familyName=family, styleName=style,
-            location={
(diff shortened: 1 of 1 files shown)


Lib/glyphsLib/interpolation.py
```diff
@@ -86,33 +86,39 @@ def build_designspace(masters, master_dir, out_dir, instance_data):
...
+        if base_family is None:
             base_family = family
-        elif base_family not in family:
-            raise ValueError('Inconsistent family names for masters')
-        if style in base_style or not base_style:
-            base_style = style
+        assert family == base_family, 'Masters must all have same family name'
+        if base_style is None:
+            base_style = style.split()
+        base_style = [s for s in style.split() if s in base_style]
```
dougfelt: else: ?

Lib/glyphsLib/interpolation.py
```diff
@@ -86,33 +86,39 @@ def build_designspace(masters, master_dir, out_dir, instance_data):
...
+            for s in ('weight', 'width', 'custom')}))
+
+    base_style = ' '.join(base_style)
+    info_source = 0
+    for i, (path, family, style, location) in enumerate(master_data):
+        if family == base_family and style == (base_style or 'Regular'):
+            info_source = i
+            break
+
+    for i, (path, family, style, location) in enumerate(master_data):
```
dougfelt: Not sure why you loop over master_data twice.  is_base is false until the family and style match, then it is true that one time, then it is false thereafter.

Lib/glyphsLib/interpolation.py
jamesgk: I don't think it makes any difference in behavior, but yea I'll add that for clarity.

Lib/glyphsLib/interpolation.py
jamesgk: This is for when there is no family/style match, in which case we default to index 0 and just use the first master. Does that make sense? If so, I will add a comment.

dougfelt: Yeah, that makes sense, didn't think about that case.  A comment would help.

Lib/glyphsLib/interpolation.py
dougfelt: Yeah, behavior is same, it's just extra work.  Or you could use 'if base_style is None or s in base_style'

behdad:
@schriftgestalt How does Glyphs figure out which master to use for metadata?  We probably want to use the same heuristic.

schriftgestalt:
What kind of metadata do you mean? If there are numerical values, they are interpolated.

behdad:
@jamesgk do you know what we copy from base master?  Or you just pass it to mutatorMath?

jamesgk:
MutatorMath handles this, but by my understanding it is all the non-numeric, non-master-specific info from the UFO info: http://unifiedfontobject.org/versions/ufo3/fontinfo.plist (for example copyright and version number), as well as feature sources which aren't interpolated such as GSUB.

schriftgestalt:
Copyright and version number will not be set on the master. It will always be set on the font.

That is one of the main problems with masters in ufos. There was a discussion about that at Typographics/TypoLab in NY. The idea was to have a file that holds all font level info. Like an extended designSpace file.

Would you be interested in doing the interpolation yourself. That would allow support for brace layers and such. MutatorMath could be used to calculate the factors for each master and then just multiply+add any number in the font ;) I could port some of my code as a foundation...

behdad:
> Would you be interested in doing the interpolation yourself.

That's what we do already, isn't it?

> That would allow support for brace layers and such.

In theory we can do that; I don't think anyone tested it.

I'm also interested in the other trick; bracket?  The substitutions.  Where are they stored?  We should translate them into .designspace substitutions.

schriftgestalt:
> That's what we do already, isn't it?

I didn’t look at the code but I supposed you let Mutator Math do it for you.

behdad:
> That's what we do already, isn't it?
> 
> I didn’t look at the code but I supposed you let Mutator Math do it for you.

Correct.

PR #50, opened 2016-05-20
Give more descriptive warnings about kerning

jamesgk:
For one thing, these are strictly conflicts and not duplicates. Also,
identify the master as there can be more than one conflict for the
same pair in a single Glyphs source (but not in a single master).


dougfelt:
LGTM

PR #51, opened 2016-05-25
Kerning resolution

jamesgk:
Matches Glyphs for kerning resolution.

Part of https://github.com/googlei18n/glyphsLib/issues/24, https://github.com/googlei18n/fontmake/issues/83


behdad:
Thanks!

PR #52, opened 2016-06-30
Add a "family-name" parameter to public functions

jamesgk:
This allows one to only build certain instances found in a Glyphs source, filtering by family name, or to give a custom family name to generated masters.

Related to https://github.com/googlei18n/noto-source/pull/6, we want to generate separate normal/UI masters for a certain script each with it's own interpolation rules (found in Glyphs source). This seemed like the most convenient way.


dougfelt:
LGTM

PR #53, opened 2016-07-07
Overhaul anchor propagation

jamesgk:
This makes anchor propagation (which would have failed both of the
added test cases) more robust, in response to recent updates at
https://github.com/googlei18n/glyphsLib/issues/33.

Specifically, we now a) adjust anchor positions when marks on
components were added at that anchor, not just marks on the parent
(done by recursively propagating and adjusting) and b) add ligature
anchors when there's more than one base component with a certain
anchor.


dougfelt:
LGTM

PR #54, opened 2016-07-08
Always parse into ordered dictionaries

jamesgk:
Now that kerning conflict resolution depends on the order the rules
are defined, it is always necessary to use an ordered dictionary to
store the Glyphs data.


dougfelt:
LGTM

PR #58, opened 2016-07-11
[parser] Only parse hex values up to 0xFFFF

jamesgk:
For consistency with older Glyphs versions. Should close https://github.com/googlei18n/glyphsLib/issues/55
cc @felipesanches @behdad


dougfelt:
LGTM

PR #65, opened 2016-08-02
Check version

jamesgk:
Enough to close https://github.com/googlei18n/glyphsLib/issues/64?


behdad:
Sgtm. Thanks.

schriftgestalt:
Maybe it would be a good idea to test that version number is at least 895. That marks the stable version of 2.3. Before that, there where some problems with auto alignment.

schriftgestalt:
But if you know that the file came from a stable 2.2, it should be fine. Only that there is no easy way to fine out.

PR #68, opened 2016-08-10
Use UFO3 guidelines and background

moyogo:


jamesgk:
There's probably no reason to keep the old functionality in. I just did things that way because that's what Glyphs does, but I have no idea what makes this data useful.

moyogo:
OK, it makes sense to just have UFO3 guidelines and background since the output format is UFO 3  anyway.

schriftgestalt:
What do you mean by 'What Glyphs does'? The way the guides are written? I did that for round trip compatibility with RiboFont. But I'm happy to change it to use the UFO3 syntax.

jamesgk:
> What do you mean by 'What Glyphs does'? The way the guides are written?

Right, I was just trying to match the output of Glyphs exactly for guidelines and backgrounds.

jamesgk:
So it seems like UFO3 guideline support is not great in Defcon or MutatorMath:
https://github.com/googlei18n/fontmake/issues/132
https://github.com/googlei18n/fontmake/issues/134

I'd really like to revert these changes until this is resolved.

anthrotype:
yeah.. :-/
Denis sent a PR for defcon https://github.com/typesupply/defcon/pull/85, but not (yet) for MutatorMath.

jamesgk:
Well, I'm going to merge https://github.com/googlei18n/glyphsLib/pull/70 in the meantime and we can re-incorporate these changes once the support is there.

anthrotype:
SGTM.

PR #73, opened 2016-09-16
[builder] make last node the first for closed contours

anthrotype:
This is in order to _keep_ the original starting point's position.

https://github.com/googlei18n/fontmake/issues/142


anthrotype:
Actually, it turns out Glyphs.app always treat the _last_ node as the starting point of closed contour, so I changed this accordingly.

anthrotype:
ping @jamesgk

anthrotype:
👍

PR #75, opened 2016-09-29
Expose more functionality for interpolating

jamesgk:
This returns a designspace path and instance data when a designspace
is generated in the build_master function, and adds a function for
incorporating this instance data (which reflects the structure of
Glyphs source) into actual UFOs.

Ideally I'd like to take out interpolation from glyphsLib completely,
and only have options for returning master UFOs and a designspace. But
for now I'm not sure if anyone is using the interpolation feature.

### DIFF
--- a/Lib/glyphsLib/__init__.py
+++ b/Lib/glyphsLib/__init__.py
@@ -73,17 +73,23 @@ def build_masters(filename, master_dir, designspace_instance_dir=None,
             written alongside the master UFOs though no instances will be built.
         family_name: If provided, the master UFOs will be given this name and
             only instances with this name will be included in the designspace.
+
+    Returns:
+        A list of master UFOs, and if designspace_instance_dir is provided, a
+        path to a designspace and a list of (path, data) tuples with instance
+        paths from the designspace and respective data from the Glyphs source.
     """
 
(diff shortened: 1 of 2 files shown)


Lib/glyphsLib/__init__.py
```diff
@@ -73,17 +73,23 @@ def build_masters(filename, master_dir, designspace_instance_dir=None,
...
     """
 
     ufos, instance_data = load_to_ufos(
         filename, include_instances=True, family_name=family_name)
     if designspace_instance_dir is not None:
-        build_designspace(ufos, master_dir, designspace_instance_dir,
-                          instance_data)
+        designspace_path, instance_data = build_designspace(
+            ufos, master_dir, designspace_instance_dir, instance_data)
+        return ufos, designspace_path, instance_data
```
dougfelt: I guess it's a python thing, to return tuples of different lengths based on arguments you pass in.  But it's easy to trip yourself up when refactoring, it seems to me.  I guess my Java bias is showing.

jamesgk: Yea, I don't love it myself. But I'm not sure how else to return this information without breaking the existing API.

dougfelt:
LGTM

PR #77, opened 2016-10-02
Use font.lib['public.postscriptNames'] instead of glyph.lib['public.postscriptName']

### LINKED ISSUE #76: use new `public.postscriptNames` lib key
As I explained in https://github.com/googlei18n/ufo2ft/issues/72, we need to replace uses of `public.postscriptName` (e.g. [here](https://github.com/googlei18n/glyphsLib/blob/9387810ea3c540bb4d0412282e8a83c9e8b6bc21/Lib/glyphsLib/builder.py#L618)) with the new `public.postcriptNames`, that was added last week to the UFO spec (see https://github.com/unified-font-object/ufo-spec/commit/c8620fff3a5380bbe433da89e4a70b5549614205).

moyogo:
Fixes #76.

### DIFF
--- a/Lib/glyphsLib/builder.py
+++ b/Lib/glyphsLib/builder.py
@@ -615,7 +615,10 @@ def load_glyph(glyph, layer, glyph_data):
         glyph.lib[glyphlib_prefix + 'Export'] = export
     production_name = glyph_data.get('production')
     if production_name is not None:
-        glyph.lib[PUBLIC_PREFIX + 'postscriptName'] = production_name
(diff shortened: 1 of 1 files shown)


anthrotype:
i suggest we merge this after the corresponding https://github.com/googlei18n/ufo2ft/issues/72 is addressed, and then bump fontmake's requirements for both glyphsLib and ufo2ft.

moyogo:
See googlei18n/ufo2ft#73.

PR #81, opened 2016-11-01
Explicitly set OpenType ascender/descender fields

jamesgk:
These turn out to be pretty complicated and we can't trust they'll
be set correctly by a UFO compiler just from the generic fields.

Glyphs does something a little different, but this is a simple
treatment which yields predictable results for the common case where
every master has the same ascender and descender values.

### DIFF
--- a/Lib/glyphsLib/builder.py
+++ b/Lib/glyphsLib/builder.py
@@ -293,14 +293,16 @@ def set_redundant_data(ufo):
     ufo.info.openTypeNamePreferredSubfamilyName = style_name
 
     ascender = ufo.info.ascender
-    ufo.info.openTypeHheaAscender = ascender
-    ufo.info.openTypeOS2TypoAscender = ascender
-    ufo.info.openTypeOS2WinAscent = ascender
+    if ascender is not None:
+        ufo.info.openTypeHheaAscender = ascender
+        ufo.info.openTypeOS2TypoAscender = ascender
+        ufo.info.openTypeOS2WinAscent = ascender
 
     descender = ufo.info.descender
-    ufo.info.openTypeHheaDescender = descender
-    ufo.info.openTypeOS2TypoDescender = descender
-    ufo.info.openTypeOS2WinDescent = abs(descender)
+    if descender is not None:
+        ufo.info.openTypeHheaDescender = descender
+        ufo.info.openTypeOS2TypoDescender = descender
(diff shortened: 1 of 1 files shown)


anthrotype:
As Georg and Denis have explained in the other thread, what font editors and type designers mean by "ascender" and "descender" is not the same as as those from hhea and OS/2 specs. The line spacing has to be defined explicitly.

jamesgk:
OK, but Glyphs still populates the hhea and OS/2 fields with these generic values in the common case where every master has the same ascender and descender values. I think this PR is at least an improvement and if it's still insufficient I'd welcome someone else to make another.

anthrotype:
Why no let ufo2ft use the new fallback values from https://github.com/googlei18n/ufo2ft/pull/79 ?

jamesgk:
Because then fontmake will generate different output than Glyphs from the same source.

anthrotype:
Even if one defines explicit custom parameters for hhea and OS/2 typo/win ascenders/descenders?

jamesgk:
No, not in that case.

anthrotype:
So IMO the solution would be to fix the font sources. But if you say that Glyphs app does that, I trust you. I haven't checked. Needless to say, if Glyphs does something does not mean that is always correct. :)

jamesgk:
> Needless to say, if Glyphs does something does not mean that is always correct. :)

Indeed! I thought this behavior was strange and for all I know it's a bug in Glyphs. But as long as it is the way it is I feel fontmake should basically behave like Glyphs when working from Glyphs source (at least so fontmake users won't complain).

anthrotype:
Sgtm. I wonder what @schriftgestalt has to say about this.

schriftgestalt:
What do you recommend should be done?

jamesgk:
@schriftgestalt is the behavior I described at https://github.com/googlei18n/glyphsLib/issues/78#issuecomment-257463078 expected? It seemed a little strange to me that when all ascender values were the same the results were different than what you described. But since that's what our vendors expect, that's what I tried to emulate here.

jamesgk:
Closing in favor of https://github.com/googlei18n/ufo2ft/pull/82

schriftgestalt:
same [answer](https://github.com/googlei18n/glyphsLib/issues/78#issuecomment-259076137)

PR #82, opened 2016-11-29
Add util module for shared methods, use logging

jamesgk:


Lib/glyphsLib/util.py
```diff
@@ -0,0 +1,59 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+import logging
+import os
+import shutil
+
+logger = logging.getLogger('glyphsLib')
```
anthrotype: same comment as in https://github.com/googlei18n/cu2qu/pull/61#discussion_r90267521 applies here.

We don't need to use a single named logger for the whole package, sometimes it's handy to know from where a message was emitted, or configure sub-loggers differently.

Note that if fontmake only wishes to display the top-level logger name whenever, e.g., glyphsLib or cu2qu sub-loggers emit a message, then it could use a custom formatter to do that.

jamesgk:
Thanks @anthrotype, PTAL

PR #83, opened 2016-12-07
set up PyPI deployment on tags

anthrotype:
This is just copy-pasta from my previous packaging work for ufo2ft and similar.

One thing worth mentioning here is a pending PR for MutatorMath, https://github.com/LettError/MutatorMath/pull/63. Once the latter is merged, and MutatorMath is finally uploaded to PyPI, we can unlock that package as part of the glyphLib's `install_requires`.
Until that, MutatorMath has to be installed separately to be used with `glyphsLib.interpolation`.

James, Behdad, etc. if you sign up to PyPI (https://pypi.python.org/pypi?%3Aaction=register_form) I can give you admin access to all the PyPI projects belonging to googlei18n which I'm maintaining.


anthrotype:
https://github.com/LettError/MutatorMath/pull/63 is now merged (thanks @LettError!), so MutatorMath is now on PyPI too.

This PR is good to go now.

anthrotype:
I just moved the test modules in a separate `tests` folder, outside of the `Lib/glyphsLib` package, just like I did for ufo2ft (https://github.com/googlei18n/ufo2ft/pull/90).
I think this is better, as users won't run those tests anyway, and we can add more test data in the future.

anthrotype:
@jamesgk could we merge this and tag a release?

anthrotype:
OK, [v1.1.0](https://github.com/googlei18n/glyphsLib/releases/tag/v1.1.0) tagged.

The sdist and wheel packages are now on PyPI:
https://pypi.org/project/glyphsLib/

All fontmake's dependencies can now be pip install-ed.
Coming up, well.. fontmake itself :)

PR #92, opened 2016-12-21
Store production name and unicode string from GlyphData

brawer:
Currently, the result has attributes `name`, `production_name`
and `unicode`. Additional properties will be added later.

Bump requirements to a new version of fonttools. The implementation
depends on a recently added function `fontTools.agl.toUnicode()`.

https://github.com/googlei18n/glyphsLib/issues/88
https://github.com/googlei18n/glyphsLib/issues/12


anthrotype:
though we haven't tagged fonttools 3.3.2 yet.
I think we should do that before we can merge this.

anthrotype:
yep
https://travis-ci.org/googlei18n/glyphsLib/jobs/185720089#L176

anthrotype:
also, note that the `requirements.txt` is only used to set up the environment that runs the test suite.
To bump the minimum required version of a install requirement you need to do it in the setup.py `install_requires` (use `>=` instead of `==` in setup.py).

anthrotype:
btw, I think the next fonttools should be 3.4.0, not 3.3.2. As we added some new (backward-compatible) features, not just fixed bugs.

anthrotype:
so.... Python 3.4 does not support `b"string %s" % (b"interpolation")`, but Python 3.5 (and of course 2.7) does.

Shall we take the change to kill python 3.4, now that 3.6 is out next week? 😁

anthrotype:
oh my.. too many tabs open.
ignore this please.. :flushed:

anthrotype:
I just tagged fonttools 3.4.0 which includes among other things the required `agl.toUnicode()` function.

I also sent a PR to `brawer:glyphdata` branch that updates the glyphsLib min requirements:

https://github.com/brawer/glyphsLib/pull/1

anthrotype:
yes @googlebot, I am okay with my commits being contributed to this project.

@brawer I wonder if the `glyphdata_generated.py` file should record the commit hash of the `schriftgestalt/GlyphsInfo` repository from which the GlyphData.xml was downloaded?

anthrotype:
Another option is to add `schriftgestalt/GlyphsInfo` repo as a git submodule of `glyphsLib`, and run the `generate_glyphdata.py` script from the local submodule checkout, instead of curl-ing the file from `raw.githubusercontent.com` as it does now.

brawer:
Sounds like a good idea. Let’s get consensus on how to do it: https://github.com/googlei18n/glyphsLib/issues/93

In the meantime, I’m taking the liberty to merge this change since a couple bugs (which cause glyphsLib to emit broken output) are blocked on getting access to GlyphData.

PR #98, opened 2017-01-23
Use GlyphData for default production names

brawer:
Resolves https://github.com/googlei18n/glyphsLib/issues/12

### DIFF
--- a/tests/builder_test.py
+++ b/tests/builder_test.py
@@ -282,15 +282,14 @@ class ToUfosTest(unittest.TestCase):
         data = self.generate_minimal_data()
         self.add_glyph(data, 'foo')['production'] = 'f_o_o.alt1'
         ufo = to_ufos(data)[0]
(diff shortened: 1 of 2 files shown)


tests/builder_test.py
```diff
@@ -285,6 +285,14 @@ def test_postscript_name_from_data(self):
         postscriptNames = ufo['foo'].font.lib.get('public.postscriptNames')
         self.assertEqual(postscriptNames, {'foo': 'f_o_o.alt1'})
 
+    def test_postscript_name_from_glyph_name(self):
+        data = self.generate_minimal_data()
+        self.add_glyph(data, 'C-fraktur')
+        ufo = to_ufos(data)[0]
+        postscriptNames = ufo['C-fraktur'].font.lib.get(
```
anthrotype: doesn't `ufo.lib.get('public.postscriptNames')` work the same here?

brawer: Indeed, thanks for noticing! Changed.

PR #100, opened 2017-01-25
Generate GlyphClassDef statements in auto-generated table GDEF

brawer:
Resolves https://github.com/googlei18n/glyphsLib/issues/85

### DIFF
--- a/Lib/glyphsLib/builder.py
+++ b/Lib/glyphsLib/builder.py
@@ -151,10 +151,10 @@ def to_ufos(data, include_instances=False, family_name=None, debug=False):
         set_robofont_glyph_background(glyph, bg_name, bg_data)
 
     for ufo in ufos.values():
+        ufo.lib[PUBLIC_PREFIX + 'glyphOrder'] = glyph_order
         propagate_font_anchors(ufo)
         add_features_to_ufo(ufo, feature_prefixes, classes, features)
         add_groups_to_ufo(ufo, kerning_groups)
-        ufo.lib[PUBLIC_PREFIX + 'glyphOrder'] = glyph_order
 
     for master_id, kerning in data.pop('kerning', {}).items():
         load_kerning(ufos[master_id], kerning)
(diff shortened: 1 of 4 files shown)


Lib/glyphsLib/builder.py
```diff
@@ -695,17 +695,44 @@ def add_groups_to_ufo(ufo, kerning_groups):
...
+        # Currently, we leave components empty but that is probably wrong.
+        # https://github.com/googlei18n/glyphsLib/issues/99
+        if has_caret or subCategory == 'Ligature':
+            ligatures.add(glyph.name)
+        elif category == 'Mark' and (subCategory == 'Nonspacing' or
+                                     subCategory == 'Spacing Combining'):
+            # https://github.com/googlei18n/ufo2ft/issues/91#issuecomment-266530863
+            marks.add(glyph.name)
+        else:
+            bases.add(glyph.name)
```
anthrotype: I guess that means any glyph which is neither Ligature nor Mark is currently implied to be Base?

brawer: Yes, but what would be the correct behavior?

anthrotype: i'm not so sure.. There's also a fifth option, i.e. no category at all:

> The font developer does not have to classify every glyph in the font, but any glyph not assigned a class value falls into Class zero (0). 

See https://www.microsoft.com/typography/otspec/GDEF.htm

anthrotype: From what I understand this "Base" class is not interpreted as residual (i.e. anything non-Mark or non-Ligature). A user could define "Base" class selectively, as in the example from the GDEF spec where one only assigns class values for Arabic glyphs, but not for Latin -- which are implicitly assigned class 0, the non-class.
When writing these features manually, one does what is most suitable for the case at hand.
But the usual question for glyphsLib is, how does Glyphs.app generates this classes automatically?

For example, by reading Georg's comment in https://github.com/googlei18n/glyphsLib/issues/99#issuecomment-275087699, it seems like anchors' `name.startswith('caret_')` is all that is required to place ligatures in GDEF class 2.

Whereas here, when a glyph does not have those special anchors, you are classifying glyphs as ligatures based on the GlyphData.xml "Ligature" sub-category. Is that what Glyph does?

brawer: @schriftgestalt?

schriftgestalt:
The Liga class is set if the subCategory "Ligature" and it has at least one attachment anchor.

brawer:
Thank you, @schriftgestalt! What’s the logic for assigning a glyph to GlyphClassDef Base and Compound? To summarize your replies so far:

* Base: ?
* Liga: if the subCategory "Ligature" and it has at least one attachment anchor
* Mark: category=="Mark" and (subCategory=="Nonspacing" || subCategory=="Spacing Combining")
* Compound: ?

schriftgestalt:
Base: Everything that has an attaching anchor and !Ligature and !Mark
Compound: not set

anthrotype:
Makes sense, thanks Georg!

Lib/glyphsLib/builder.py
```diff
@@ -695,17 +695,47 @@ def add_groups_to_ufo(ufo, kerning_groups):
...
+        # Glyphs.app assigns glyph classes like this:
+        # * Base: any glyph that has an attaching anchor and is neither
+        #   classified as Ligature nor Mark using the definitions below;
+        # * Ligature: if subCategory is "Ligature" and the glyph has at least
+        #   one attachment anchor;
+        # * Mark: if category is "Mark" and subCategory is either "Nonspacing"
+        #   or "Spacing Combining";
+        # * Compound: never assigned by Glyphs.app.
+        # https://github.com/googlei18n/glyphsLib/issues/85
+        if subCategory == 'Ligature' and glyph.anchors:
```
anthrotype: so, now you're adding ligatures to GDEF if they contain at least one anchor of any kind (`if glyph.anchors`).

But previously, you were classifying them as ligatures if they had an anchor whose name started with "caret_".

in https://github.com/googlei18n/glyphsLib/issues/99#issuecomment-275087699, @schriftgestalt said:

> Glyphs puts the ligatures in the GDEF if the glyph has caret_X anchors.

While in https://github.com/googlei18n/glyphsLib/pull/100#issuecomment-275195121, he said:

> The Liga class is set if the subCategory "Ligature" and it has at least one attachment anchor.

I don't know what he means exactly by "attachment" anchor, but I think the two statements above are a bit contradictory...

anthrotype: by combining the two statements above, he most likely meant that the Ligature GDEF class is set if the subcategory is "Ligature" and the glyph as at least one "caret_" anchor.

brawer: Changed.

anthrotype: Sorry if I'm such a pain, but I'm still not sure this is what Glyphs.app does.

Have you tried exporting a font with such ligatures+caret_anchors and see if it includes them in the Ligaure GDEF class?

Maybe by "attachment anchors", Georg meant those anchors used for cursive attachment in scripts like Arabic, for which it uses special names 'entry' and 'exit':

https://glyphsapp.com/tutorials/arabic

brawer:
Well, maybe @schriftgestalt could have a quick look at this change to confirm that it’s correct? I’ve squashed it.

schriftgestalt:
I’m still not 100% fit after having the flu.
there are two ligature related things in GDEF. The liga class and the liga caret definition. The later is defined by the `caret_X` anchors.

anthrotype:
sorry to hear that... :face_with_thermometer: 
I hope you get well soon!

> there are two ligature related things in GDEF. The liga class and the liga caret definition

it's the first one (`GlyphClassDef`) which we are trying to understand how Glyphs.app sets it.

Is a glyph classified as "Ligature" in the GDEF whenever its sub-category from the GlyphsData.xml is also "Ligature"?
Or is does it have to have anchors of some kind?
Or both?

schriftgestalt:
It need to be set as 'Ligature' needs at least one attaching anchor ('top'), '_caret_X' or '_top' does not count.

anthrotype:
So in other words it must have at least one anchor whose name does not start with a "_".

anthrotype:
@brawer I pushed 054d2d6 to `glyphclassdef` branch to fix a merge conflict following c469dc1. I'm gonna tag v1.3.0 without this; we can cut a new release once this is completed.

anthrotype:
and @googlebot is drunk.

PR #105, opened 2017-01-30
Fix mapping of GlyphData categories

brawer:
https://github.com/googlei18n/glyphsLib/issues/103
https://github.com/googlei18n/fontmake/issues/234


anthrotype:
Thank you Sascha!
I'll tag a patch release when I get to the office (unless you have something else to push).

brawer:
Thank you! (No changes planned from my side.)

anthrotype:
@brawer since master includes #100, and that is a new feature rather than a bug fix, I guess I should bump the minor version as well, so it will be `v1.4.0` instead of `v1.3.1`.

Or do we wait to release the auto-GlyphClassDef feature while we wait for @schriftgestalt feedback, and now we just tag `v1.3.1` without it? I could make a branch and tag from there.

WDYT?

brawer:
I’d rather keep it simple, as in bumping the version and be done.

anthrotype:
done!
https://github.com/googlei18n/glyphsLib/releases/tag/v1.4.0
https://pypi.python.org/pypi/glyphsLib/1.4.0

schriftgestalt:
> we wait for @schriftgestalt feedback, 

What exactly do you need from me? I didn’t see any need to cover all unicode categories. I just used it as a basis for things I needed to know. And as those glyphs don't need any special processing, I was not interested in them. I can add them to the glyphData if is changes anything.
So every time you think about coding some exceptions, you could also send a pull request to the glyphData repo ;)

anthrotype:
> What exactly do you need from me?

I meant the issue of how exactly glyphs are added to the GDEF table's `GlyphClassDef` in the Glyphs.app-generated features.fea:

https://github.com/googlei18n/glyphsLib/issues/85#issuecomment-275671785
https://github.com/googlei18n/glyphsLib/pull/100

> every time you think about coding some exceptions, you could also send a pull request to the glyphData repo ;)

sounds good to me!

schriftgestalt:
I can’t find it right now but someone made a summary and that was quite good.

anthrotype:
the point which is still unclear is how Glyphs.app assigns GDEF class "Ligature".

Currently, glyphsLib assigns the "Ligature" GDEF class if:
- the GlyphData.xml subCategory for the glyph is "Ligature", and 
- the glyph has at least one anchor whose name does not start with underscore

https://github.com/googlei18n/glyphsLib/pull/100/files#diff-a584754af9d770fb6242be7bf1b357bdR731

However, if you manage to read my long comment https://github.com/googlei18n/glyphsLib/issues/85#issuecomment-275671785, you'll see that these criteria either do too much or not enough.

Also, such auto-generated GlyphClassDef statement seems to be linked with other auto-generated features (e.g. mark), in the sense that it's only generated when also those feature which are supposed to make use of it are. Whereas here we treat in isolation.

I wonder if defining these GlyphClassDef should actually be the job of ufo2ft markFeatureWriter.
Or if glyphsLib should take the responsibility of auto-generating the other features as well.

brawer:
> I wonder if defining these GlyphClassDef should actually be the job of ufo2ft markFeatureWriter.

See https://github.com/googlei18n/ufo2ft/issues/91 for a past discussion.

@anthrotype, did you realize that the `features.fea` saved from Glyphs.app into UFOs is not the full story? To figure out what Glyphs really does, I’d recommend entering custom OpenType feature definitions which have an intentional syntax error. When exporting such a borked font to OTF, the Adobe OpenType compiler will crash, and Glyphs will allow you to inspect the failing feature file via a “Show Details” button. When you click that, Glyphs will open the failing folder in the Macintosh Finder. Other than the `features.fea` saved by Glyphs.app into UFO, the feature file inside that failed font folder will include the file which Glyphs.app sent to the Adobe compiler. This file includes `table GDEF` and everything else.

anthrotype:
Yes, of course. That's what I'm look at: the files in` ~/Library/Application Support/Glyphs/Temp` folder, which are sent to the embedded makeotf compiler.

PR #110, opened 2017-02-07
Round integer custom parameters

anthrotype:
glyphsLib expect some of the custom parameters, e.g. `underlinePosition`, to contain integer values. But if the glyphs source happens to contain a float string, casting that to int() would raise a ValueError. Better rounding the float to integer instead.

Partial fix for googlei18n/fontmake#247

BTW, I wonder from where @jamesgk got the full list of custom parameters, and whether they have to be interpreted as integer, float, boolean, list of integers or plain strings.
Does anybody know?

### DIFF
--- a/Lib/glyphsLib/casting.py
+++ b/Lib/glyphsLib/casting.py
@@ -145,11 +145,10 @@ class RWNum(RWGlyphs):
 
     def read(self, src):
         float_val = float(src)
-        int_val = int(float_val)
-        return int_val if int_val == float_val else float_val
+        return int(float_val) if float_val.is_integer() else float_val
 
     def write(self, val):
-        assert isinstance(val, float) or isinstance(val, int)
+        assert isinstance(val, (float, int))
         return str(val)
 
 
--- a/Lib/glyphsLib/casting.py
+++ b/Lib/glyphsLib/casting.py
@@ -15,7 +15,7 @@
 
 from __future__ import (print_function, division, absolute_import,
                         unicode_literals)
-from fontTools.misc.py23 import basestring
+from fontTools.misc.py23 import basestring, round
 import datetime
 import logging
 import re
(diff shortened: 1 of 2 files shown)


jamesgk:
See https://github.com/googlei18n/glyphsLib/pull/14#issuecomment-184906546
about the list of custom parameters.

anthrotype:
Excellent, thank you James!

anthrotype:
In the customParameterTypes.plist `underlinePosition` is simply referred to as `Number`. There seems to be no distinction between `int` and `float`.

I'm not sure if we need to round to `int` when reading in stuff like `underlinePosition` (as this PR does), or if we should read floats for customParameter that refer to generic `Number` values.

I mean, `underlinePosition` as defined in the TrueType `post` table is a signed short integer `int16`, so it would make sense if it was rounded.

r? @schriftgestalt

anthrotype:
I tried to set `underlinePosition` to a float value `-77.5`, and saved the glyphs source file. Then I opened it in a text editor and I see the value is saved as a float, un-rounded.

When I export the TTF, the underlinePosition in the `post` table is `-52`, an integer. The value is the result of adding `round(underlinePosition) + (underlineThickness//2)`. The default underlineThickness is `50`, so `-77 + 25 = -52`.

So, it seems that Glyphs.app allows those values to be stored as float numbers in the source file, but then it rounds them to integer when it exports the binary fonts.

schriftgestalt:
> In the customParameterTypes.plist underlinePosition is simply referred to as Number

That is strange. If you look at the file itself, it says `integer`.

anthrotype:
Sorry Georg, you're right! Yesterday, I opened the plist with Xcode.app, and in there it was showing just "Number"... If I open it in a simple text editor, it does show "integer" or "real".

But if `underlinePosition` should be an integer according to the plist file, why is Glyphs.app allowing it to be saved as a float in the glyphs source file?

schriftgestalt:
The rounding is not enforced and currently it is not distinguished between float and int inside the custom parameters. Most values that usually should be ints are stored as floats. This is mostly to avoid rounding errors while interpolating. 

The values are rounded when 'used' to generate a otf.

anthrotype:
Thanks for clarifying this.
This means glyphsLib should also allow those values to stay float upon reading in the data. We should only round to int upon building a UFO if the UFO spec says they must be `int`.
I'll close this and write a new PR.

PR #123, opened 2017-02-21
Add unit tests for building designspace files

brawer:


tests/interpolation_test.py
```diff
@@ -0,0 +1,83 @@
...
+    }
+    return [m1, m2], instances
+
+
+class DesignspaceTest(unittest.TestCase):
+    def expect_designspace(self, masters, instances, expectedFile):
+        master_dir = tempfile.mkdtemp()
+        designspace, _ = build_designspace(
+            masters, master_dir, os.path.join(master_dir, "out"), instances)
+        with codecs.open(designspace, "r", "utf-8") as f:
```
anthrotype: `io.open` is the recommended way to open files in py2.py3, `codecs.open` is kept in python3 only for backward compatibility.

anthrotype: you can also do `from fontTools.misc.py23 import open`, which is bascially a wrapper around `io.open`.

brawer: Thank you! Done in https://github.com/googlei18n/glyphsLib/commit/5676819089be73a184b9487519f2a481c05d93ea.

PR #125, opened 2017-02-21
Support instances with custom family names

brawer:
Fixes https://github.com/googlei18n/glyphsLib/issues/124.

See also https://github.com/googlei18n/fontmake/issues/257.


anthrotype:
LGTM, thanks Sascha!

PR #130, opened 2017-02-22
Do not generate inactive instances

brawer:
Before this change, the code was looking for an attribute named
`active`, but this is not what Glyphs.app stores in its files when an
instance gets marked as inactive in the Glyphs user interface. Instead,
Glyphs calls this attribute `exports` in its files.

Fixes https://github.com/googlei18n/glyphsLib/issues/129.

### DIFF
--- a/Lib/glyphsLib/interpolation.py
+++ b/Lib/glyphsLib/interpolation.py
@@ -152,8 +152,11 @@ def add_instances_to_writer(writer, family_name, instance_data, out_dir):
             dimension_names.append(s)
 
     for instance in instance_data:
+        # Glyphs.app recognizes both "exports=0" and "active=0" as a flag
+        # to mark instances as inactive. Those should not be instantiated.
         # https://github.com/googlei18n/glyphsLib/issues/129
-        if not instance.pop('exports', True):
+        if (not int(instance.pop('exports', 1))
+            or not int(instance.pop('active', 1))):
             continue
 
         instance_family = default_family_name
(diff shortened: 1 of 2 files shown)


brawer:
@schriftgestalt, did early versions of Glyphs use `active` instead of `exports` as attribute for flagging instances as (in)active?

anthrotype:
I just tried to manually replace `exports = 0` with `active = 0` in a .glyphs source file, and Glyphs.app still recongizes the instance as inactive.
If I set `active = 1` or if I remove the attribute, the instance becomes active again.

So it appears that both `exports` and `active` are recognized by Glyphs.app (2.4.1, 976) and that they have the same meaning, although `exports` is the one currently used when ticking the checkbox from the GUI.

brawer:
Thanks, @anthrotype. Changed to recognize both. Have another look?

Lib/glyphsLib/interpolation.py
```diff
@@ -152,8 +152,11 @@ def add_instances_to_writer(writer, family_name, instance_data, out_dir):
             dimension_names.append(s)
 
     for instance in instance_data:
-
-        if not instance.pop('active', True):
+        # Glyphs.app recognizes both "exports=0" and "active=0" as a flag
+        # to mark instances as inactive. Those should not be instantiated.
+        # https://github.com/googlei18n/glyphsLib/issues/129
+        if (not int(instance.pop('exports', 1))
+            or not int(instance.pop('active', 1))):
```
anthrotype: In PEP8, one would indent a two-line `if` statement like this:
```diff
          if (not int(instance.pop('exports', 1))
-             or not int(instance.pop('active', 1))):
+                 or not int(instance.pop('active', 1))):
              continue
```

brawer: Done.

PR #132, opened 2017-02-22
Recognize custom parameters for weightClass and widthClass

brawer:
The new behavior matches what is documented in the Glyphs handbook.
After this change, glyphsLib will produce the same `openTypeOS2WeightClass`
and `openTypeOS2WidthClass` values as Glyphs.app.

Note that this change does _not_ fix the `fvar` table in variation
fonts, but at least glyphsLib now produces interpolation masters
with correct OS/2 tables so it's a step towards correctness.

### DIFF
--- a/Lib/glyphsLib/builder.py
+++ b/Lib/glyphsLib/builder.py
@@ -263,8 +263,10 @@ def set_redundant_data(ufo):
     family_name, style_name = ufo.info.familyName, ufo.info.styleName
 
     width, weight = parse_style_attrs(style_name)
-    ufo.info.openTypeOS2WidthClass = WIDTH_CODES[width]
-    ufo.info.openTypeOS2WeightClass = WEIGHT_CODES[weight]
+    if ufo.info.openTypeOS2WidthClass is None:
+        ufo.info.openTypeOS2WidthClass = WIDTH_CODES[width]
+    if ufo.info.openTypeOS2WeightClass is None:
+        ufo.info.openTypeOS2WeightClass = WEIGHT_CODES[weight]
 
     if weight and weight != 'Regular':
         ufo.lib[GLYPHS_PREFIX + 'weight'] = weight
@@ -326,6 +328,8 @@ def set_custom_params(ufo, parsed=None, data=None, misc_keys=(), non_info=()):
             ('hhea', 'Hhea'), ('description', 'NameDescription'),
             ('license', 'NameLicense'), ('panose', 'OS2Panose'),
             ('typo', 'OS2Typo'), ('unicodeRanges', 'OS2UnicodeRanges'),
+            ('weightClass', 'OS2WeightClass'),
+            ('widthClass', 'OS2WidthClass'),
             ('win', 'OS2Win'), ('vendorID', 'OS2VendorID'),
             ('versionString', 'NameVersion'), ('fsType', 'OS2Type'))
         for glyphs_prefix, ufo_prefix in opentype_attr_prefix_pairs:
@@ -339,6 +343,10 @@ def set_custom_params(ufo, parsed=None, data=None, misc_keys=(), non_info=()):
         # enforce that winAscent/Descent are positive, according to UFO spec
         if name.startswith('openTypeOS2Win') and value < 0:
             value = -value
+
+        # Some Glyphs files store weightClass/widthClass as string values.
+        if name in ('openTypeOS2WeightClass', 'openTypeOS2WidthClass'):
+            value = int(value)
 
         if name == 'glyphOrder':
             # store the public.glyphOrder in lib.plist
--- a/tests/builder_test.py
+++ b/tests/builder_test.py
@@ -326,6 +326,54 @@ class ToUfosTest(unittest.TestCase):
         postscriptNames = ufo.lib.get('public.postscriptNames')
         self.assertEqual(postscriptNames, {'C-fraktur': 'uni212D'})
 
+    def test_weightClass_default(self):
+        data = self.generate_minimal_data()
+        ufo = to_ufos(data)[0]
+        self.assertEqual(ufo.info.openTypeOS2WeightClass, 400)
+
+    def test_weightClass_from_customParameter_weightClass(self):
+        # In the test input, the width is specified twice: once as weight,
+        # once as customParameters.weightClass. We expect that the latter wins
+        # because the Glyphs handbook documents that the weightClass value
+        # overrides the setting in the Weight drop-down list.
+        # https://glyphsapp.com/content/1-get-started/2-manuals/1-handbook-glyphs-2-0/Glyphs-Handbook-2.3.pdf#page=202
+        data = self.generate_minimal_data()
+        master = data['fontMaster'][0]
+        master['weight'] = 'Bold'  # 700
+        master['customParameters'] = ({'name': 'weightClass', 'value': '698'},)
+        ufo = to_ufos(data)[0]
+        self.assertEqual(ufo.info.openTypeOS2WeightClass, 698)  # 698, not 700
+
+    def test_weightClass_from_weight(self):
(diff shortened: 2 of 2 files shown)


tests/builder_test.py
```diff
@@ -326,6 +326,54 @@ def test_postscript_name_from_glyph_name(self):
         postscriptNames = ufo.lib.get('public.postscriptNames')
         self.assertEqual(postscriptNames, {'C-fraktur': 'uni212D'})
 
+    def test_weightClass_default(self):
+        data = self.generate_minimal_data()
+        ufo = to_ufos(data)[0]
+        self.assertEqual(ufo.info.openTypeOS2WeightClass, 400)
+
+    def test_weightClass_from_customParameter_weightClass(self):
+        # In the test input, the width is specified twice: once as weight,
```
anthrotype: in this comment here you mean the "weight", not the "width".

Lib/glyphsLib/builder.py
```diff
@@ -340,6 +344,10 @@ def set_custom_params(ufo, parsed=None, data=None, misc_keys=(), non_info=()):
         if name.startswith('openTypeOS2Win') and value < 0:
             value = -value
 
+        # Some Glyphs files store weightClass/widthClass as string values.
+        if name in ('openTypeOS2WeightClass', 'openTypeOS2WidthClass'):
+            value = int(value)
```
anthrotype: the `casting.py` module defines `weightClass` and `widthClass` as `string`, maybe we should change that to `integer`?
https://github.com/googlei18n/glyphsLib/blob/3937aaa9866e1087b3b00826cb6c0e492cc9f098/Lib/glyphsLib/casting.py#L502-L503

brawer: I tried, but it wouldn’t work. The `weightClass` parameter has enum values such as `Bold`, whereas `customParameters.weightClass` has numeric values inside a string. For example, here’s a snippet from the [WeightTest.glyphs](https://github.com/googlei18n/fontmake/files/793138/WeightTest.zip) file posted on the bug:

```
{
  customParameters = ({ name = weightClass; value = "707"; });
  interpolationWeight = 215;
  isBold = 1;
  linkStyle = Regular;
  name = Fat;
  weightClass = Bold;
}
```

tests/builder_test.py
brawer: Fixed. Thanks for noticing.

Lib/glyphsLib/builder.py
anthrotype: Oh, now I see. The lines I linked above refer to the types of the `weightClass` and `widthClass` *attributes* of the `instances` element in the glyphs file structure.
Those should stay `string` since, as you pointed out, their value is an enum.

What I meant was to cast the `weightClass` and `widthClass` *custom parameters* to integer as we read them in, by adding them to the `CUSTOM_INT_PARAMS` set at the top of `casting.py`.

Note that in the handbook the `weightClass` and `widthClass` custom parameters are said to be of *integer* type:
https://glyphsapp.com/content/1-get-started/2-manuals/1-handbook-glyphs-2-0/Glyphs-Handbook-2.3.pdf#page=202

anthrotype: I propose this change:

```diff
diff --git a/Lib/glyphsLib/builder.py b/Lib/glyphsLib/builder.py
index 8a497e8..5dcc835 100644
--- a/Lib/glyphsLib/builder.py
+++ b/Lib/glyphsLib/builder.py
@@ -344,10 +344,6 @@ def set_custom_params(ufo, parsed=None, data=None, misc_keys=(), non_info=()):
         if name.startswith('openTypeOS2Win') and value < 0:
             value = -value
 
-        # Some Glyphs files store weightClass/widthClass as string values.
-        if name in ('openTypeOS2WeightClass', 'openTypeOS2WidthClass'):
-            value = int(value)
-
         if name == 'glyphOrder':
             # store the public.glyphOrder in lib.plist
             ufo.lib[PUBLIC_PREFIX + name] = value
diff --git a/Lib/glyphsLib/casting.py b/Lib/glyphsLib/casting.py
index e379020..bc6b745 100644
--- a/Lib/glyphsLib/casting.py
+++ b/Lib/glyphsLib/casting.py
@@ -51,7 +51,7 @@ CUSTOM_INT_PARAMS = frozenset((
     'smallCapHeight', 'typoAscender', 'typoDescender', 'typoLineGap',
     'underlinePosition', 'underlineThickness', 'unitsPerEm', 'vheaVertAscender',
     'vheaVertDescender', 'vheaVertLineGap', 'winAscent', 'winDescent',
-    'xHeight', 'year'))
+    'xHeight', 'year', 'weightClass', 'widthClass'))
 
 CUSTOM_FLOAT_PARAMS = frozenset((
     'postscriptBlueScale',))
diff --git a/tests/builder_test.py b/tests/builder_test.py
index aae1bf7..c9e4d85 100644
--- a/tests/builder_test.py
+++ b/tests/builder_test.py
@@ -340,7 +340,7 @@ class ToUfosTest(unittest.TestCase):
         data = self.generate_minimal_data()
         master = data['fontMaster'][0]
         master['weight'] = 'Bold'  # 700
-        master['customParameters'] = ({'name': 'weightClass', 'value': '698'},)
+        master['customParameters'] = ({'name': 'weightClass', 'value': 698},)
         ufo = to_ufos(data)[0]
         self.assertEqual(ufo.info.openTypeOS2WeightClass, 698)  # 698, not 700
 
@@ -364,7 +364,7 @@ class ToUfosTest(unittest.TestCase):
         data = self.generate_minimal_data()
         master = data['fontMaster'][0]
         master['width'] = 'Extra Condensed'  # 2
-        master['customParameters'] = ({'name': 'widthClass', 'value': '7'},)
+        master['customParameters'] = ({'name': 'widthClass', 'value': 7},)
         ufo = to_ufos(data)[0]
         self.assertEqual(ufo.info.openTypeOS2WidthClass, 7)  # 7, not 2
```

anthrotype: Actually, even if we add `'weightClass'` and `'widthClass'` to `CUSTOM_INT_PARAMS` set, we still have to ensure that `value = int(value)` inside the `set_custom_params` function.

That's because after https://github.com/googlei18n/glyphsLib/pull/120 we read custom parameters as generic numbers: so if their value happen to be a float (it shouldn't in this case, but Glyphs.app allows that as it does not enforce integers in custom parameters), then we would read them as float (or generic `num`).
However, for the UFO spec, the `openTypeOS2{Weight,Width}Class` must be integers, so we must make them so when setting these in the builder, if they're not already.

anthrotype: Just tried to set `weightClass` custom parameter to a float in Glyphs.app, and the latter didn't complain, the `weightClass` is stored like this in the .glyphs file:
```
{
name = weightClass;
value = "111.9";
}
```

After exporting the font, the OS/2 table has `<usWeightClass value="111"/>`.

So it seems that Glyphs.app is simply truncating the value, as in `int(value)`, and not `round(value)`.

brawer: Ah, got it; thanks for the explanation. Done.

Lib/glyphsLib/builder.py
```diff
@@ -344,10 +344,6 @@ def set_custom_params(ufo, parsed=None, data=None, misc_keys=(), non_info=()):
         if name.startswith('openTypeOS2Win') and value < 0:
             value = -value
 
-        # Some Glyphs files store weightClass/widthClass as string values.
-        if name in ('openTypeOS2WeightClass', 'openTypeOS2WidthClass'):
-            value = int(value)
```
anthrotype: I think we should still keep this `value = int(value)`, as the value could be a float, and ufoLib/defcon expect an int.

brawer: Done.

PR #142, opened 2017-03-02
[glyphdata] Improve the unicode value returned by get_glyph. Fixes #141

### LINKED ISSUE #141: [glyphdata] glyphdata.get_glyph sometimes returns multiple chars for `unicode` when `name` implies a ligature
Here's an example:

```py
>>> from glyphsLib import glyphdata
>>> glyphdata.get_glyph('T_h')
Glyph(name='T_h', production_name='T_h', unicode=u'Th', category=u'Letter', subCategory=u'Ligature')
```

The `unicode` value comes from  [the call to `agl.toUnicode(prodname`](https://github.com/googlei18n/glyphsLib/blob/master/Lib/glyphsLib/glyphdata.py#L37)

```py
>>> from fontTools import agl
>>> agl.toUnicode('T_h')
u'Th'
```

`agl.toUnicode` splits the string on `_` and maps each component to the result string.

I did expect as result a unicode string with a length of 1 or `None`. So that I can do `ord(result.unicode)` if unicode is not None.

graphicore:
Ensure get_glyph returns a unicode with length==1

Use special unicode string length calculation

As pointed out by @anthrotype in https://github.com/googlei18n/glyphsLib/issues/141#issuecomment-283495409
Thanks!

Don't use agl if a '.' or '_' is in name; restructuring. See #141

### DIFF
--- a/Lib/glyphsLib/glyphdata.py
+++ b/Lib/glyphsLib/glyphdata.py
@@ -32,16 +32,29 @@ Glyph = namedtuple("Glyph", "name,production_name,unicode,category,subCategory")
 
 def get_glyph(name, data=glyphdata_generated):
     prodname = data.PRODUCTION_NAMES.get(name, name)
-    unistr = data.IRREGULAR_UNICODE_STRINGS.get(name)
-    if unistr is None:
-        unistr = agl.toUnicode(prodname)
-    if unistr != "" and name not in data.MISSING_UNICODE_STRINGS:
-        unistr_result = unistr
-    else:
-        unistr_result = None
+    unistr = _get_unicode(name, prodname, data)
     category, subCategory = _get_category(name, unistr, data)
-    return Glyph(name, prodname, unistr_result, category, subCategory)
+    return Glyph(name, prodname, unistr, category, subCategory)
 
+if NARROW_PYTHON_BUILD:
+    def unilen(text):
+        if isinstance(text, bytes):
+            text = text.decode("utf_8")
+        text_utf32 = text.encode("utf-32-be")
+        return len(text_utf32)//4
+else:
+    unilen = len
+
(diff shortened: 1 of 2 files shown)


anthrotype:
hey Lasse, thank you for working on this. Sorry if I didn't reply to you earlier.
I was waiting for @brawer to have a look at this, as he originally wrote the `glyphdata` module.

The more I look into this, the more I'm convincing myself that we should replace the current approach of generating a minimal mapping plus a table of exceptions, and simply add the `GlyphsData.xml` as a package data in `glyphsLib`, parsing the XML data as it is, at runtime. Basically what Georg and Behdad where suggesting in https://github.com/googlei18n/glyphsLib/issues/88. 
I think cElementTree is fast enough, and there are much worse bottlenecks in the pipeline.
Also, doing that should make it easier to use a custom GlyphsData.xml file instead of the built-in one.

Anyway, I'll leave it to Sascha to reply when he has time.

graphicore:
I was thinking about parsing GlyphsData.xml myself as well, because that's basically what I need, all the rest this module does is too much for me. I need just the mapping from glyphs friendly names and production names to unicodes. For my case especially the `agl` approach is arguable, but with the changes I made to `_get_unicode` it's usable.

So, I'm happy either way, it's just nice to have an interface to query GlyphsData.xml.

madig:
Is this PR still relevant and if so, what's needed to go on?

anthrotype:
> we should replace the current approach of generating a minimal mapping plus a table of exceptions, and simply add the GlyphsData.xml as a package data in glyphsLib, parsing the XML data as it is, at runtime.

that's still my preferred approach going forward. I haven't had the time to work on this yet. No big priority

anthrotype:
will be superseded by #410

PR #147, opened 2017-03-09
write UFO3 guidelines and background

moyogo:
This reverts commit b0b8002701f4a4871683e84ad29c1a8079000de5.

See: Use UFO3 guidelines and background #68
https://github.com/googlei18n/glyphsLib/pull/68
UFO3 guidelines and background since glyphsLib writes UFO3.
Issues with UFO3 guideliens have been fixed for a while


schriftgestalt:
As a node, Glyphs does not write RoboFont guides and layers any more. The author of that app complained and demanded removal of that feature. He said nobody other than RoboFont should touch anything with a 'robofont' key.

adrientetar:
@moyogo Why did you close this? I think it can be merged now.

moyogo:
@adrientetar sorry, I noticed there was an issue with the layers, only the default and its background are being exported.

Issue #119, opened 2017-02-15
[parser] add support for binary data values


anthrotype:
The glyphsLib parser should support binary data values as defined in the ASCII property list specification:

https://developer.apple.com/library/content/documentation/Cocoa/Conceptual/PropertyLists/OldStylePlists/OldStylePLists.html

> NSData
> Binary data is enclosed in angle brackets and encoded in hexadecimal ASCII. Spaces are ignored.
> For example:
```
<0fbd777 1c2735ae>
```

see https://github.com/googlei18n/fontmake/issues/247#issuecomment-279165940

anthrotype:
According to the above mentioned "OldStylePList" specification, the NSData enclosed in angle brackets should be encoded as "hexadecimal ASCII", however when I set some `userData` from within Glyphs.app and save the .glyphs file, I don't get hexadecimal ASCII, but a *base64-encoded* string:

```python
from Foundation import *
import base64

font = Glyphs.font

glyph = font.glyphs['a']

data = NSData.alloc().initWithBytes_length_('the bytes', 9)

glyph.userData["customUserData"] = data

# font.save()
#
# the following data is written:
#
#   customUserData = <dGhlIGJ5dGVz>;

print(base64.b64decode('dGhlIGJ5dGVz'))

# this will print:
#   the bytes
```

anthrotype:
I suppose that the .glyphs format can no longer be considered equivalent to NeXTSTEP ascii plist.
So the glyphsLib parser should follow whatever Glyphs.app does; in this case, encode/decode strings enclosed in angle brackets as base64 rather than hexadecimal.

anthrotype:
Oh jeez... :man_facepalming: 

It turns out that a `userData` attribute (a dict with arbitrary used defined content) can be contained, not just in a font, font master, glyph and layer (where it is encoded as Base64), but also inside a node. However, *in this case only* an NSData value will be encoded as *hexadecimal* string instead of Base64.

```python
from Foundation import *
import base64

font = Glyphs.font
glyph = font.glyphs['a']
data = NSData.alloc().initWithBytes_length_('the bytes', 9)
glyph.layers[0].paths[0].nodes[0].userData['customNodeUserData'] = data
font.save()
```

This is what you get in the .glyphs source: 
```
"328 -3 CURVE SMOOTH {\ncustomNodeUserData = <74686520 62797465 73>;\n}",
```

As you can see, it's an hex string:
```python
>>> from binascii import unhexlify
[...]

PR #150, opened 2017-03-17
Add command line option to pass in a custom GlyphData.xml file

punchcutter:
Moved GlyphData.xml loading code from generate_glyphdata.py to
glyphsLib/util.py to use with custom loading.

Glyph data and therefore GDEF are now constructed by
1. Reading the default GlyphData.xml (saved in glyphdata_generated.py)
2. Reading a custom GlyphData.xml if passed
3. Reading any overrides saved in the .glyphs file

Zero the width of Nonspacing Marks when writing to UFO (like Glyphs)
As discussed in https://github.com/googlei18n/glyphsLib/issues/115 this could
be moved to ufo2ft? Also we need to check for the Glyphs custom parameter 
DisableAllAutomaticBehaviour that ignores this width zero behaviour.
Tested and looks good with Noto Sans Hebrew as reported here: 
https://github.com/googlei18n/fontmake/issues/273

Added test for loading a custom GlyphData.xml.
Added test for reading a .glyphs source file with category overrides saved in the file.

### DIFF
--- a/Lib/glyphsLib/__init__.py
+++ b/Lib/glyphsLib/__init__.py
@@ -21,11 +21,12 @@ from __future__ import (print_function, division, absolute_import,
 from io import open
 import logging
 
+from glyphsLib import glyphdata_generated
 from glyphsLib.builder import to_ufos
 from glyphsLib.casting import cast_data
 from glyphsLib.interpolation import interpolate, build_designspace
 from glyphsLib.parser import Parser
-from glyphsLib.util import write_ufo
+from glyphsLib.util import fetch, write_ufo, fetch_all_glyphs, build_data, test_data, GlyphData
 
 
 __version__ = "1.5.3.dev0"
@@ -56,8 +57,52 @@ def loads(value):
     return data
 
 
+def load_glyph_data(data=glyphdata_generated, glyphs_source=None, custom_glyph_xml=None):
+    """Load GlyphData.xml
+    If a custom GlyphData.xml exists merge it with the default GlyphData.xml
+    Two overrides are possible:
+        1. Reading from a custom GlyphData.xml
+        2. Manually changing the info in Glyphs app
+    Args:
+        data: pre-generated glyph data
+        glyphs_source: unpacked .glyphs object
+        custom_glyph_xml: Path to a custom GlyphData.xml file
+    """
+    glyph_dict = data.DEFAULT_GLYPH_DICT.copy()
+    if custom_glyph_xml:
+        custom_glyphs = {}
+        lines = None
+        with open(custom_glyph_xml, 'r') as f:
+            lines = f.read()
+        if lines:
+            custom_glyphs = fetch_all_glyphs(paths=(lines,))
+        glyph_dict.update(custom_glyphs)
+
+    # Check the .glyphs file for any overrides
+    if glyphs_source:
+        for g in glyphs_source['glyphs']:
+            name = g.get('glyphname')
+            category = g.get('category')
+            subCategory = g.get('subCategory')
+            if g.get('category') is None and g.get('subCategory') is None:
+                continue
+            if name in glyph_dict:
+                glyph_dict[name]['category'] = category
+                glyph_dict[name]['subCategory'] = subCategory
+            else:
+                glyphs_gdef = {name: {}}
+                glyphs_gdef[name]['category'] = category
+                glyphs_gdef[name]['subCategory'] = subCategory
+                glyph_dict.update(glyphs_gdef)
+
+    glyph_data = build_data(glyph_dict)
+    test_data(glyph_dict, glyph_data)
+    logger.info('Loading custom GlyphData.xml')
+    return glyph_data
+
+
 def load_to_ufos(file_or_path, include_instances=False, family_name=None,
-                 debug=False):
+                 debug=False, custom_glyph_xml=None):
     """Load an unpacked .glyphs object to UFO objects."""
 
     if hasattr(file_or_path, 'read'):
@@ -65,13 +110,16 @@ def load_to_ufos(file_or_path, include_instances=False, family_name=None,
     else:
         with open(file_or_path, 'r', encoding='utf-8') as ifile:
             data = load(ifile)
+
+    glyph_data = load_glyph_data(glyphs_source=data, custom_glyph_xml=custom_glyph_xml)
+
     logger.info('Loading to UFOs')
     return to_ufos(data, include_instances=include_instances,
-                   family_name=family_name, debug=debug)
+                   family_name=family_name, debug=debug, glyph_data=glyph_data)
 
 
 def build_masters(filename, master_dir, designspace_instance_dir=None,
-                  family_name=None):
+                  family_name=None, custom_glyph_xml=None):
     """Write and return UFOs from the masters defined in a .glyphs file.
 
     Args:
@@ -80,6 +128,7 @@ def build_masters(filename, master_dir, designspace_instance_dir=None,
             written alongside the master UFOs though no instances will be built.
         family_name: If provided, the master UFOs will be given this name and
             only instances with this name will be included in the designspace.
+        custom_glyph_xml: Path to a custom GlyphData.xml file.
 
     Returns:
         A list of master UFOs, and if designspace_instance_dir is provided, a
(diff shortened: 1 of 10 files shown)


punchcutter:
CLA signed

MetaTools/generate_glyphdata.py
```diff
@@ -242,11 +112,14 @@ def generate_python_source(data, out):
         out.write('\t"%s": %s,\n' % (glyphName, glyphsCat))
     out.write("}\n\n")
 
+    out.write(
+        "DEFAULT_GLYPH_DICT = %s\n\n" % glyphs
```
anthrotype: Hmmm. I believe this PR was merged to hastily.. 

We are basically now storing the GlyphData twice in `glyphdata_generated.py`: once, as the previously existing dictionaries (`PRODUCTION_NAMES`, `IRREGULAR_UNICODE_STRINGS`, etc.) that only have the stuff that cannot be derived from unicodedata, and then again, as a huge dictionary ( `DEFAULT_GLYPH_DICT`) that basically contains the whole GlyphData.xml, as it was read in by ElementTree.

Does that mean @brawer effectively gave up on the idea of storing only the delta, and is now happy with storing the full GlyphData as is? If that's the case, then I think it'd be better (ie. easier to maintain, not much slower to parse) if we just included the original xml file as a package data (possibly fetching it via git submodule), instead of writing that out as a dictionary to a python file.

Also, I don't like that the functions that parse/build a GlyphData are in that residual `util.py` module, instead of where one would expect to find them, ie. the `glyphsLib.glyphdata` module.

Lib/glyphsLib/__init__.py
```diff
@@ -56,22 +57,69 @@ def loads(value):
...
+        custom_glyph_xml: Path to a custom GlyphData.xml file
+    """
+    glyph_dict = data.DEFAULT_GLYPH_DICT.copy()
+    if custom_glyph_xml:
+        custom_glyphs = {}
+        lines = None
+        with open(custom_glyph_xml, 'r') as f:
+            lines = f.read()
+        if lines:
+            custom_glyphs = fetch_all_glyphs(paths=(lines,))
```
anthrotype: this `fetch_all_glyphs` function is no longer "fetching" anything, as in downloading something from a some remote server. And that `paths` argument is now being used to pass on the data itself (the string content), rather than the filesystem location where the containing file is.

Lib/glyphsLib/builder.py
```diff
@@ -649,10 +654,11 @@ def load_glyph(glyph, layer, glyph_data):
     if export is not None:
         glyph.lib[glyphlib_prefix + 'Export'] = export
     production_name = glyph_data.get('production')
+    extra_glyph_info = None
     if production_name is None:
-        glyphinfo = glyphsLib.glyphdata.get_glyph(glyph.name)
-        if glyphinfo:
-            production_name = glyphinfo.production_name
+        extra_glyph_info = glyphdata_generated.DEFAULT_GLYPH_DICT.get(glyph.name)
```
anthrotype: why do we have a `get_glyph` function if we don't use it here to look up the production names from `glyphdata_generated`?

Lib/glyphsLib/glyphdata.py
```diff
@@ -80,6 +80,15 @@ def _get_category(name, unistr, data=glyphdata_generated):
...
     if basename.startswith("uniF9"):
         return ("Letter", "Compatibility")
+
+    glyph_info = glyphdata_generated.DEFAULT_GLYPH_DICT.get(name)
+    if glyph_info is None:
+        glyph_info = glyphdata_generated.DEFAULT_GLYPH_DICT.get(basename)
+    if glyph_info is not None:
+        category, subCategory = glyph_info.get('category'), glyph_info.get('subCategory')
+        if (category, subCategory) != (None, None):
+            return (category, subCategory)
```
anthrotype: this is confusing. `data` already defaults to `glyphdata_generated`; if one gives a `data` value, then this function should treat that as a replacement for the default value.. But here you still looking up the original `glyphdata_generated` nevertheless.

Lib/glyphsLib/util.py
```diff
@@ -68,3 +109,104 @@ def clear_data(data):
                 i += 1
         return data
     return True
+
+
+def fetch_all_glyphs(paths=()):
+    glyphs = {}
+    for fileobj in paths:
+        for glyph in etree.fromstring(fileobj).findall("glyph"):
```
anthrotype: `fileobj` is definitely not a file object; it's a string, and that's why `fromstring` method can do its job.

MetaTools/generate_glyphdata.py
anthrotype: not that file size matters to me -- but the current `glyphdata_generated.py` is now 8.7 MB, whereas the sum of the original `GlyphData.xml` and `GlyphData_Ideographs.xml` files is 5.9 MB.

Lib/glyphsLib/glyphdata.py
punchcutter: Definitely. No idea what I was thinking there. Should just be data.

Lib/glyphsLib/builder.py
```diff
@@ -76,7 +77,7 @@
     '': 5}
 
 
-def to_ufos(data, include_instances=False, family_name=None, debug=False):
+def to_ufos(data, include_instances=False, family_name=None, debug=False, glyph_data=None):
```
anthrotype: This is wrong. Why are you setting `glyph_data` default to `None` here, if a value is required for `build_gdef` and `get_glyph` to be able to work? Setting it as None implies it's optional, but it isn't. That's why you also changed all the calls to `to_ufos` in the `builder_test` to explicitly (superfluously) pass a `glyph_data` argument... Maybe you wanted to set the default value to `glyph_data=glyphdata_generated`.

Lib/glyphsLib/util.py
punchcutter: Yes, that was dumb. It was called filename before which was confusing because it also wasn't a filename.

anthrotype:
My vote is for a complete revert, followed by a new, cleaned-up patch.

I'm ever more convinced we should drop all the custom code and simply parse the original GlyphData.xml at runtime. We'll spare lots of headache and we'd match the Glyphs.app behaviour.

Lib/glyphsLib/glyphdata.py
anthrotype: @punchcutter hey! I hope you don't get my comments bad. I really appreciate your efforts and contribution! It's just that I've been trying to keep things clear and simple for my future self that will have to maintain all this.. :)

punchcutter:
Sounds good to me. I was expecting your feedback and keen eye before any merging. I know it's not ideal, but I wanted to get the ball rolling somehow instead of standing by doing nothing.

anthrotype:
Thank you @punchcutter. Let's wait for @brawer to comment before reverting. Better that we are all aligned before going in a different direction.

behdad:
So, what's the status here?

anthrotype:
I reverted the #150 merge with 899246d, then added back some of the features with #154: i.e. zeroing the width nonspacing marks and reading category/subCategory overrides stored inside the .glyphs source file itself.

The question of whether/how to handle custom GlyphData.xml is still undecided, although maybe it's less of an urgency since users can store production names, category and subCategory inside the .glyphs source file itself.

The cleanest/simplest approach, in my point of view, is to store the default GlyphData.xml as iti is inside the glyphsLib package, load it at runtime (fast with cElementTree), and optionally update its content with a user-provided GlyphData.xml file.

anthrotype:
I haven't tagged a glyphsLib release with this yet, as I was trying to fix another issue (#119) -- but maybe that can wait, as we need more clarifications from Georg as to how to parse angle-bracket enclosed  binary data.

behdad:
> The cleanest/simplest approach, in my point of view, is to store the default GlyphData.xml as it is inside the glyphsLib package, load it at runtime (fast with cElementTree), and optionally update its content with a user-provided GlyphData.xml file.

That sounds good to me.  So, not pass it around all the time?  Something like `fontTools.unicode.setUnicodeData()`?

anthrotype:
yeah, in that case it would be global for all the life of the current python process

PR #154, opened 2017-03-21
use category/subCategory overrides; zero nonspacing marks width...

### LINKED ISSUE #115: Set width of glyphs with GDEF class 3 (mark) to zero?
James opened a similar issue on ufo2ft repo:
https://github.com/googlei18n/ufo2ft/issues/55

> I noticed that in fonts output by Glyphs, any glyph with GDEF class 3 has a width of zero. Should ufo2ft do the same? If so, we have to go back and change widths the compiled OTF/TTFs after the GDEF table has been compiled by feaLib.

I think it should be the job of glyphsLib to set the width of such mark glyphs to zero when building the UFO.

anthrotype:
We decided to revert #150 as it needed some clean up.
However, a couple of things from there could easily be incorporated.

1) if the glyph contain custom 'category' or 'subCategory' attributes embedded in the .glyphs source file itself (cf. CMD+ALT+i panel), then use that instead of global values from GlyphData.xml. Additionally, store these inside the GLIF lib under a corresponding private key. This way we can roundtrip them when we go back from UFO to .glyphs source.

2) Set the width of glyphs with category "Mark" and subCategory "Nonspacing" to 0, like Glyphs.app does upon export (fixes #115)

Finally, only store production names in the UFO lib when they differ from original glyph names. Currently we are storing even superfluous stuff like 'A' -> 'A'...

I will add some tests tomorrow.

### DIFF
--- a/Lib/glyphsLib/builder.py
+++ b/Lib/glyphsLib/builder.py
@@ -35,6 +35,7 @@ logger = logging.getLogger(__name__)
 
 PUBLIC_PREFIX = 'public.'
 GLYPHS_PREFIX = 'com.schriftgestaltung.'
+GLYPHLIB_PREFIX = GLYPHS_PREFIX + 'Glyphs.'
 ROBOFONT_PREFIX = 'com.typemytype.robofont.'
 
 GLYPHS_COLORS = (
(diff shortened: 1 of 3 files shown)


marekjez86:
@anthrotype : thank you

PR #155, opened 2017-03-22
add ufo2ft filters key in ufo.lib

zjusbo:
This is the first step of solving issue https://github.com/googlei18n/fontmake/issues/275 , following Cosimo's advice.

Keep the Filters in ufo.lib['com.github.googlei18n.ufo2ft.filters'] and let ufo2ft handle it.

### DIFF
--- a/Lib/glyphsLib/builder.py
+++ b/Lib/glyphsLib/builder.py
@@ -391,8 +391,8 @@ def parse_glyphs_filter(filter_str):
     """
     elements = filter_str.split(';')
 
-    if len(elements) == 0 or elements[0] == '':
-        logger.warn('Failed to parse glyphs filter, expecting a filter name: \
+    if elements[0] == '':
+        logger.error('Failed to parse glyphs filter, expecting a filter name: \
              %s', filter_str)
         return None
 
--- a/Lib/glyphsLib/builder.py
+++ b/Lib/glyphsLib/builder.py
@@ -22,7 +22,7 @@ import logging
 import re
 
 from glyphsLib.anchors import propagate_font_anchors
-from glyphsLib.util import clear_data
+from glyphsLib.util import clear_data, cast_to_number_or_bool
 import glyphsLib.glyphdata
 
 __all__ = [
@@ -400,20 +400,19 @@ def parse_glyphs_filter(filter_str):
     result['name'] = elements[0]
     result['args'] = []
     result['kwargs'] = {}
-    result['include'] = []
-    result['exclude'] = []
-    for elem in elements[1:]:
+    for idx, elem in enumerate(elements[1:]):
         if ':' in elem:
             # Key value pair
             key, value = elem.split(':', 1)
-            if key.lower() == 'include':
-                result['include'] = re.split('[ ,]+', value)
-            elif key.lower() == 'exclude':
-                result['exclude'] = re.split('[ ,]+', value)
+            if key.lower() in ['include', 'exclude']:
+                if idx != len(elements[1:]) - 1:
+                    logger.error('{} can only present as the last argument in the filter. {} is ignored.'.format(key, elem))
+                    continue
+                result[key.lower()] = re.split('[ ,]+', value)
             else:
-                result['kwargs'][key] = value
+                result['kwargs'][key] = cast_to_number_or_bool(value)
         else:
-            result['args'].append(elem)
+            result['args'].append(cast_to_number_or_bool(elem))
     return result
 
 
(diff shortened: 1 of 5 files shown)


anthrotype:
hey Bo, I was wondering. Instead of passing on the same string value that Glyphs.app uses for its Filter custom parameters, e.g. "Transformations;OffsetX:40;OffsetY:60;include:uni0334,uni0335"... wouldn't it be better if we specify as input format for a ufo2ft filter something that's a bit structured?

We could take advantage of the fact that the UFO lib is a property list, so instead of a string that needs to parsed again inside ufo2ft, we would store inside `ufo.lib['com.github.googlei18n.ufo2ft.filters']` a list of dictionaries, one for each filter, each containing a "name" (the name of the filter), an "args" list (for positional parameters), a "kwargs" dictionary (for optional, named parameters), and finally "include" and "exclude" lists of strings.
 
WDYT? Does that make sense?

zjusbo:
Haha. I was thinking of it when I'm writing a parser in ufo2ft to parse these strings. I think it's a good idea to parse it early in glyphsLib. Let me move my parser logic into glyphsLib. Thanks for the advice!

Lib/glyphsLib/builder.py
```diff
@@ -371,6 +374,49 @@ def set_custom_params(ufo, parsed=None, data=None, misc_keys=(), non_info=()):
...
+        Args:
+            filter_str - a string of glyphs app filter
+
+        Return:
+            A dictionary contains the structured filter.
+            Return None if parse failed.
+    """
+    elements = filter_str.split(';')
+
+    if len(elements) == 0 or elements[0] == '':
```
anthrotype: if `filter_str ` is a string (even an empty one), and you call `.split()` method on it, the resulting `len(elements)` can never be 0; it'll always be at least 1 (the list would contain the original string, if the separator is not present; or more than one substring).

Lib/glyphsLib/builder.py
```diff
@@ -371,6 +374,49 @@ def set_custom_params(ufo, parsed=None, data=None, misc_keys=(), non_info=()):
...
+        logger.warn('Failed to parse glyphs filter, expecting a filter name: \
```
anthrotype: I'd elevate this log message level to ERROR instead of WARNING, even though we pass and ignore the filter.

setup.py
```diff
@@ -41,6 +41,7 @@
     setup_requires=pytest_runner,
     tests_require=[
         'pytest>=2.8',
+        'mock>=2.0.0',
```
anthrotype: you also need to add mock to tox.ini file, which is where the tox test runner is getting the test requirements.
Add a line after pytest in there.

anthrotype:
```diff
diff --git a/tox.ini b/tox.ini
index 57b363c..9b0e7ff 100644
--- a/tox.ini
+++ b/tox.ini
@@ -4,6 +4,7 @@ envlist = py27, py35
 [testenv]
 deps =
        pytest
+       mock>=2.0.0
        -rrequirements.txt
 commands =
        pytest {posargs}
```

Lib/glyphsLib/builder.py
zjusbo: Updated. So what's our principle of whether to use error or warn? I used warn here since the font can still be compiled though with some defects.

anthrotype: The only possibility of elements[0] being an empty string is that the value of the custom parameter is empty or it starts with a semicolon; in both cases it's invalid and it should not pass unnoticed. The user should fix it in the source file. But I don't mind if you decide to keep it as warning.

anthrotype:
Last thing, we need to decide what to do when both include and exclude are passed. The handbook is not very clear about that, but if you try to do that in glyphs, it appears that the last one of exclude or include key/value pairs is the one that wins. The previous ones seem to be ignored by glyphs.app.

If you confirm that is the case, then we should maybe do the same and only accept either include or exclude, but not both (which could be conceivable though not very logic), and print a warning to the user to make clear we will apply only the last one among include or exclude statements.

schriftgestalt:
The include/exclude can’t be active at the same time as only on of the two can be the last argument in the string.

anthrotype:
Ah, and another thing: we need to pre-parse integers and floats from the custom parameter string value, instead of only writing strings to the lib plist.
That's what the Filter plugin classes would receive in the Glyphs.app python API. Ufo2ft should not do that, it should be glyphsLib that e.g. should store a int as 1, not as '1'

anthrotype:
Thanks Georg. Exactly, so we need to check for that here and maybe warn in case both are present.

zjusbo:
Thanks Cosimo and Georg. Fix coming.

PR #159, opened 2017-04-08
Use weightClass and interpolationWeight as they appear in actual fonts

brawer:
No change in behavior yet. This change merely updates the test input,
so that the unit tests reflect what the compiler is actually seeing
in its input sources.


brawer:
Regarding width, I’ll look into that later.

PR #161, opened 2017-04-10
Write variation axes into designspace documents

brawer:
The output files now contain the information for building `avar` tables,
and for building `fvar` tables with correct value ranges.

### DIFF
--- a/Lib/glyphsLib/interpolation.py
+++ b/Lib/glyphsLib/interpolation.py
@@ -16,6 +16,7 @@
 from __future__ import (print_function, division, absolute_import,
                         unicode_literals)
 
+from collections import OrderedDict, namedtuple
 import logging
 import os
 
@@ -68,9 +69,13 @@ def build_designspace(masters, master_dir, out_dir, instance_data):
     tmp_path = os.path.join(master_dir, 'tmp.designspace')
     writer = DesignSpaceDocumentWriter(tmp_path)
 
-    base_family, base_style = add_masters_to_writer(writer, masters)
+    instances = list(filter(is_instance_active, instance_data.get('data', [])))
+    axes = get_axes(masters, instances)
+    write_axes(axes, writer)
+    base_family, base_style = add_masters_to_writer(masters, axes, writer)
+    base_family = instance_data.get('defaultFamilyName', base_family)
     instance_files = add_instances_to_writer(
-        writer, base_family, instance_data, out_dir)
+        writer, base_family, axes, instances, out_dir)
 
     basename = '%s%s.designspace' % (
         base_family, ('-' + base_style) if base_style else '')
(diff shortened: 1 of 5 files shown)


anthrotype:
hm, wonder why the python 3 tests are failing while the python 2.7 ones are passing..
maybe an issue in mutatormath?

Lib/glyphsLib/interpolation.py
```diff
@@ -68,9 +69,13 @@ def build_designspace(masters, master_dir, out_dir, instance_data):
     tmp_path = os.path.join(master_dir, 'tmp.designspace')
     writer = DesignSpaceDocumentWriter(tmp_path)
 
-    base_family, base_style = add_masters_to_writer(writer, masters)
+    instances = filter(is_instance_active, instance_data.get('data', []))
```
anthrotype: oh, it's because `filter` returns an iterator in python 3, not a list, but you are reusing `instances` later in `add_instances_to_writer`; however by that time the iterator has already been consumed by `get_axes` and it's empty. You should probably do `list(filter(...))` so you can iterate more than once

brawer:
Thank you! Fixed.

Lib/glyphsLib/interpolation.py
```diff
@@ -79,24 +84,89 @@ def build_designspace(masters, master_dir, out_dir, instance_data):
...
+def get_axes(masters, instances):
+    # According to Georg Seifert, Glyphs 3 will have a better model
+    # for describing variation axes.  The plan is to store the axis
+    # information globally in the Glyphs file. In addition to actual
+    # variation axes, this new structure will probably also contain
+    # stylistic information for design axes that are not variable but
+    # should still be stored into the OpenType STAT table.
+    #
+    # We currently take the minima and maxima from the instances, and
+    # have hard-coded the default value for each axis.  We could be
```
anthrotype: I don't like that we are hardcoding the default axis value here.
If the weight always has a default of 400, and my font has no master with that default value because, say, it only has two Light and Bold as masters, then `varLib`  would raise an `AssertionError`, since it's expecting to find the "base" master, i.e. one which has default locations (0 in the normalized space) along all the axes:

https://github.com/fonttools/fonttools/blob/932b938d7d80fe9182730dbc29e51fe6af3c7de5/Lib/fontTools/varLib/__init__.py#L535-L542

In theory varLib could generate such a base master given the default values in `<axes>`, but that's not the point here. We need a way to specify these defaults, and we can't wait for Glyphs 3 to be out.

Can't we define a private custom parameter (named I don't now) to store this default locations for the three weight/width and custom axes?

/cc @brawer @behdad @schriftgestalt

behdad: The default should come from whatever master gets its `info` attribute set to True.  @brawer can you do that?

In the future, Georg said he'll add a custom property for base master.

PR #164, opened 2017-04-11
Avoid duplicates in designspace axis maps for multi-axis fonts

brawer:
Before this change, glyphsLib would produce a designspace for
NotoSansArabic-MM.glyphs that contained duplicate entries in the axis
maps. This change uniques the entries, so they're listed only once.
After this change, the generated designspace for NotoSansArabic looks
reasonable.

### DIFF
--- a/Lib/glyphsLib/interpolation.py
+++ b/Lib/glyphsLib/interpolation.py
@@ -127,7 +127,7 @@ def get_axes(masters, instances):
                         userLoc = float(param.get('value', DEFAULT_LOC))
                         break
                 mapping.append((userLoc, interpolLoc))
-            mapping.sort()
+            mapping = sorted(set(mapping))  # avoid duplicates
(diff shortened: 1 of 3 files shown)


Lib/glyphsLib/interpolation.py
```diff
@@ -127,7 +127,7 @@ def get_axes(masters, instances):
                         userLoc = float(param.get('value', DEFAULT_LOC))
                         break
                 mapping.append((userLoc, interpolLoc))
-            mapping.sort()
+            mapping = list(sorted(set(mapping)))  # avoid duplicates
```
anthrotype: no need to do `list(sorted(...))`, as `sorted` returns a new list in both python2 and 3.

brawer: Thank you! Fixed.

PR #166, opened 2017-04-11
Emit locations of output designspaces in deterministic order

brawer:
The non-determinism was occasionally breaking unit tests. It was
caused by MutatorMath's DesignSpaceDocumentWriter, which iterates over
the location dictionary without sorting. The proper fix would of
course be in MutatorMath, but we'll switch to DesignSpaceDocument as
soon as that has been integrated into fonttools. So we just stop the
bleeding by passing an OrderedDict to MutatorMath.

Fixes https://github.com/googlei18n/glyphsLib/issues/165.


brawer:
Yes, filed https://github.com/LettError/designSpaceDocument/issues/10 for this.

PR #171, opened 2017-04-14
[interpolation] use 0 as default value for custom axis

anthrotype:
Fixes https://github.com/googlei18n/glyphsLib/issues/169


anthrotype:
You can verify that by setting the third "custom" axis value in either the masters or the instances to 0, the `customValue` and the `interpolationCustom` properties are not written to the .glyphs source file.
Hence the default seems to be 100 for weight and width, but 0 for custom axis.

schriftgestalt:
@anthrotype is right.

anthrotype:
Thanks Georg

PR #173, opened 2017-04-24
fix regression in detecting active instances

anthrotype:
Fixes a regression introduced with commit https://github.com/googlei18n/glyphsLib/blob/d9d1da4e/Lib/glyphsLib/interpolation.py#L146

The 'exports' parameter was being interpreted as a string instead of as a boolean, hence the following test `if '0': ...` would always return `True`.
Notice how in line 146, the cast to `int()` was not used for the 'exports' parameter, but only for the 'active' one.

Because of this, all the instances were always generated no matter what their 'exports' value...

The tests in `interpolation_test.py` did not catch this bug, because in there we were passing an integer instead of a string as the 'exports' parameter value.

The 'exports' parameter is now explicitly being cast to boolean (db826fe).

### DIFF
--- a/Lib/glyphsLib/casting.py
+++ b/Lib/glyphsLib/casting.py
@@ -499,6 +499,7 @@ _TYPE_STRUCTURE = {
     },
     'instances': {
         'active': truthy,  # undocumented
+        'exports': truthy,  # undocumented
         'customParameters': custom_params,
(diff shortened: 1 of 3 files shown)


moyogo:
LGTM

brawer:
LGTM

PR #174, opened 2017-04-26
Refactor how glyphsLib finds the "Regular" master

brawer:
In MutatorMath, one of the masters is used as the base for copying
default font information, and glyphsLib tries to use the "Regular"
master for this.

Before this change, the logic for finding the "regular" master was
interleaved with writing out the masters into the output designspace
file for MutatorMath. After this change, the logic is in a separate
function.

No change in functionality, just refactoring.

### DIFF
--- a/Lib/glyphsLib/interpolation.py
+++ b/Lib/glyphsLib/interpolation.py
@@ -78,15 +78,15 @@ def build_designspace(masters, master_dir, out_dir, instance_data):
     writer = DesignSpaceDocumentWriter(tmp_path)
 
     instances = list(filter(is_instance_active, instance_data.get('data', [])))
-    axes = get_axes(masters, instances)
+    regular = find_regular_master(masters)
+    axes = get_axes(masters, regular, instances)
     write_axes(axes, writer)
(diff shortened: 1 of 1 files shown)


Lib/glyphsLib/interpolation.py
```diff
@@ -203,27 +208,12 @@ def add_masters_to_writer(ufos, axes, writer):
         for axis in axes:
             location[axis] = font.lib.get(
                 GLYPHS_PREFIX + axis + 'Value', DEFAULT_LOCS[axis])
-        master_data.append((font.path, family, style, location))
-
-    # pick a master to copy info, features, and groups from, trying to find the
-    # master with a base style shared between all masters (or just Regular) and
-    # defaulting to the first master if nothing is found
```
anthrotype: would be good to keep this comment in the `find_regular_master` function, to summarize how the base master is currently determined.

brawer: Done in https://github.com/googlei18n/glyphsLib/commit/21293067128ae87be7b7664a9302a46b13a368de.

PR #175, opened 2017-04-26
Take default axis values from "regular" interpolation master

brawer:
Fixes https://github.com/googlei18n/glyphsLib/issues/170.


brawer:
Regarding the newly added glyphsLib parameter, let’s do that in a separate change; see https://github.com/googlei18n/glyphsLib/issues/176.

PR #177, opened 2017-04-26
Support “Variation font origin” parameter

brawer:
Fixes https://github.com/googlei18n/glyphsLib/issues/176.

### DIFF
--- a/Lib/glyphsLib/interpolation.py
+++ b/Lib/glyphsLib/interpolation.py
@@ -78,7 +78,9 @@ def build_designspace(masters, master_dir, out_dir, instance_data):
     writer = DesignSpaceDocumentWriter(tmp_path)
 
     instances = list(filter(is_instance_active, instance_data.get('data', [])))
-    regular = find_regular_master(masters)
+    regular = find_regular_master(
+        masters=masters,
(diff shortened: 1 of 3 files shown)


Lib/glyphsLib/interpolation.py
```diff
@@ -177,13 +179,16 @@ def write_axes(axes, writer):
...
-def find_regular_master(masters):
+def find_regular_master(masters, regularName=None):
     """Find the "regular" master among the master UFOs.
     """
     assert len(masters) > 0
     base_family = masters[0].info.familyName
     assert all(m.info.familyName == base_family for m in masters), \
         'Masters must all have same family'
+    for font in masters:
+        if font.info.styleName == regularName:
```
anthrotype: would this mean that if a `font.info.styleName` happens to be None, thus it is equal to the default `regularName=None`, it would be chosen as the regular master?

brawer: https://github.com/googlei18n/glyphsLib/commit/8f1b44cb634e7a92f2a59ac00cd76ca3c0159372

anthrotype: thanks!

PR #187, opened 2017-05-18
Sort generated UFO kerning groups by glyphOrder if present

anthrotype:
Currently we populate the UFO glyphs in the order in which glyphs are found in the .glyphs source. For the kerning group, we append each new glyph to the respective `leftKerningGroup` and `rightKerningGroup`. 

However, UFO groups (both kerning and not) are stored as arrays of strings in the groups.plist, whereas kerning groups are simply attributes of the glyph objects in Glyphs. Thus, the original ordering of the glyphs from the UFO kerning groups is lost if these were imported in a glyphs file.

Although such order is not relevant to make a kern feature, it's important if one wishes to roundtrip UFO groups.plist -> Glyphs' {left,right}KerningGroup -> UFO groups.plist, and avoid producing spurious changes in the resulting diff.

By using the glyphOrder, we can try to ensure a more predictable order -- provided that the members of the UFO kerning groups also were sorted by 'public.glyphOrder' to begin with.

This patch makes so that when a 'glyphOrder' custom parameter is present, we load the the glyphs data (including left/right kerning group membership) following that order, otherwise we keep using the glyphs data's intrinsic order.

### DIFF
--- a/Lib/glyphsLib/__init__.py
+++ b/Lib/glyphsLib/__init__.py
@@ -28,7 +28,7 @@ from glyphsLib.parser import Parser
 from glyphsLib.util import write_ufo
 
 
-__version__ = "1.7.5"
+__version__ = "1.7.6.dev0"
 
 __all__ = [
     "build_masters", "build_instances", "load_to_ufos", "load", "loads",
--- a/Lib/glyphsLib/builder.py
+++ b/Lib/glyphsLib/builder.py
@@ -166,15 +166,26 @@ def to_ufos(data, include_instances=False, family_name=None, debug=False):
         glyph_order = []
     sorted_glyphset = set(glyph_order)
 
+    glyphs = {}
     for glyph in data['glyphs']:
-        add_glyph_to_groups(kerning_groups, glyph)
-
-        glyph_name = glyph.pop('glyphname')
+        glyph_name = glyph['glyphname']
         if glyph_name not in sorted_glyphset:
             # glyphs not listed in the 'glyphOrder' custom parameter but still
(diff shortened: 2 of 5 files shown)


anthrotype:
actually...... I have just now realized that after one sets the `glyphOrder` custom parameter and saves the glyphs source file, Glyphs.app will rewrite the glyphs in the file in the same glyphOrder.
Thus, loading the glyphs in the order in which they are written in the source data (which is what we currently do) *is* the same thing as using the `glyphOrder` custom parameter, as latter determines the former.

well, this makes the patch practically useless.
(I wish I had known this before 😁 )

@schriftgestalt can you confirm this?

anthrotype:
the current patch would only be needed if the glyphOrder custom parameter does not match the order in which the glyphs are written out in the .glyphs source file; but this doesn't seem to be the case with Glyphs.app, and so far the latter is the only existing editor of *.glyphs file format.

schriftgestalt:
I was wondering why using the glyphOrder would make a difference. 

there is currently no way to preserve the ordering of the kerning classes. But Glyphs will normalise it once and the next time your round trip, there won’t be changes any more. And the same change would happen if you would "roundtrip" through an .otf ;)

anthrotype:
yeah, in fact it does not make a difference, because we are already loading the glyphs (and populating the UFO kerning groups) in the order in which the glyphs are written in the .glyphs file, and this order happens to match the (optional) `glyphOrder` custom parameter, when present.
So it is indeed not needed, was just a misunderstanding on my part. I thought the two orders were independent, and changing the custom parameter did not also reorder the way glyphs are stored in the source file.

PR #189, opened 2017-05-26
Don’t choke on QCURVE

khaledhosny:
I’m not entirely sure if this is correct, but it fixes https://github.com/googlei18n/glyphsLib/issues/184 and fontmake can now process the font correctly.


anthrotype:
thanks Khaled. Would be nice to also have a test.
But if it's too much work, we can add it later.

khaledhosny:
I added a couple of tests.

PR #198, opened 2017-07-14
Remove shebangs from non-executable modules

athos-ribeiro:
Since these files are not supposed to be executable, there is not point
in having shebangs there. Also, since they just point to the system
python binary, they end up pointing to the wrong python version if both
python2 an python3 packages are installed for one of the python versions.


athos-ribeiro:
I am packaging glyphsLib into Fedora and these shebangs would point to the wrong python interpreter for the python3 package (we usually generate both python2 and python3 packages from the same sources when upstream supports both).

For now, I am applying a patch downstream to remove these shebangs

anthrotype:
thanks!

PR #199, opened 2017-07-31
fix issues when providing custom family_name

anthrotype:
Fixes https://github.com/googlei18n/fontmake/issues/314 and https://github.com/googlei18n/fontmake/issues/337

See commit messages for info.


anthrotype:
@googlebot I signed! 

(twice, actually...)

PR #202, opened 2017-08-07
Fix styleMapFamilyName in generated UFOs

### LINKED ISSUE #200: styleMapFamilyName incorrectly if style name contains "Bold"
as reported in https://github.com/googlei18n/fontmake/issues/209#issuecomment-320113351

> Generated SemiBold and ExtraBold fonts seem to have bad names, at least based on the fonts in noto-fonts-alpha...ttf.
> For example, NotoSansTamilUI-SemiCondensedExtraBold.ttf has the name "NotoSans Tamil UI SemiCondensed Extra". The "Bold" is missing.

The current styleMapFamilyName generation logic is here:
https://github.com/googlei18n/glyphsLib/blob/e3bae79/Lib/glyphsLib/builder.py#L353-L362

anthrotype:
Strip "Bold" from `styleMapFamilyName` only if it's a whole word, not if it's is part of a compound word, e.g. "ExtraBold".

Fixes #200


anthrotype:
I decided to get rid of the catch-all `set_redundant_data` function that was used for both the master UFOs and post-mutatorMath instance UFOs to derive OS/2 weight/width classes and style linking names from the style name.

In Glyphs.app, masters don't have a weight/width class in the same sense that instances have; their `width` or `weight` (or `custom`) properties are just user labels to display the masters in the UI. Similarly, masters do not have style linking properties attached. It's true that they gain an " Italic" string in their name when italic angle is `!= 0` but that's again to distinguish from upright masters. Masters as such are not meant to be exported to binary OTF or TTF; only instances are meant to be exported, and thus have `weightClass` and `widthClass` properties, as well as corresponding custom parameters overriding those, and also have `isBold`, `isItalic` and `linkStyle` to define the nameID 1 and 2 for RBBI-only apps.
It doesn't make sense to make up those properties in the generated master UFOs, nor it make sense that we ignore the properties that Glyphs.app provides to determine the corresponding weight/width and style linking of the *instance* UFOs.

So with this patch we no longer set the `openTypeOS2WeightClass`, `openTypeOS2WidthClass`, `styleMapFamilyName` and `styleMapStyleName` on the UFO masters, but we only do for the instances.

For the `styleMap*` names, we pass these as arguments to mutatorMath's designspace document writer so they will be stored in the generated designspace and used when building the instance UFOs.

For the OS/2 weight and width classes, since designspace format does not allow to define those for instances (maybe once we add support for custom instance `lib` element?), we currently set them on the instances returned by mutatorMath in the `apply_instance_data` function (called by fontmake after interpolation). The default values (mapping the weight/width standard names from the drop-down menus to OS/2 codes) can also be overridden through custom parameters. However, adding such `weightClass` and `widthClass` custom parameters is no longer needed if the desired value is equal to the one that Glyphs.app would assign an instance given the latter's `weightClass` and `widthClass` property. The latter properties were previously ignored, forcing users to add redundant custom parameters to fix up the OS/2 weight and width.

Related issues:
https://github.com/googlei18n/glyphsLib/issues/139#issuecomment-297962171
https://github.com/googlei18n/fontmake/issues/338#issuecomment-318889989

PR #203, opened 2017-08-08
Classes and GlyphsApp Python API instead of data dict

moyogo:


anthrotype:
Who's coming to ATypI Montreal next week? Maybe we could spare some time during the conference and work together on the remaining issues of the `classtree` branch...
/cc @schriftgestalt @yanone @moyogo

schriftgestalt:
I'll be there.

anthrotype:
I only scratched the surface today. I tried (unsuccessfully) to resolve the merge conflict with current master. The branch was mergeable (well, at least from a mere git point of view) with master at the `v1.8.0` tag.

The commits from `v1.8.0` to current `HEAD` were meant to address the issue of roundtripping .glyphs files without unwanted changes in quoted strings (#209) by making the unescape/escape optional in parser/writer.
I'm not sure how this fits into the way the parser/writer work in the classtree branch where there's no intermediate `casting`  step that turns collections of strings into collections of int, float, bool, etc.; there, the parser and writer go directly from strings to objects and back; disabling the (un)escaping while parsing/writing doesn't make much sense in this context.

Another solution to this deadlock could be to revert master to the state it was in v1.8.0, merge classtree (after the review/cleanup is completed), and then replay the changes that have occurred on master since v1.8.0, adapting these to the new classes etc.

If the parser/writer in classtree branch could guarantee that they can roundtrip (load/dump) any .glyphs file as generated from Glyphs.app, then we may no longer need to `unescape` and `escape` options.

schriftgestalt:
As I said, the classTree has a new writer. And it might need different solutions, if there are any needed. Are the files that produced the change available for testing (unit tests)? Then I would check the round tripping, after the merge...

anthrotype:
@schriftgestalt I used the following script with a bunch of fonts from googlei18n/noto-source
https://gist.github.com/anthrotype/3f4c9b3e04c22ddcfda196b970de8549

The script takes one or more glyphs files, loads and dump them again, and prints the diff to the console.

These are some of the differences I noticed after roundtripping from the classtree branch:

- A glyph's `layers` sometimes are written in a different order from the input; the `GlyphLayerProxy.values` method uses `sorted` on the underlying `_layers.values()`, but the latter is an `OrderedDict` so it doesn't need reordering; removing the call to `sorted` drastically reduces the changes;
- `instanceInterpolations` is a `dict`; the order sometimes changes when roundtripping; `userData` is also `dict` though I haven't tested yet if that also produces roundtrip differences; perhaps an `OrderedDict` should be used wherever a `dict` is (I don't know);
- `instanceInterpolations` contain float values with different precision than in the input (input has 6 decimal positions, output is rounded to 5 positions)
- `unicode` values are written as quoted strings (they are not quoted in the input);
- `.appVersion` is written without quotes (it's quoted in the input);
- `panose` values are written without quotes (quoted numbers in the input);

The list might not be exhaustive and I encourage anyone the script on their .glyphs files and report other differences they encounter. Thanks

schriftgestalt:
thanks about the info. I have a look tomorrow.

schriftgestalt:
I’m not sure what is wrong with the Travis build. 
But the changes make sure that the round trip will produce the same result than Glyphs 2.5 (a beta will be released soon).

anthrotype:
What is the role of the separate `types.py` module? I don't see the difference conceptually between `classes.py` and `types.py`.

Also, why do you use lowercase names for classes in types.py? That's usually for Python built-in types, like int or float, not for user-defined ones which are typically in CamelCase.

schriftgestalt:
The stuff in types.py defines simple base types. The could live in classes.py or init.py, too. 
but those are not really related with the GS-classes so I thought they could have their own file.

anthrotype:
@schriftgestalt ok, thanks for the clarification. It's just that sometimes "class" and "type" are used interchangeably and the distinction between the two terms is not well defined. The lowercase names look odd though, I'd prefer to stick to CamelCase for user-defined classes.

schriftgestalt:
I don't have a strong opinion about this.

anthrotype:
I'm finally going merge this, so we can proceed with the clean-ups directly from a single master branch.
It's ok if something breaks, as users are encouraged to install the latest stable version from PyPI.
There's still more work to do before we can release 2.0.
But I'd like to thank all who have been working in the past months on this huge PR -- despite me! ;)

schriftgestalt:
Many thanks for me too. Especially the last few days where quite amazing. New commits came in quicker than I could respond.

PR #210, opened 2017-08-26
Export 'store' api to write glyphs files from data.

dougfelt:
Also tweaks writing to leave tabs in strings unescaped, which is what
Glyphs appears to do.


anthrotype:
#212 supersedes this

Issue #209, opened 2017-08-23
glyphsLib treats some quoted values as numeric


dougfelt:
The glyphsLib parsing/casting code treats some values as numeric although they are stored in (at least some versions of, e.g. appVersion 998) .glyphs files as quoted strings.  These are:

appVersion
panose values in list
weightClass value

Believing they are numeric, glyphsLib writes these values out without quotes.

This difference in representation means glyphsLib cannot read and then write files without modification.  When using glyphsLib to make bulk changes to .glyphs files, this leads to extraneous changes in the diffs.  Glyphs appears to read these values in ok when unquoted.

On the other hand, it seems like panose and weightClass (if not appVersion) are indeed numeric.  It seems consumers of the parsed data returned by glyphsLib should be able to continue to expect them to be numbers and not strings. So it's not clear glyphsLib is at fault here.

In addition, glyphsLib represents all character data in strings under 0x20 as octal escape sequences.  This causes tabs in strings to be converted to the sequence "\011".  Perhaps glyphsLib should make an exception for tab.

anthrotype:
Let me summarise what I've understood so far from the discussion at https://github.com/googlei18n/noto-source/pull/85.

It would be great if Georg (@schriftgestalt) could find some spare time to comment on each of these points.

There seems to be two kinds of custom parameters: the built-in ones which are supported by Glyphs.app and whose type (whether string or numeric) is known, and arbitrary ones that are user-defined and, because their type is not known in advance, are implicitly treated as strings.

For example, `xHeight` is not one of the supported custom parameters; it's actually a public property of font masters, and the type is numeric (unquoted).

```
xHeight = 500;
```

The Noto fonts however sometimes store `xHeight` as an instance custom parameter in addition to the homonymous `xHeight` master property. In the case of the instance custom parameter, even though its value only contains decimal numbers, Glyphs.app stores it as a quoted string: that's because all the non-supported custom parameters' values are treated as strings; and in order to make it clear that their type is string and not integer or real, Glyphs.app quotes them when they look like a number. E.g.:

```
name = xHeight;
value = "530";
}
```

Actually, it's sufficient that such unsupported custom parameters values only _start with_ a digit for them to be wrapped in quotes. They are not quoted if they only comprise letters, or they contain digits but not in the start position.

```
{
name = all_letters_unquoted;
value = abc;
},
{
name = starts_with_digit_quoted;
value = "1abc";
},
{
name = contains_digit_unquoted;
value = abc1;
}
[...]

PR #212, opened 2017-08-30
Fix parse 2

dougfelt:


Lib/glyphsLib/parser.py
```diff
@@ -137,21 +141,25 @@ def _unescape_fn(m):
...
+        double quotes.  Also convert escapes to unicode."""
+
+        if text[0] == '"':
+            assert text[-1] == '"'
+            text = text[1:-1].replace('\\"', '"')
+        return Parser._unescape_re.sub(Parser._unescape_fn, text)
+
+
+    def _trim_value(self, value):
+        return Parser.unescape_text(value) if self.unescape else value
```
anthrotype: here you may use `self.unescape_text` instead of `Parser.unescape_text` (e.g. a subclass could potentially override the static method).

Lib/glyphsLib/parser.py
```diff
@@ -278,9 +288,19 @@ def dumps(obj):
     dump(obj, fp, **kwargs)
     return fp.getvalue()
 
-def main(args=None):
-    for arg in args:
-        dump(load(open(arg, 'r', encoding='utf-8')), sys.stdout)
+
+def _load_dump(filenames):
```
anthrotype: this `_load_dump` function seems unused

anthrotype:
@dougfelt I've been away last week and need to catch up with things. 
Is this PR related to #209 and https://github.com/googlei18n/noto-source/pull/85, and how?

Lib/glyphsLib/parser.py
dougfelt: Yeah.  Left it because perhaps we want to test this again someday.

dougfelt:
Swatting sources was introducing additional changes to the sources beyond those intended because of inconsistencies in the glyphs source files.  In particular, the code was normalizing quoting assuming anything that didn't require quote marks for parsing (e.g. if it looked like a number or symbol) didn't need to be quoted when writing the file.

The inconsistencies are such that even trying to identify which places in the file get special treatment was failing, because in a single file the 'same' type was written different ways.  So I gave up on that.

This lets you invoke the parsing/writing code so as to inhibit escaping when you read or write the tree, and makes it possible to roundtrip read/write (without casting) leaving the file mostly as-is (indenting / spaces would still be normalized, but this does not seem to be inconsistent in the noto sources).

That in turn will let me avoid normalizing the files to have a clean starting point for further changes.  Instead, I can write code to swat the files (introduce the short names for those fonts that require them) and the result will have only the intended changes).

behdad:
lgtm

anthrotype:
@dougfelt ok thanks for the explanation.

anthrotype:
@dougfelt are you ok if we revert this PR? It conflicts with classtree branch, which has a different writer implementation that's not compatible with this.
Making (un)escaping optional in the parser/writer to ensure round-tripping only works in so far as you don't do the (un)casting. But the latter is no longer possible in the classtree implementation. The parser there instantiates custom objects directly, and the writer delegates to the objects to return their plist value.

dougfelt:
I've resigned myself to no longer being able to swat the files without introducing additional unintended changes.

anthrotype:
We'll try to ensure roundtrippability. @belluzj is adding tests for the new writer at #227

dougfelt:
Once the files are normalized, I assume they'll round trip-- they'd better.  Our current files have been generated by different versions of Glyphs over the past year and a half.  Originally I was going to normalize all of them first, then modify in a second pass, but instead we decided to ensure we didn't touch parts of the file that we weren't intentionally changing.  We'll just go back to the first approach.

anthrotype:
Yes. We shall try to match the latest version of Glyphs.app (some 2.5 beta), and keep it up to date as Glyphs.app changes over time.

PR #213, opened 2017-09-01
[Writer] correctly handle bytes/unicode; dumps return unicode; etc.

anthrotype:
- `Writer.write` now accepts either unicode strings or bytes string; the latter are decoded as UTF-8 internally.
- If the file object accepts bytes (i.e. is opened in binary mode), the unicode strings are encoded to UTF-8 via `codecs.StreamWriter`.
- If the file object already accepts unicodes, it is used directly without the codecs wrapper.

Other changes:
- `dumps` function now always return a unicode string, for consistency with python3 json module.
- in `Writer.__init__` the `out` argument (renamed `fp`) no longer defaults to `sys.stdout`, it's now a required positional argument (like in the `dump` function).
- make the `indent` argument behave like in json dump/dumps: ie. accept either an int or a string (e.g. `indent="\t"`).
- use same method that fonttools `XMLWriter` uses to make sure that the indentation whitespace is only written once per line (see `self._needindent` flag).
- pass `**kwargs` from `load` and `loads` functions down to `Parser` instance, like `dump`  and `dumps` do with `Writer`.

### DIFF
--- a/Lib/glyphsLib/parser.py
+++ b/Lib/glyphsLib/parser.py
@@ -176,7 +176,8 @@ class Writer(object):
     _sym_re = re.compile(
         r'^(?:-?\.[0-9]+|-?[0-9]+\.?[0-9]*|[_a-zA-Z0-9/\.][_a-zA-Z0-9\.]*)$')
 
-    def __init__(self, fp, indent=None, sort_keys=False, escape=True):
+    def __init__(self, fp, indent=None, sort_keys=False, escape=True,
+                 ensure_ascii=False):
         # figure out whether file object expects bytes or unicodes
         try:
             fp.write(b'')
@@ -191,6 +192,7 @@ class Writer(object):
 
         self.sort_keys = sort_keys
         self.escape = escape
+        self.ensure_ascii = ensure_ascii
         if not indent:
             self._indentstr = ''
         elif isinstance(indent, basestring):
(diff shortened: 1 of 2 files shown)


dougfelt:
If you're going to do this perhaps you could make another change.  I seems it would be better if the Writer didn't have the fp as part of its state, but instead the fp was passed in to write() and down through the other calls.  Then one could configure a Writer and use it to write multiple files, rather than reconfigure a new Writer for each file.

anthrotype:
@dougfelt Ah, now I see what you meant with your comment `# can resuse stdout, poor api design though`. I'll try to do that.

But there's still one thing which is not clear to me...
I noticed that if e.g. I input a non-ASCII string by setting a `localizedFamilyName = Greek;ελληνικά` custom parameter and save the file, Glyphs.app will keep it as a UTF-8 string, unescaped.

Whereas, the glyphsLib Writer always writes any non-ASCII characters using unicode escapes like `"\UFFFF"`.

We should probably change that to make the glyphsLib Writer behaves like Glyphs.app, i.e. it keeps UTF-8 strings as such without unescaping them...
Though it's not clear to me when Glyphs.app always does one or the other, or whether it used to do one thing and it changed recently by supporting UTF-8.

By the way, wasn't the NeXTSTEP property list supposed to be ASCII only?...
Hm.

Maybe we could have an optional `ensure_ascii=False` keyword argument like in the python2 `json.dump`, which when set to `True` it will do like we currenlty do, i.e. escape all non-ASCII characters with \UXXXX sequences; but by default it would strings as UTF-8?

***

Also, our unicode escape sequence only support BMP characters with 4 hex digits (`'\\U%04X'`). What about non-BMP characters, how does Glyphs.app escape them (if it does)? Maybe with 8 digits after the `\U`?

/cc @schriftgestalt

anthrotype:
@dougfelt PTAL to dcf2442. I still store `fp` temporarily in `self.file` instead of passing it down to the other calls; doing the latter would make it look too verbose, IMO.
--

dougfelt:
But writer is still not thread safe then.

anthrotype:
Where / how would you envision using Writer instances from multiple threads?

schriftgestalt:
I decided to use UTF-8 in .glyphs files some time ago. The escaping/unascaping was slow and error prone. And makes the files less readable. The plist format might be ASCII only, but all of Apples implementation are fine with UTF-8.

dougfelt:
I don't, however I don't see why we shouldn't support it.  If the object is not going to be usable from multiple threads we shouldn't make it look like it is, and should leave the fp as part of the state of the writer and make this explicit by keeping it in the constructor.

anthrotype:
@dougfelt ok, I prefer to revert the last commit for now and add a TODO for later. Eventually we could replace Writer with an Encoder that doesn't write directly to a file but yields chunks of data (like an iterator) that are written to file by `dump` or concatenated by `dumps` (JSONEncoder works like that).

@schriftgestalt Thanks for the info. So we should make the `\UXXXX`-style escapes optional in glyphsLib, and default to UTF-8.
What about the octal escape sequences used for control characters like line feeds (`\012`)? I see they are still present e.g. in features code.

anthrotype:
> The plist format might be ASCII only, but all of Apples implementation are fine with UTF-8

@schriftgestalt does Glyphs.app use any Apple API to serialize the ascii property list, or only to parse it?
Out of curiosity, I tried with pyobjc but I got an error saying that the "format kCFPropertyListOpenStepFormat not supported for writing".

schriftgestalt:
I used to use the build in serialisation for reading and writing. now I just use c file functions. The Apple API is more then 20 times slower.

schriftgestalt:
I don't remember 100%. But I think I used `NSDocument.writeToFile_atomically_(path, atomic)`. That produces the old style plist content.

schriftgestalt:
Before we do a lot of work on the writer, could we first decide on the new classes implementation? It uses a different writer.

moyogo:
I’d much rather prefer we finish the Classtree branch as well. Some expressed that they didn’t want glyphsLib to get slower but the Classtree branch is only 10% slower (in part because it write UFO3 layers instead of private robofont layers in the lib). Furthermore switching to fastUFO would make it twice as fast since a lot of the execution time is currently spent by defcon on broadcasting changes.

PR #217, opened 2017-09-07
use 'fileName' customParameter for instance filenames

moyogo:


anthrotype:
Is this a new built-in custom parameter?

moyogo:
It’s in the 2.3 manual https://glyphsapp.com/downloads/handbook/Glyphs-Handbook-2.3.pdf#page=184

schriftgestalt:
This parameter is at least four years old.

anthrotype:
I was just asking :)

moyogo:
OK. If that was the only question, I’ll merge then ;-).
3eefb4e broke the tests, changing the layer order.

anthrotype:
lgtm

PR #220, opened 2017-09-13
Allow interpolating instances with non rounded geometry

moyogo:
When interpolating instances one may want to not round values to integers.

This can be useful when adding masters to a designspace without piling up rounding differences.
For example, if  you have 4 masters and a bunch of instances in between:
```
A-B
| |
C-D
```
And you want to interpolate new masters where instances W, X, Y, Z and R were:
```
A--W--B
|  |  |
X--R--Z
|  |  |
C--Y--D
```
You do not want to round to integers in those new masters otherwise the other instances will be too different from the original 4-master instances. Say an instance interpolated between A and B will be different from an instance interpolated between A and rounded-W or rounded-W and B, whereas it will be closer to that instance interpolated between A and non-rounded-W or non-rounded-W and B.

### DIFF
--- a/Lib/glyphsLib/__init__.py
+++ b/Lib/glyphsLib/__init__.py
@@ -97,7 +97,8 @@ def build_masters(filename, master_dir, designspace_instance_dir=None,
         return ufos
 
 
-def build_instances(filename, master_dir, instance_dir, family_name=None):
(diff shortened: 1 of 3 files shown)


moyogo:
Updated description.

behdad:
Is there any reason to want the rounded option at all?  Rounding should only happen during the export IMO.

PR #222, opened 2017-09-21
Read the 'Replace Feature' customParameter for each instance

punchcutter:
In some cases instance-specific features are required which can be done inside of Glyphs by using the Replace Feature custom parameter on each instance. Currently this parameter is just ignored so none of the custom code gets inserted when running fontmake.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -41,11 +41,11 @@ Read and write Glyphs data as Python objects
 
 .. code:: python
 
-    with open('MyFont.glyphs', 'rb') as glyphs_file:
-        font = glyphsLib.load(glyphs_file)
-
-    with open('MyFont.glyphs', 'wb') as glyphs_file:
-        glyphsLib.dump(font, glyphs_file)
+    from glyphsLib import GSFont
+    
+    font = GSFont(glyphs_file)
+    
+    font.save(glyphs_file)
 
 The ``glyphsLib.classes`` module aims to provide an interface similar to
 Glyphs.app's `Python Scripting API <https://docu.glyphsapp.com>`__.


schriftgestalt:
can you check the Classtree branch and rework this to work with the new class structure?

punchcutter:
I don't see what needs reworking. ufo.features.text is the only thing getting changed here and that doesn't seem any different in master vs classtree.

Lib/glyphsLib/builder.py
```diff
@@ -463,6 +463,16 @@ def set_custom_params(ufo, parsed=None, data=None, misc_keys=(), non_info=()):
...
             value = int(value)
 
+        if name == 'Replace Feature':
+            if len(value.split(';')) > 2:
+                tag = value.split(';')[0]
+                code = ';'.join(value.split(';')[1:])
+                replace_dict = {'feature': tag, 'code': code}
+                new_code = 'feature %(feature)s {\n %(code)s \n} %(feature)s;' % replace_dict
+                pattern = re.compile('feature\s*%(feature)s[^;].*%(feature)s\s*;' % replace_dict, re.DOTALL)
+                replaced = re.sub(pattern, new_code, ufo.features.text)
```
anthrotype: minor: you could also directly use the `sub` method of the compiled pattern object
```diff
- replaced = re.sub(pattern, new_code, ufo.features.text)
+ replaced = pattern.sub(new_code, ufo.features.text)
```

Lib/glyphsLib/builder.py
```diff
@@ -463,6 +463,16 @@ def set_custom_params(ufo, parsed=None, data=None, misc_keys=(), non_info=()):
         if name in ('openTypeOS2WeightClass', 'openTypeOS2WidthClass'):
             value = int(value)
 
+        if name == 'Replace Feature':
+            if len(value.split(';')) > 2:
```
anthrotype: why requiring two semicolons instead of one? The Glyphs Handbook says 

>  The first four letters must be the feature name (such as ‘liga’), followed by a semicolon and the new feature code.

I guess you did that to ignore replacing with an empty or incomplete feature code (every statement must end with a semicolon in FEA).

Maybe we should log a warning saying something like "the custom parameter value is not properly formatted" when we hit the `else` block instead of silently ignoring the custom parameter?

Lib/glyphsLib/builder.py
```diff
@@ -463,6 +463,16 @@ def set_custom_params(ufo, parsed=None, data=None, misc_keys=(), non_info=()):
...
+                tag = value.split(';')[0]
+                code = ';'.join(value.split(';')[1:])
```
anthrotype: if you only want to split at the first occurrence of a semicolon, you could do
```python
code = value.split(';', 1)
```
instead of splitting and re-joining

Lib/glyphsLib/builder.py
punchcutter: Yes, it would definitely be good to do a little more validation here since the value could be any kind of junk.

schriftgestalt:
I would apply the `Replace Feature` parameter in an earlier state where you still have the GSFeature objects. Then you don't need to deal with regex and stuff.

punchcutter:
@schriftgestalt I would like apply it while we have the GSFeature objects, but right now those are only read once while creating the masters. Ideally we could run something like `add_features_to_ufo` on each instance in `interpolation.apply_instance_data`, but then that would also require passing the relevant data along. GSInstance could be given a features attribute so the instance_data passed along also contains the features. But to use `apply_instance_data` there we'd also need GSInstance to hold featurePrefixes and classes.

schriftgestalt:
The hole point of having the classes is to use them. The ideal solution would be to interpolate ourselves. The needed interpolation factors are stored in the instances but could also be computed by MutatorMath. Later on, that would allow to implement brace and bracket layers, too.

One reason to have the proxies is that the instances have a reference to the containing font object. So each instance can access the features already. (by `instance.parent.features`). So you only need to pass the instance along.

PR #227, opened 2017-10-04
Roundtrip .glyphs -> objects -> .glyphs

belluzj:
@moyogo 

This is a work in progress to have full round-trip between .glyphs files and UFO3 + designspace.

Plan:

1. Add tests for the writer (GS* classes -> .glyphs file)
    * Document (and fix?) small inconsistencies between the GS* classes and the Glyphs Python API documentation at https://docu.glyphsapp.com/
    * Add tests for the .glyphs -> GS classes -> .glyphs round-trip (e.g. https://gist.github.com/anthrotype/3f4c9b3e04c22ddcfda196b970de8549)
2. Merge classtree + the new tests in master

In another MR:
1. Test the round-trip using the writer:
```
 Test input = GSFont ------------------> designspace+UFO  ------------------> GSFont      
                |     builder/ufo.py   MutatorMath+defcon   builder/glyphs.py   |         
                |                                                (TODO)         |         
                |writer.py                                                      |writer.py
                |                                                               |         
                v                                                               v         
             .glyphs    <------------------------------------------------>   .glyphs      
                                 Test result = SHOULD BE THE SAME                                      
```

2. Implement all of `builder/glyphs.py` (UFO -> GSFont, the opposite of `builder/ufo.py`)

### DIFF
--- a/Lib/glyphsLib/__init__.py
+++ b/Lib/glyphsLib/__init__.py
@@ -58,13 +58,6 @@ def loads(s):
     return data
 
 
-def normalize(value):
-    """Read a .glyphs file from a bytes object.
-    Return a bytes object with the file's contents, normalized:
-    dictionary keys are put in order.
-    """
-    pass
-
 def load_to_ufos(file_or_path, include_instances=False, family_name=None,
                  debug=False):
     """Load an unpacked .glyphs object to UFO objects."""
--- a/Lib/glyphsLib/__init__.py
+++ b/Lib/glyphsLib/__init__.py
@@ -58,6 +58,13 @@ def loads(s):
     return data
 
 
+def normalize(value):
+    """Read a .glyphs file from a bytes object.
+    Return a bytes object with the file's contents, normalized:
+    dictionary keys are put in order.
+    """
+    pass
+
 def load_to_ufos(file_or_path, include_instances=False, family_name=None,
                  debug=False):
     """Load an unpacked .glyphs object to UFO objects."""
--- a/Lib/glyphsLib/builder/glyphs.py
+++ b/Lib/glyphsLib/builder/glyphs.py
@@ -13,7 +13,8 @@
 # limitations under the License.
 
 from glyphsLib import classes
-from glyphsLib import builder
+
+from .constants import GLYPHS_PREFIX
 
 # def designspace_from_ufos(ufos):
 #     pass
@@ -47,5 +48,5 @@ def to_glyphs_master(ufo):
     Extract from the given UFO the data that goes into a GSFontMaster.
     """
     master = classes.GSFontMaster()
-    master.id = ufo.lib[builder.GLYPHS_PREFIX + 'fontMasterID']
+    master.id = ufo.lib[GLYPHS_PREFIX + 'fontMasterID']
     return master
--- a/Lib/glyphsLib/classes.py
+++ b/Lib/glyphsLib/classes.py
@@ -939,7 +939,7 @@ class UserDataProxy(Proxy):
             del self._owner._userData[key]
 
     def __contains__(self, item):
-        return item in self._owner._userData.values()
+        return item in self._owner._userData
 
     def __iter__(self):
         for value in self._owner._userData.values():
@@ -2299,7 +2299,6 @@ class GSLayer(GSBase):
         "visible": bool,
         "width": float,
         "widthMetricsKey": unicode,
-        "smartComponentPoleMapping": dict,
     }
     _defaultsForName = {
         "weight": 600,
@@ -2329,9 +2328,6 @@ class GSLayer(GSBase):
         "vertWidth",
         "width",
     )
-    # FIXME: (jany) with the line below, user data would be shared between all
-    #   instances of GSLayer!?
-    # _userData = {}
 
     def __init__(self):
         super(GSLayer, self).__init__()
@@ -2341,7 +2337,6 @@ class GSLayer(GSBase):
         self.guides = []
         self._paths = []
         self._selection = []
-        self.smartComponentPoleMapping = {}
         self._userData = {}
 
     def __repr__(self):
@@ -2414,6 +2409,16 @@ class GSLayer(GSBase):
     userData = property(
         lambda self: UserDataProxy(self),
         lambda self, value: UserDataProxy(self).setter(value))
+
+    @property
+    def smartComponentPoleMapping(self):
+        if "PartSelection" not in self.userData:
+            self.userData["PartSelection"] = {}
+        return self.userData["PartSelection"]
+
+    @smartComponentPoleMapping.setter
+    def smartComponentPoleMapping(self, value):
+        self.userData["PartSelection"] = value
 
     @property
     def bounds(self):
--- a/Lib/glyphsLib/classes.py
+++ b/Lib/glyphsLib/classes.py
@@ -1010,7 +1010,7 @@ class GSCustomParameter(GSBase):
         'underlinePosition', 'underlineThickness', 'unitsPerEm',
         'vheaVertAscender',
         'vheaVertDescender', 'vheaVertLineGap', 'weightClass', 'widthClass',
-        'winAscent', 'winDescent', 'xHeight', 'year', 'Grid Spacing'))
+        'winAscent', 'winDescent', 'year', 'Grid Spacing'))
     _CUSTOM_FLOAT_PARAMS = frozenset((
         'postscriptBlueScale',))
 
--- a/Lib/glyphsLib/classes.py
+++ b/Lib/glyphsLib/classes.py
@@ -2064,7 +2064,7 @@ class GSInstance(GSBase):
     _classesForName = {
         "customParameters": GSCustomParameter,
         "exports": bool,
-        "instanceInterpolations": OrderedDict,
+        "instanceInterpolations": dict,
         "interpolationCustom": float,
         "interpolationCustom1": float,
         "interpolationCustom2": float,
@@ -2677,6 +2677,7 @@ class GSFont(GSBase):
         "gridSubDivision": 1,
         "unitsPerEm": 1000,
         "kerning": OrderedDict(),
+        "keyboardIncrement": 1,
     }
 
     def __init__(self, path=None):
@@ -2693,7 +2694,6 @@ class GSFont(GSBase):
         self._classes = []
         self.filepath = None
         self._userData = None
-        self.keyboardIncrement = None
 
         if path:
             assert isinstance(path, (str, unicode)), \
@@ -2712,7 +2712,7 @@ class GSFont(GSBase):
         return "<%s \"%s\">" % (self.__class__.__name__, self.familyName)
 
     def shouldWriteValueForKey(self, key):
-        if key in ("unitsPerEm", "versionMinor", "keyboardIncrement"):
+        if key in ("unitsPerEm", "versionMinor"):
             return True
         return super(GSFont, self).shouldWriteValueForKey(key)
 
--- a/Lib/glyphsLib/classes.py
+++ b/Lib/glyphsLib/classes.py
@@ -98,7 +98,7 @@ TTROUNDUP = 1
 TTROUNDDOWN = 2
 TRIPLE = 128
 
-#annotations:
+# Annotations:
 TEXT = 1
 ARROW = 2
 CIRCLE = 3
@@ -1354,7 +1354,10 @@ class GSNode(GSBase):
 
     def __init__(self, position=(0, 0), nodetype=LINE,
                  smooth=False, name=None):
-        self.position = point(position[0], position[1])
+        if isinstance(position, point):
+            self.position = position
+        else:
+            self.position = point(position[0], position[1])
         self.type = nodetype
         self.smooth = smooth
         self.name = name
@@ -1900,24 +1903,6 @@ class GSHint(GSBase):
         "options",
         "settings"
     )
-    # Hint types
-    TOPGHOST = -1
-    STEM = 0
-    BOTTOMGHOST = 1
-    TTANCHOR = 2
-    TTSTEM = 3
-    TTALIGN = 4
-    TTINTERPOLATE = 5
-    TTDIAGONAL = 6
-    TTDELTA = 8
-    CORNER = 16
-    CAP = 17
-    # Hint options
-    TTROUND = 0
-    TTROUNDUP = 1
-    TTROUNDDOWN = 2
-    TTDONROUND = 4
-    TRIPLE = 128
 
     def shouldWriteValueForKey(self, key):
         if key == "stem":
@@ -1958,6 +1943,45 @@ class GSHint(GSBase):
             return "<GSHint %s %s>" % (self.type, self.name)
         else:
             return "<GSHint %s %s>" % (self.type, direction)
+
+    @property
+    def originNode(self):
+        if self._originNode is not None:
+            return self._originNode
+        if self._origin is not None:
+            # Find the GSNode that is refered to by the indices in _origin
+            path_index, node_index = self._origin
+            layer = self.parent  # FIXME: (jany) I don't have access to the parent from the GSHint!!!
+            path = layer.paths[path_index]
+            node = path.nodes[node_index]
+            return node
+
+    @originNode.setter
+    def originNode(self, node):
+        self._originNode = node
+        self._origin = None
+
+    @property
+    def origin(self):
+        if self._origin is not None:
+            return self._origin
+        if self._originNode is not None:
+            # Find the path_index & node_index of the _originNode
+            path = self._originNode.parent
+            layer = path.parent
+            for path_index in range(len(layer.paths)):
+                if path == layer.paths[path_index]:
+                    for node_index in range(len(path.nodes)):
+                        if self._originNode == path.nodes[node_index]:
+                            return point(path_index, node_index)
+            return None
+
+    @origin.setter
+    def origin(self, origin):
+        self._origin = origin
+        self._originNode = None
+
+    # FIXME: (jany) if the above is OK after review, do the same for the others
 
 
 class GSFeature(GSBase):
@@ -2012,13 +2036,6 @@ class GSAnnotation(GSBase):
         "width": float,  # the width of the text field or size of the cicle
     }
     _parent = None
-
-    # Annotation types
-    TEXT = "Text"
-    ARROW = "Arrow"
-    CIRCLE = "Circle"
-    PLUS = "Plus"
-    MINUS = "Minus"
 
     @property
     def parent(self):
--- a/Lib/glyphsLib/classes.py
+++ b/Lib/glyphsLib/classes.py
@@ -828,6 +828,13 @@ class LayerPathsProxy(IndexedObjectsProxy):
         super(LayerPathsProxy, self).__init__(owner)
 
 
+class LayerHintsProxy(IndexedObjectsProxy):
+    _objects_name = "_hints"
+
+    def __init__(self, owner):
+        super(LayerHintsProxy, self).__init__(owner)
+
+
 class LayerComponentsProxy(IndexedObjectsProxy):
     _objects_name = "_components"
 
@@ -1962,12 +1969,16 @@ class GSHint(GSBase):
         else:
             return "<GSHint %s %s>" % (self.type, direction)
 
+    @property
+    def parent(self):
+        return self._parent
+
     def _find_node_by_indices(self, point):
         """"Find the GSNode that is refered to by the given indices."""
         path_index, node_index = point
-        layer = self.parent  # FIXME: (jany) I don't have access to the parent from the GSHint!!!
-        path = layer.paths[path_index]
-        node = path.nodes[node_index]
+        layer = self.parent
+        path = layer.paths[int(path_index)]
+        node = path.nodes[int(node_index)]
         return node
 
     def _find_indices_for_node(self, node):
@@ -2449,6 +2460,7 @@ class GSLayer(GSBase):
     def __init__(self):
         super(GSLayer, self).__init__()
         self._anchors = []
+        self._hints = []
         self._annotations = []
         self._components = []
         self.guides = []
@@ -2506,6 +2518,10 @@ class GSLayer(GSBase):
     anchors = property(
         lambda self: LayerAnchorsProxy(self),
         lambda self, value: LayerAnchorsProxy(self).setter(value))
+
+    hints = property(
+        lambda self: LayerHintsProxy(self),
+        lambda self, value: LayerHintsProxy(self).setter(value))
 
     paths = property(
         lambda self: LayerPathsProxy(self),
--- a/Lib/glyphsLib/classes.py
+++ b/Lib/glyphsLib/classes.py
@@ -1249,6 +1249,8 @@ class GSFontMaster(GSBase):
         self.italicAngle = 0.0
         self.customValue = 0.0
         self._userData = None
+        self.xHeight = None
+        self.capHeight = None
 
     def __repr__(self):
         return '<GSFontMaster "%s" width %s weight %s>' % \
@@ -1258,6 +1260,9 @@ class GSFontMaster(GSBase):
         if key in ("width", "weight"):
             if getattr(self, key) == "Regular":
                 return False
+            return True
+        if key in ("xHeight", "capHeight") and getattr(self, key) is not None:
+            # Write those values even when == 0
             return True
         return super(GSFontMaster, self).shouldWriteValueForKey(key)
 
--- a/Lib/glyphsLib/classes.py
+++ b/Lib/glyphsLib/classes.py
@@ -23,7 +23,7 @@ import glyphsLib
 from glyphsLib.types import (
     transform, point, rect, size, glyphs_datetime, color, floatToString,
     readIntlist, writeIntlist, needsQuotes, feature_syntax_encode, baseType,
-    encode_dict_as_string_for_gsnode
+    encode_dict_as_string_for_gsnode, decode_dict_as_string_from_gsnode
 )
 from glyphsLib.parser import Parser
 from glyphsLib.writer import GlyphsWriter
@@ -1405,8 +1405,9 @@ class GSNode(GSBase):
 
         # TODO: Use proper string parsing used in other classes
         if '{\\nname' in line:
-            nameMatch = re.match(r'\{\\nname = "?(.+?)"?;\\n\}', m[4]).groups()
-            self.name = nameMatch[0]
+            parser = Parser()
+            value = decode_dict_as_string_from_gsnode(m[4])
+            self.name = parser.parse(value)["name"]
         else:
             self.name = None
 
--- a/Lib/glyphsLib/classes.py
+++ b/Lib/glyphsLib/classes.py
@@ -2059,7 +2059,7 @@ class GSInstance(GSBase):
     _classesForName = {
         "customParameters": GSCustomParameter,
         "exports": bool,
-        "instanceInterpolations": dict,
+        "instanceInterpolations": OrderedDict,
         "interpolationCustom": float,
         "interpolationCustom1": float,
         "interpolationCustom2": float,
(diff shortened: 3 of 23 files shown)


tests/writer_test.py
```diff
@@ -0,0 +1,471 @@
...
+class WriterTest(unittest.TestCase, test_helpers.AssertLinesEqual):
+
+    def assertWrites(self, glyphs_object, text):
+        """Assert that the given object, when given to the writer,
+        produces the given text.
+        """
+        expected = text.splitlines()
+        actual = test_helpers.write_to_lines(glyphs_object)
+        # print(expected)
+        # print(actual)
```
moyogo: Commented print shouldn’t be there.

tests/writer_test.py
```diff
@@ -0,0 +1,471 @@
...
+        ])
+        # userData
+        font.userData = {
+            'a': 'test',
+            'b': [1, {'c': 2}]
+        }
+        # grid
+        font.grid = 35
+        # gridSubDivisions
+        # FIXME: (jany) In this library it is called "gridSubDivision" (no s)
```
moyogo: It’s "gridSubDivisions" in [GlyphsSDK](https://github.com/schriftgestalt/GlyphsSDK/blob/master/ObjectWrapper/GlyphsApp/__init__.py#L2530). We can fix the typo in classes.py.

Lib/glyphsLib/classes.py
```diff
@@ -2732,3 +2723,13 @@ def note(self):
     @note.setter
     def note(self, value):
         self.customParameters["note"] = value
+
+    # FIXME: (jany) I don't understand why this is needed for
+    #   gridSubDivisions but it works without it for upm
+    @property
+    def gridSubDivisions(self):
+        return self.gridSubDivision
```
anthrotype: A property getter cannot recursively return itself, nor should the setter overwrite itself..
Either you get or set something else (e.g. `self._gridSubDivision`) or simply remove the property if it's a simple read/write attribute.

anthrotype: woah.. I totally missed that "s" at the end. Please ignore my comment, sorry :)

belluzj: No problem :) Do you know why there is a difference between `upm` vs `unitsPerEm` (which don't have the getter/setter) and this `gridSubDivision(s)` which need a getter/setter for the tests to pass?

anthrotype:
I am not sure `roundtrip.py` is the right name for the new module. It conveys the idea of going from one format (glyphs) to another (UFO) and back to the original one, while in fact it's meant to go from UFO to glyphs. Then we have builder.py which currently only builds UFOs from glyphs. 

Maybe, we could make `builder` into a package with an `__init__.py` and have two sub-modules in there, one called `builder/ufo.py` and the other `builder/glyphs.py`. The `builder/__init__.py` could export all the API-level functions, e.g. `to_ufos`, `to_glyphs`, etc.

anthrotype:
Another question. If we wish to roundtrip Glyphs -> Designspace + UFO masters -> Glyphs, how do we handle the fact that currently we have no ways to store the Glyphs-specific *instance* custom parameters in the designspace? The way we do right now is we return this data back to fontmake, which then calls `apply_instance_data` on the UFO instances after mutatorMath has completed.
I have proposed to add a `lib` element to the designspace document specification, though it didn't trigger any comments:

https://github.com/LettError/designSpaceDocument/issues/19

Lib/glyphsLib/classes.py
anthrotype: i think it's because in the GSBase constructor, we are not translating the key from `_classesForName` dict using the `_wrapperKeysTranslate` dict, so `gridSubDivision` (without the "s") is being set.

anthrotype: I suggest you add a line `key = self._wrapperKeysTranslate.get(key, key)` before calling `setattr(self, key, value)` in the `GSBase.__init__`, like the `__setitem__` method does..

Although I really don't like that one says to a dict, "please set this key-value pair", and the dict replies, "Ok, i'll set the value but using a slightly different key".
Also, I've never seen a write-only dictionary, with a `__setitem__` but no corresponding `__getitem__`...

anthrotype: Btw, the `__contains__` method also does not do the _wrapperKeysTranslate trick, but simply delegates to `hasattr`.
So one may set `font["unitsPerEm"] = 1000`, and it will set the `upm` attribute to 1000, but then if you test `"unitsPerEm" in font` it will return `False`. This is very weird.

anthrotype:
I think that, before focussing on roundtripping glyphs -> UFOs -> glyphs, we should make sure that the parser and writer can roundtrip themeselves, glyphs -> GS objects -> glyphs.
Is that the case for the classtree branch?
Check the discussion ensuing from this https://github.com/googlei18n/glyphsLib/pull/203#issuecomment-327013988
We need to address this problem before we can merge classtree to master. There are merge conflicts with the commits that were pushed to master since v1.8.0 tag.
I am thinking of reverting master to how it was at v1.8.0, merge classtree branch and then reapply manually those changes, but only when they make sense in the new parser/writer implementations from classtree.

Lib/glyphsLib/classes.py
moyogo: The GlyphsSDK seems to have `font.unitsPerEm()` which returns the objc font value for unitsPerEm, while `font.upm` is the Python API wrapper around that which can be set.

It’s similar for`gridSubDivision`, it’s the obj font function whereas `gridSubDivisions` is the Python API wrapper around `gridSubDivision()` and `setGridSubDivision(value)`.

moyogo: Maybe we should stick to only having `font.upm` and `font.gridSubDivisions` and not have `font.unitsPerEm` nor `font.gridSubDivision` in the Python API.

/cc @schriftgestalt

belluzj:
I updated the description of what I intend to do, to take your comments into account.

Lib/glyphsLib/classes.py
anthrotype: yes, it doesn't make sense to have both, at least here in glyphsLib.

schriftgestalt:
I did some work on the .glyphs -> GS objects -> .glyphs and fixed most of the issues I found (there are only some float rounding differences left).

belluzj:
OK cool! So I can just keep adding tests to prevent future regressions on the work that has been done: unit tests for the writer on all classes plus one or two integration tests with a .glyphs file, like what Cosimo did in his gist?

anthrotype:
sounds good to me. Thanks Jany for working on this!

anthrotype:
I think the bot is drunk, because early on it said the CLA was ok...

gmlewis:
Investigating drunk googlebot...

anthrotype:
I think we should rename `writer.GlyphsWriter` class to simply `writer.Writer`, for consistency with `parser.Parser`, or alternatively rename `Parser` to `GlyphsParser`.

gmlewis:
After a bit of investigation, it appears that @belluzj has been removed from the Corporate CLA that @belluzj was originally found to be included in.

As a result, this PR can unfortunately not be merged until @belluzj resolves this issue with the administrators of @belluzj's corporate CLA.  Please feel free to contact me directly at gmlewis@google.com in order to provide you with your corporate CLA primary contact information.

anthrotype:
@gmlewis thanks for update. I'm sure it's just a temporary glitch.

gmlewis:
Yes indeed - issue resolved.  Thanks!

tests/roundtrip_test.py
```diff
@@ -19,8 +19,7 @@
 import sys
 
 from glyphsLib import classes
-from glyphsLib.builder import to_ufos
-from glyphsLib.roundtrip import to_glyphs
+from glyphsLib.builder import (to_ufos, to_glyphs)
```
anthrotype: you don't need parenthesis unless you want to wrap a multi-line import

gmlewis:
WOW!!!  Again?  Investigating...

anthrotype:
@belluzj it looks like there's a circular import between `glyphsLib.builder.glyphs` and `glyphsLib.builder`. The glyphs module imports builder which in turn imports the glyphs module. In this cases the order of the imports matters. Perhaps in python3 the import mechanism handles this in a different way.

Lib/glyphsLib/builder/glyphs.py
```diff
@@ -12,8 +12,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from glyphsLib import classes
 from glyphsLib import builder
```
anthrotype: get rid of the circular import here, instead import the `GLYPHS_PREFIX` that you need below directly from the `ufo` module where it's defined.

Actually these symbols that are shared by both `ufo` and `glyphs` could be defined in the builder/__init__.py, or in a `constants` or similar sub-module of builder package.

Lib/glyphsLib/__init__.py
```diff
@@ -54,6 +54,13 @@ def loads(value):
     return data
 
 
+def normalize(value):
+    """Read a .glyphs file from a bytes object.
```
anthrotype: why bytes? It's better to work with strings (`unicode`, or `str` on py3) since this is really text, not binary data.
The parser should be able to handle both bytes or unicode input: the former has to be decoded from utf-8, the latter used as is.
Similarly the writer should work with unicode strings internally and then encode them to utf-8 when writing out to a "binary" file/stream, or write unicode directly to a file opened for "text" (which handles the encoding for us), or to an `io.StringIO` which is a stream of unicode strings.  

https://github.com/googlei18n/glyphsLib/pull/213

belluzj:
I have now added writer tests for all the classes. However, some of those tests have FIXMEs or are wrong (those about the Smart Components) so I will fix all of that.

Next, I will work on the .glyphs -> classes -> .glyphs roundtrip. We discussed it with @moyogo and since there seems to be a problem with the order of the keys in the dictionaries (it changes with the versions of Glyphs, so it would be hard to try to match it) we considered the idea of a normalization step, as mentioned by @dougfelt in https://github.com/googlei18n/glyphsLib/pull/212#issuecomment-334529175

This step must be very simple in terms of code and do the minimum amount of work, so that unit testing is easy and that we can be confident that no data will be lost. The step could use the Parser to instantiate simple Python dictionaries (not using any code from classes.py, and not doing any smart casting/coercion/decoding...) and the Writer to write them. The only "smart" things that will be done are in the Writer:
* ordering the keys of the dictionaries, with a simple rule, like "alphabetically". This would allow to remove the "_keyOrder" tables.
* normalizing values = for example with strings: quote everything and encode everything with UTF-8 (or any valid strategy that is simple and systematic)

Also, the normalized output must be readable by Glyphs.

With this step and the more simple writer, we could have a round-trip like this:

```
     .glyphs
(original format
= random key order +----------->   .glyphs    +-------> classes +------->   .glyphs
version-dependent)   normalize   (normalized)  "smart"            write   (normalized)
                         =            ^         parse                          ^
                   "dumb" parse       |                                        |
                     + write          |                                        |
                                      |                                        |
                                      +----------------------------------------+
                                               They should be the same

```

Does that sound OK?

schriftgestalt:
I did change some of the ordering abut that was while working on the round trippping to make it easier to replicate or more clear. It is not changed randomly and should be quite stable now.

anthrotype:
> I did change some of the ordering abut that was while working on the round trippping to make it easier to replicate or more clear. It is not changed randomly and should be quite stable now.

If glyphsLib writer were able to match the current Glyphs.app's formatting, and if Georg guarantees this is relatively stable, it'd be nice if the "normalized" output was identical to the output from the current Glyphs.app version. But I'm ok with anything that moves this forward.

belluzj:
OK. I'm adding a test that does the round-trip on GlyphsUnitTestSans.glyphs without normalization, and so far it looks like there is "just" a problem with the order of the layers, but most of the other formatting looks identical, so maybe normalization will not be needed :)

belluzj:
The roundtrip `.glyphs` -> classes -> `.glyphs` works without normalization on GlyphsUnitTestSans (it also has smart components now).

I propose that now I should collect a few open-source `.glyphs` files and add them as round-trip tests? I guess that I need to cover:
1. various versions of Glyphs
2. all the possible Glyphs features

However I don't think it's a good idea to add files that are too big, because that would slow down the unit tests and make the git repository too big. What do you think?

anthrotype:
I would try to run a round-trip script on the *.glyphs files in noto-source repository, since they are many and were produced over different time periods.
That could give you an idea.
But yeah, we can't add too big or too many test files.
I wouldn't care too much about old versions of Glyphs. As long we roundtrip ourselves and we roundtrip the latest and greatest, we should be ok.

moyogo:
Roundtriping the noto-source fonts could be good if it’s not too slow.
But there are probably structures that aren’t used so it would be better to have a solid GlyphsUnitTestSans.glyphs.

belluzj:
OK, so what I do is: roundtrip Noto, and add all the features that don't work as unit tests and in GlyphsUnitTestSans

Lib/glyphsLib/classes.py
```diff
@@ -1959,6 +1944,45 @@ def __repr__(self):
...
+                        if self._originNode == path.nodes[node_index]:
+                            return point(path_index, node_index)
+            return None
+
+    @origin.setter
+    def origin(self, origin):
+        self._origin = origin
+        self._originNode = None
+
+    # FIXME: (jany) if the above is OK after review, do the same for the others
```
belluzj: @moyogo Please review this

schriftgestalt:
If there are questions from the Noto round tripping tests, I’m happy to help.

belluzj:
Thanks! Here are the issues that I have identified so far, starting with those that I know how to fix:

* The order of instanceInterpolations is changed (I will put and OrderedDict for that)
* The keyboardIncrement key is removed (I will fix the test for that)

However I'm not sure what to do about these ones:

* Some fonts have quotes around some numbers (panose options, xHeight, weightClass). Those quotes always get removed by glyphsLib
* Some fonts have 6 digits of precision for float values in instanceInterpolations, but the output precision of glyphsLib seems to be 5
* Some fonts have control points whose coordinates are slightly off, like "505.999". glyphsLib outputs an integer coordinates, e.g. "506"
* One font has a DisplayString with the value `/uni10447`, not quoted, and glyphsLib outputs it with quotes
* I think I saw a "width" key moving around somewhere, but I'm not sure

I'm running the tests again and saving the output, I will post it as a gist and we will be able to refer to specific lines of the diff. I have included the "appVersion" in the name of tests methods so that we can discard tests that fail on "bad" versions (too old or unstable?).

belluzj:
https://gist.github.com/belluzj/1e9715052bd35fdf021ebd698c2b7bd0

Lib/glyphsLib/classes.py
moyogo: We can drop `hint.origin` as the GlyphsSDK has `hint.originNode` instead. The same goes for  and `hint.targetNode`, `hint.otherNode1` and `hint.otherNode2`. I think `hint.origin` is an objc thing, not part of the Python API.

We can also pythonize the names of `Hint__origin__pos()` to `_origin_pos()` and `Hint__width__pos()` to `_width_pos()`.

belluzj:
@georg: with @moyogo, we're testing the round-trip on Noto fonts after opening and saving all the files using a more recent version of Glyphs.app. We tried with 2.4.3 1064, but that did not solve all the problems. For example: [this file](https://raw.githubusercontent.com/moyogo/noto-source/normalized-1064/src/NotoSans-ItalicMM.glyphs) has both quoted and unquoted panose values, so the roundtrip fails because it unquotes everything.

However, when saved using Glyphs 2.5b 1071, [the same file](https://raw.githubusercontent.com/moyogo/noto-source/normalized-1071/src/NotoSans-ItalicMM.glyphs) has only unquoted panose values, which is good. **Does that mean that we should aim for the round-trip to work with Glyphs 2.5 (and up) only?**

@moyogo made a Noto branch where all files have been loaded and saved with version 1071, so I'm going to test on this and post the output soon.

anthrotype:
yeah, let's content ourselves of matching the latest version.

it's @schriftgestalt btw, not [@]georg ;)

schriftgestalt:
>Some fonts have quotes around some numbers (panose options, xHeight, weightClass). Those quotes always get removed by glyphsLib
> Some fonts have 6 digits of precision for float values in instanceInterpolations, but the output precision of glyphsLib seems to be 5
> I think I saw a "width" key moving around somewhere, but I'm not sure

That was adjusted in Glyphs 2.5. So if you run the files through he latest version, you should get the same changes. 

> Some fonts have control points whose coordinates are slightly off, like "505.999". glyphsLib outputs an integer coordinates, e.g. "506"
> One font has a DisplayString with the value /uni10447, not quoted, and glyphsLib outputs it with quotes

That should be fixed in Glyphs, too. But I would need to test that. What file was it?

> Does that mean that we should aim for the round-trip to work with Glyphs 2.5 (and up) only?

Yes. As I said, I adjusted the behaviour of Glyphs 2.5 and glyphsLib together to make it more compatible.

belluzj:
Thanks for the clarifications :)

> Some fonts have control points whose coordinates are slightly off, like "505.999". glyphsLib outputs an integer coordinates, e.g. "506"

I think I fixed this one in the previous commit.

> One font has a DisplayString with the value /uni10447, not quoted, and glyphsLib outputs it with quotes

I have not looked into that one yet. It's in NotoSansDeseret-MM.glyphs: https://gist.github.com/belluzj/1e9715052bd35fdf021ebd698c2b7bd0#file-pytest_noto-txt-L12186-L12213

belluzj:
These are the diffs for Noto files with Glyphs version 1071: (the size of the diffs went from 1,3 Mb to 300 Kb!)

https://gist.github.com/belluzj/6b19d9d72d8b9a0a9b8a95012a945f04

The remaining problems are:
* Bad rounding of coordinates (should be fixed) [example](https://gist.github.com/belluzj/6b19d9d72d8b9a0a9b8a95012a945f04#file-noto_1071-txt-L234)
* Quotes around xHeight values; [example](https://gist.github.com/belluzj/6b19d9d72d8b9a0a9b8a95012a945f04#file-noto_1071-txt-L689)
* Empty collections are removed from the output; [example](https://gist.github.com/belluzj/6b19d9d72d8b9a0a9b8a95012a945f04#file-noto_1071-txt-L9755)
* Empty category and subCategory are removed; [example](https://gist.github.com/belluzj/6b19d9d72d8b9a0a9b8a95012a945f04#file-noto_1071-txt-L14262)

@schriftgestalt Could you tell which of those should be fixed here vs. fixed in Glyphs?

schriftgestalt:
Resaving the NotoSansDeseret-MM.glyphs file doesn’t introduce any changes in Glyphs 2.5 (except the new version number).

schriftgestalt:
> - Quotes around xHeight values; example

That should keep the quotes as it is read as string and strings that only contain numbers should be quoted to retain the type.

schriftgestalt:
And the glyph names in the NotoSansChakma are good. The one pointed at in the example is `uni1112Duni11100.short` and it should be `u1112D11100.short`.

belluzj:
> > Quotes around xHeight values; example
>
> That should keep the quotes as it is read as string and strings that only contain numbers should be quoted to retain the type.

There is some code in `GSCustomParameter` that explicitly casts the `xHeight` to int (`xHeight` is in `_CUSTOM_INT_PARAMS`). Should I change this, so that the type of the `xHeight` is kept from the file? Should I also change it for other custom parameters?

schriftgestalt:
Custom parameters that are explicitly typed should always be cast. But ‘xHeight’ shouldn’t be in that list.

belluzj:
110 out of 112 green tests for Noto files!

https://gist.github.com/belluzj/cf176ac062ad9bec4b7fd5751646c04d

anthrotype:
@belluzj now that classtree is merged into master, you should probably merge master into your branch in order to sync, and then you can re-open the pull request, this time directed towards master instead of classtree.

anthrotype:
oh, I didn't know one could modify the base branch for an open pull request directly from Github!
https://github.com/blog/2224-change-the-base-branch-of-a-pull-request
:+1:

Lib/glyphsLib/parser.py
```diff
@@ -61,6 +61,10 @@ def parse_into_object(self, res, text):
         return i
 
     def _guess_dict_type(self, parsed, value):
+        if value in ('infinity', 'inf', 'nan'):
+            # Those values would be accepted by `float()`
+            # But `infinity` is a glyph name
```
anthrotype: good catch!

Lib/glyphsLib/classes.py
```diff
@@ -1261,6 +1250,9 @@ def __init__(self):
         self._custom2 = None
         self.italicAngle = 0.0
         self.customValue = 0.0
+        self._userData = None
+        self.xHeight = None
+        self.capHeight = None
```
belluzj: @schriftgestalt What are the defaults for `xHeight` and `capHeight`? I guess it's not `0` because some files specify a value of `0`

anthrotype: @belluzj what are the defaults when you create an empty new font? (I don't have my mac at hand)

Lib/glyphsLib/classes.py
```diff
@@ -1931,6 +1960,45 @@ def __repr__(self):
...
+    def originNode(self):
+        if self._originNode is not None:
+            return self._originNode
+        if self._origin is not None:
+            # Find the GSNode that is refered to by the indices in _origin
+            path_index, node_index = self._origin
+            layer = self.parent  # FIXME: (jany) I don't have access to the parent from the GSHint!!!
+            path = layer.paths[path_index]
+            node = path.nodes[node_index]
+            return node
```
belluzj: @schriftgestalt I have a question about this piece of code: when I read `origin` from the file, I have a path id & node id. In order to return the matching `GSNode`, what I want to do is: from the GSHint where I am, I want to go up one level to find the layer, and access the path and node by id. Except I don't have `GSHint::parent`, according to the Python API documentation, so I can't go up. Should I add the parent to this class? Or is there another way? Or maybe it's not the right place to do that? Thanks in advance

Lib/glyphsLib/types.py
```diff
@@ -362,6 +363,23 @@ def needsQuotes(string):
     return needsQuotes
 
 
+# FIXME: (jany) why is `feature_syntax_encode` different?
+def encode_dict_as_string_for_gsnode(value):
+    """Takes the PLIST string of a dict, and returns the same string
+    encoded such that it can be included in the string representation
+    of a GSNode."""
+    value = value.replace('"', '\\"')
+    return value.replace("\n", "\\n")
```
belluzj: @schriftgestalt There is an inconsistency here in string escaping between "\n" vs "\012", as noted in https://github.com/googlei18n/glyphsLib/issues/119. Do you still plan to fix it in Glyphs 2.5? If so, I can remove these two functions.

tests/writer_test.py
```diff
@@ -0,0 +1,1048 @@
...
+        layer.leftMetricsKey = "A"
+        # rightMetricsKey
+        layer.rightMetricsKey = "A"
+        # widthMetricsKey
+        layer.widthMetricsKey = "A"
+        # bounds: read-only, computed
+        # selectionBounds: read-only, computed
+        # background
+        # FIXME: (jany) why not use a GSLayer like the official doc suggests?
+        background_layer = classes.GSBackgroundLayer()
```
belluzj: @schriftgestalt Why is there a `GSBackgroundLayer` class?

tests/writer_test.py
```diff
@@ -0,0 +1,1048 @@
...
+            other1 = "{0, 2}";
+            other2 = "{1, 0}";
+            type = 16;
+            name = "My favourite hint";
+            options = 128;
+            }
+        """))
+
+        # FIXME: (jany) What about the undocumented scale & stem?
+        #   -> Add a test for that
```
belluzj: @schriftgestalt The [file format documentation](https://github.com/schriftgestalt/GlyphsSDK/blob/master/GlyphsFileFormat.md) mentions `stem` and `scale`, but I don't see them in the [Python API for GSHint](https://docu.glyphsapp.com/#gshint). How can I set those values using an official API?

belluzj:
Noto roundtrip results: https://gist.github.com/belluzj/593bc9bebf1ef2f256891b1f0cfc23d0

@schriftgestalt The only problem left for roundtrip is the float precision problem that we discussed on the commit that I removed, to be fixed in Glyphs.

schriftgestalt:
> @schriftgestalt What are the defaults for xHeight and capHeight? I guess it's not 0 because some files specify a value of 0

xHeight = 500
capHeight = 700
ascender = 800

but the values are always written.

> Except I don't have GSHint::parent, according to the Python API documentation, so I can't go up.

Thats why we have all the proxies. They are responsible to set the parent. So we need a parent and a proxy to the `layer.hints` property for the hints, too.

> @schriftgestalt There is an inconsistency here in string escaping between "\n" vs "\012", as noted in #119. Do you still plan to fix it in Glyphs 2.5? If so, I can remove these two functions.

I just had a look at it. I would like to change to only use \n but that breaks my parser so one could not open the files in a previous version any more. So for now, in text properties (like `feature.code`) it will have `\012`.

belluzj:
@schriftgestalt We updated the test file "GlyphsUnitTestSans.glyphs" by opening and saving it with Glyphs 2.5b 1072. The result is here: https://github.com/moyogo/glyphslib/commit/74b5280ca01db4a0e4ca86f205509d5feccb8dce

I think there are two problems in the output of v1072:
1. The type of the node's userData, `rememberToMakeCoffee` has been changed from boolean/number (`= 1`) to string (`= "1"`). Is that normal?
2. The `\n` have been removed completely from the node's user data. I think this is a problem because then I cannot use the writer to write that userData, since it puts new lines everywhere (or I can use the writer and then remove all newlines from the output? but there is a new function that removes newlines) A better solution IMO would be to encode this user data using the same function as everywhere else, ie. with \012, so I could use that function which already exists, and everything would be consistent. What do you think?

schriftgestalt:
> The type of the node's userData, rememberToMakeCoffee has been changed from boolean/number (= 1) to string (= "1"). Is that normal?

No. I’ll fix it.

> The \n have been removed completely (edit: partially) from the node's user data. I think this is a problem because then I cannot use the writer to write that userData, since it puts new lines everywhere (or I can use the writer and then remove all newlines from the output? but there is a new function that removes newlines) A better solution IMO would be to encode this user data using the same function as everywhere else, ie. with \012, so I could use that function which already exists, and everything would be consistent. What do you think?

I do a replacement like this:
```python
userDataString = userDataString.replace("    ", "") # remove indentation
userDataString = userDataString.replace("{\n", "{")
userDataString = userDataString.replace("\n}", "}")
userDataString = userDataString.replace("\n", "\\n")
```

belluzj:
> I do a replacement like this:
>    ```
>    userDataString = userDataString.replace("    ", "") # remove indentation
>    userDataString = userDataString.replace("{\n", "{")
>    userDataString = userDataString.replace("\n}", "}")
>    userDataString = userDataString.replace("\n", "\\n")
> ```
OK but why? If you're going to change the way this is output, you might as well make it consistent with other strings? Is there something that I don't see?

This issue points out the inconsistencies in a node's userData: https://github.com/googlei18n/glyphsLib/issues/119

schriftgestalt:
I’ll have a look at the other places where line breaks are need to be escaped. the current code does not like the `\n` in the strings so I need to fix the reading first and then, a few version later can change the writing, too.

belluzj:
I think this PR is mostly done, please tell me if there are other things that I can improve, tests to add...

Some questions to @schriftgestalt have not been answered yet, but I guess that can wait until the next release of Glyphs and that I could do another small PR then.

I have started working on the round-trip to UFO in another branch, but it's very WIP. Currently I'm splitting the `builder/ufo.py` into smaller files, ideally one per "feature" (like kerning, guidelines, paths, filters...) in order to add the code for the "other direction" (UFO -> Glyphs) next to its existing counterpart in each of those files.

anthrotype:
> I'm splitting the builder/ufo.py into smaller files, ideally one per "feature"

sounds good, thanks

schriftgestalt:
One thing that could be done is to move some of the functionality from the builder into the classes.

belluzj:
Could you expand on that? I probably won't do it in this PR but in the next one about UFO round-trip.

anthrotype:
I actually like the way it is now with the glyphs specific stuff in classes.py, kept separated from the UFO specific stuff and the builder module managing their interaction.

schriftgestalt:
There are things that are independent from the ufo builder. Generating instances for example.

PR #237, opened 2017-10-18
Adding control over anchor propagation

jvgaultney:
Adding an argument to to_ufos() that gives control over whether propagate_font_anchors() is run. Also enables that argument to be used by load_to_ufos(), build_masters(), build_instances(). The default remains True - that propagation is done unless specifically turned off.

### DIFF
--- a/Lib/glyphsLib/__init__.py
+++ b/Lib/glyphsLib/__init__.py
@@ -59,7 +59,7 @@ def loads(s):
 
 
 def load_to_ufos(file_or_path, include_instances=False, family_name=None,
-                 debug=False):
+                 propagate_anchors=True, debug=False):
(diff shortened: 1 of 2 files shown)


jvgaultney:
I signed it!

anthrotype:
thanks Victor! I merged it manually to resolve the merge conflicts. Next time, please make sure you git pull the latest master before branching out, or rebase your branch on the updated master.

moyogo:
I'm wondering if propagating anchors should be an operation happening in ufo2ft, optionally before mark feature generation instead. One may want to propagate anchors without having a .glyphs file

anthrotype:
@moyogo I agree. We can define a private key to store in the lib so that ufo2ft markFeatureWriter can read and apply.

PR #249, opened 2017-10-20
Various fixes

belluzj:


belluzj:
Makes sense. Should I leave the main in the parser?

anthrotype:
If it works, why not. ;)

moyogo:
LGTM!

moyogo:
@belluzj  Actually could you add the fix "if int(font.appVersion) < 895:" here as well, so we can merge it before ufo_roundtrip is done?

PR #254, opened 2017-10-23
Don't crash when updating an existing custom parameter

belluzj:


anthrotype:
ok, this one is tricky, because one may well have multiple custom parameters with the same name, especially true for Filters.
That's why I don't like to confuse the interface of sequences and that of mappings...
What does the Glyphs.app python API do in these cases?

schriftgestalt:
In Glyphs, custom parameters are normal lists. But there is a convenient method to get the (first) parameter by name. Those two ways are wrapped by allowing access by index and key. The key access simply iterated over the list and returned the first parameter it finds. Some keys should be unique, those are save to access by key. Others can have multiple entries, those need to accessed by iterating yourself.

anthrotype:
Thanks Georg. I'm ok if Glyphs.app does the same, even though I keep finding this confusing. Maybe it's just me.

moyogo:
In Glyphs.app:
```python
font = GSFont()
font.customParameters["Filter"] = "foo"
print(font.customParameters)
font.customParameters["Filter"] = "bar"
print(font.customParameters)
font.customParameters.append(GSCustomParameter("Filter", "baz"))
print(font.customParameters)
font.customParameters["Filter"] = "foo"
print(font.customParameters)
```
prints
```
(<GSCustomParameter Filter: foo>)
(<GSCustomParameter Filter: bar>)
(<GSCustomParameter Filter: bar>, <GSCustomParameter Filter: baz>)
(<GSCustomParameter Filter: foo>, <GSCustomParameter Filter: baz>)
```

anthrotype:
thanks Denis for checking. So I guess we can merge this

PR #264, opened 2017-10-30
Add missing import for glyphdata

### LINKED ISSUE #263: 'module' object has no attribute 'glyphdata' using to_ufos()
I'm getting the following error:
```
Python 2.7.12 (default, Nov 19 2016, 06:48:10) 
[GCC 5.4.0 20160609] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>>
>>> import glyphsLib
>>> gfont = open("Andika-RB.glyphs", 'r')
>>> font = glyphsLib.load(gfont)
>>> ufos = ufos = glyphsLib.to_ufos(font)
  File "build/bdist.linux-x86_64/egg/glyphsLib/builder/__init__.py", line 43, in to_ufos
  File "build/bdist.linux-x86_64/egg/glyphsLib/builder/builders.py", line 133, in masters
  File "build/bdist.linux-x86_64/egg/glyphsLib/builder/glyph.py", line 43, in to_ufo_glyph
AttributeError: 'module' object has no attribute 'glyphdata'
>>> 
```
This is with the latest download of glyphsLib.

belluzj:
Fixes #263

### DIFF
--- a/Lib/glyphsLib/__init__.py
+++ b/Lib/glyphsLib/__init__.py
@@ -26,8 +26,8 @@ from glyphsLib.interpolation import interpolate, build_designspace
 from glyphsLib.parser import load, loads
 from glyphsLib.writer import dump, dumps
 from glyphsLib.util import write_ufo
-import glyphsLib.classes
 
+from glyphsLib.classes import __all__ as __all_classes__
 from glyphsLib.classes import *
 
 __version__ = "2.0.0.dev0"
@@ -36,11 +36,10 @@ __version__ = "2.0.0.dev0"
 # "TypeError: Item in ``from list'' must be str, not unicode" on Python 2.
 # Thus we need to encode the unicode literals as ascii bytes.
 # https://bugs.python.org/issue21720
-__all__ = [tostr(s) for s in (
+__all__ = [tostr(s) for s in [
     "build_masters", "build_instances", "load_to_ufos",
     "load", "loads", "dump", "dumps",
-)]
-__all__.extend(glyphsLib.classes.__all__)
+ ] + __all_classes__]
 
 logger = logging.getLogger(__name__)
 
--- a/Lib/glyphsLib/__init__.py
+++ b/Lib/glyphsLib/__init__.py
@@ -26,6 +26,7 @@ from glyphsLib.interpolation import interpolate, build_designspace
 from glyphsLib.parser import load, loads
 from glyphsLib.writer import dump, dumps
 from glyphsLib.util import write_ufo
+import glyphsLib.classes
 
 from glyphsLib.classes import *
 
@@ -39,6 +40,7 @@ __all__ = [tostr(s) for s in (
     "build_masters", "build_instances", "load_to_ufos",
     "load", "loads", "dump", "dumps",
 )]
+__all__.extend(glyphsLib.classes.__all__)
 
 logger = logging.getLogger(__name__)
 
--- a/Lib/glyphsLib/builder/features.py
+++ b/Lib/glyphsLib/builder/features.py
@@ -18,6 +18,7 @@ from __future__ import (print_function, division, absolute_import,
 from fontTools.misc.py23 import round, unicode
 
 import glyphsLib
+import glyphsLib.glyphdata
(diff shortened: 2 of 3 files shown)


anthrotype:
the question is.. why we need to do that explicitly for this module and not for the others?

anthrotype:
and why wasn't this being caught by the test runner? 
apparently that line 43 of glyphsLib/builder/glyph.py is covered...
https://codecov.io/gh/googlei18n/glyphsLib/src/master/Lib/glyphsLib/builder/glyph.py#L43

moyogo:
Should we only import glyphsLib.glyphdata (and indirectly glyphsLib.glyphdata_generated) in the functions that use it? So we don’t load glyphsLib.glyphdata_generated when we don’t need it?

belluzj:
I guess that one of the tests imports it explicitly (`glyphdata_test.py`) so it is in the list of imported stuff for all tests.

On the other hand, we don't use that module in __init__.py so it was never included explicitly anywhere when just doing "import glyphsLib" in the REPL

Lib/glyphsLib/builder/features.py
```diff
@@ -68,6 +68,8 @@ def to_ufo_features(self, ufo):
 
 def _build_gdef(ufo):
     """Build a table GDEF statement for ligature carets."""
+    import glyphsLib.glyphdata  # Expensive import
```
anthrotype: I think `from glyphsLib import glyphdata` would be better. it spares you a extra attribute lookup

belluzj: ok

anthrotype:
since you're fixing import statements, I noticed in the `glyphsLib/__init__.py` we do a `from glyphsLib.classes import *`.
Is that really needed? Maybe we should only import `GSFont` from the classes module, and add `GSFont` to the `__all__` list of the top-level `glyphsLib` module. This is just for convencience of doing `from glyphsLib import GSFont`.
If users need some other specific class, they should import it from `glyphslib.classes` I think.

moyogo:
Hmm I don’t know. It’s not like one can do `font.newMaster()`. You have no choice but doing `font.masters.append(GSFontMaster())` or something like it.

anthrotype:
Ok, maybe it's fine if we do the star import.
But then, the `glyphsLib.__all__` shouldn't it be extended to also have the list of names that are exported in the `glyphsLib.classes.__all__`?

Lib/glyphsLib/__init__.py
```diff
@@ -39,6 +39,7 @@
     "build_masters", "build_instances", "load_to_ufos",
     "load", "loads", "dump", "dumps",
 )]
+__all__.extend(glyphsLib.classes.__all__)
```
anthrotype: I get `NameError: name 'glyphsLib' is not defined`

belluzj: I didn't even test it :/ sorry, will try again!

anthrotype: on py27, you also need to encode the `unicode_literals` with `tostr`.
I suggest you do

```python
from glyphsLib.classes import __all__ as __all_classes__

__all__ = [tostr(s) for s in [
    ...
] + __all_classes__]
```

belluzj: OK, should I do it also in classes.py? It has `unicode_literals` but doesn't take any precautions with `__all__`

anthrotype: otherwise if you attempt to do `from glyphsLib import *` you get
```
TypeError: Item in ``from list'' must be str, not unicode
```

yeah, another weird python 2.7 bug.. https://bugs.python.org/issue21720

anthrotype: >  should I do it also in classes.py? It has unicode_literals but doesn't take any precautions with __all__

that's the weird thing.. Even if the classes submodule imports unicode_literals, doing `from glyphsLib.classes import *` works without that TypeError. Only when it's in the top-level `__init__.py` it does that :-/

anthrotype:
@belluzj ok! thanks for your patience. We can merge this now :+1:

PR #270, opened 2017-11-06
[custom_params] cast xHeight value to int

anthrotype:
Fixes https://github.com/googlei18n/glyphsLib/issues/268

### DIFF
--- a/Lib/glyphsLib/builder/custom_params.py
+++ b/Lib/glyphsLib/builder/custom_params.py
@@ -121,8 +121,9 @@ def set_custom_params(ufo, parsed=None, data=None, misc_keys=(), non_info=()):
         if name.startswith('openTypeOS2Win') and value < 0:
             value = -value
 
-        # The value of these could be a float, and ufoLib/defcon expect an int.
-        if name in ('openTypeOS2WeightClass', 'openTypeOS2WidthClass'):
+        # The value of these could be a float or str, and UFO expects an int.
+        if name in ('openTypeOS2WeightClass', 'openTypeOS2WidthClass',
+                    'xHeight'):
             value = int(value)
 
         if name == 'glyphOrder':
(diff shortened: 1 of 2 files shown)


anthrotype:
we can't add "xHeight" to `GSCustomParameter._CUSTOM_INT_PARAMS` list, otherwise we can't roundtrip *.glyphs files that contain it (it would lose the quotes), so I think this is the only solution for now.
Merging.

moyogo:
Coming back to this, I don't think we should support this custom parameter if it's not supported by Glyphs.app.

Currently those files have the Arabic xHeight defined as the font xHeight attribute and then redefine the non standard xHeight custom parameter in each instance with the same value.

The source files should be changed to have the correct font xHeight attribute, and use the fontMaster xHeight string to define the Arabic xHeight.

anthrotype:
What is a “fontMaster xHeight string”?

moyogo:
> What is a “fontMaster xHeight string”?

The fontMaster `xHeight` customParameter string; for example xHeight = "arabic: 325", that Glyphs.app uses to draw the xHeight on Arabic glyphs.

In NotoSansArabic-MM.glyphs, it seems the master.xHeight was used by designers to have the Arabic xHeight line drawn at 325. That is not the correct way to do it, at least not in the latest versions of Glyphs.app. Instead, use master.customParameters["xHeight"] = "arabic: 325". That will draw the xHeight line at 325 in the Arabic glyphs.
Then the master.xHeight can be what the xHeight should be in the final TTF, and the non official instance.customParameters["xHeight"] can be deleted since master.xHeight will have the right value.

The instance.customParameters["xHeight"] is not used by Glyphs.app, we shouldn’t use it either.

moyogo:
See https://github.com/googlei18n/glyphsLib/issues/268#issuecomment-342292338

anthrotype:
Thanks Denis for the clarification. Is this documented anywhere or we just have that comment from Georg?

PR #277, opened 2017-11-07
New properties

schriftgestalt:


Lib/glyphsLib/classes.py
```diff
@@ -1157,8 +1157,10 @@ class GSFontMaster(GSBase):
         "descender": float,
         "guideLines": GSGuideLine,
         "horizontalStems": int,
+        "iconName": str,
         "id": str,
         "italicAngle": float,
+        "name": str,
```
moyogo: Both master.iconName and master.name should be unicode, no?

schriftgestalt: The `name` might be better of being unicode, but the iconName is not user accessible and only contains ascii.

belluzj:
Closed in favor of #286 and #288

PR #283, opened 2017-11-14
Skip dangling layers and layers without a name on UFO building

madig:
This change detects and skips layers that are not associated with an
existing master and layers without a name. The former are a sort of data
corruption. Empty layer names are an error as per the UFO specification
and choke ufoLib. Both may happen when you copy glyphs from a different
font in Glyphs, Glyphs simply ignores them.

### DIFF
--- a/Lib/glyphsLib/builder/builders.py
+++ b/Lib/glyphsLib/builder/builders.py
@@ -90,6 +90,13 @@ class UFOBuilder(object):
             return self._ufos.values()
         kerning_groups = {}
 
+        # Store set of actually existing master (layer) ids. This helps with
+        # catching dangling layer data that Glyphs may ignore, e.g. when
+        # copying glyphs from other fonts with, naturally, different master
+        # ids. Note: Masters have unique ids according to the Glyphs
+        # documentation and can therefore be stored in a set.
+        master_layer_ids = {m.id for m in self.font.masters}
+
         # stores background data from "associated layers"
         supplementary_layer_data = []
 
@@ -123,9 +130,10 @@ class UFOBuilder(object):
 
                 assoc_id = layer.associatedMasterId
                 if assoc_id != layer.layerId:
-                    if layer_name is not None:
-                        supplementary_layer_data.append(
-                            (assoc_id, glyph_name, layer_name, layer))
+                    # Store all layers, even the invalid ones, and just skip
+                    # them and print a warning below.
+                    supplementary_layer_data.append(
+                        (assoc_id, glyph_name, layer_name, layer))
                     continue
 
                 ufo = self._ufos[layer_id]
@@ -134,6 +142,23 @@ class UFOBuilder(object):
 
         for layer_id, glyph_name, layer_name, layer_data \
                 in supplementary_layer_data:
+            if layer_data.layerId not in master_layer_ids \
+            and layer_data.associatedMasterId not in master_layer_ids:
+                logger.warn(
+                    '{}, glyph "{}": Layer "{}" is dangling and will be '
+                    'skipped. Did you copy a glyph from a different font? If '
+                    'so, you should clean up any phantom layers not associated '
+                    'with an actual master.'.format(
+                        self.font.familyName, glyph_name, layer_data.layerId))
+                continue
+
+            if not layer_name:
+                # Empty layer names are invalid according to the UFO spec.
+                logger.warn(
+                    '{}, glyph "{}": Contains layer without a name which will '
+                    'be skipped.'.format(self.font.familyName, glyph_name))
+                continue
+
             ufo_font = self._ufos[layer_id]
             if layer_name not in ufo_font.layers:
                 ufo_layer = ufo_font.newLayer(layer_name)
--- a/tests/builder_test.py
+++ b/tests/builder_test.py
@@ -20,6 +20,8 @@ from __future__ import (print_function, division, absolute_import,
 import collections
 import datetime
 from textwrap import dedent
+import io
+import logging
 import unittest
 # unittest.mock is only available for python 3.3+
 try:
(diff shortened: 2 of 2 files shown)


madig:
Signed the CLA.

Lib/glyphsLib/builder/builders.py
```diff
@@ -90,6 +90,12 @@ def masters(self):
             return self._ufos.values()
         kerning_groups = {}
 
+        # Store list of actually existing master (layer) ids. This helps with
+        # catching dangling layer data that Glyphs may ignore, e.g. when
+        # copying glyphs from other fonts with, naturally, different master
+        # ids.
+        master_layer_ids = [m.id for m in self.font.masters]
```
anthrotype: since you don't care about the order but use this just for checking if something is inside it, you should use a set rather than a list `{m.id for m in self.font.masters}`.
I think (hope) that something that is called `id` is unique...

Lib/glyphsLib/builder/builders.py
```diff
@@ -134,6 +140,22 @@ def masters(self):
...
+            and layer_data.associatedMasterId not in master_layer_ids:
+                logger.warn(
+                    '{}, glyph "{}": Layer "{}" is dangling and will be '
+                    'skipped. Did you copy a glyph from a different font? If '
+                    'so, you should clean up any phantom layers not associated '
+                    'with an actual master.'.format(
+                        self.font.familyName, glyph_name, layer_data.layerId))
+                continue
+
+            if not layer_name:
```
belluzj: This condition is almost redundant with the condition on line 132 (`if layer_name is not None:`), unless if we have a layer with name == empty string.

Please remove the condition on line 132, so that all layers go into `supplementary_layer_data` and so that we get your nicely formatted warning for all dropped layers.

anthrotype:
Thanks. I honestly don't know how these layerId or associatedMasterID relate to each other or are supposed to work so I defer to @moyogo or @belluzj who are more knowledgeable about this stuff.

Also, we would need this to be covered by some tests.

belluzj:
@madig Could you add a unit test with a dangling layer and another with layers with no name or a blank name in builder_test.py? Each test could go like: create a minimal font, add a glyph with offending layers, run `to_ufos` while capturing the logs/stdout and check that your warning is printed.

Lib/glyphsLib/builder/builders.py
madig: https://github.com/schriftgestalt/GlyphsSDK/blob/master/GlyphsFileFormat.md says: "id string a unique id that connects the layers (associated ID) with the master."

schriftgestalt:
In Glyphs, master layers have layerID==associatedMasterID and no name (or if they happen to have one, it is ignored) The name is taken from the fontMaster.

madig:
Finally found some time again. I stumbled over the following while writing a test:
```
>>> font = GSFont()
>>> font.masters.append(GSFontMaster())
>>> font.glyphs.append(GSGlyph("a"))
>>> font.glyphs[0].layers
()
>>> font.glyphs[0].layers.append(GSLayer())
>>> font.glyphs[0].layers
(None)
>>> font.glyphs[0].layers.append(GSLayer())
>>> font.glyphs[0].layers
(None, <GSLayer "" (a)>)
```
Why is that?

madig:
So this:
```
        self.font.glyphs[0].layers[0].layerId = "yyy"
        self.font.glyphs[0].layers[0].associatedMasterId = "xxx"
```
Actually turns the layer in question into a None. Is this intentional?

belluzj:
Yes none of the results that you get make any sense. 

I have tried the code from https://github.com/googlei18n/glyphsLib/pull/283#issuecomment-345529691 in Glyphs 1088
```
>>> font = GSFont()
>>> font.masters.append(GSFontMaster())
>>> font.glyphs.append(GSGlyph("beta"))
>>> print(font.glyphs[0].layers)
(<GSLayer "Regular" (beta)>, <GSLayer "Regular" (beta)>)

>>> font.glyphs[0].layers.append(GSLayer())
>>> print(font.glyphs[0].layers)
(<GSLayer "Regular" (beta)>, <GSLayer "Regular" (beta)>, <GSLayer "orphan" (beta)>)

>>> font.glyphs[0].layers.append(GSLayer())
>>> print(font.glyphs[0].layers)
(<GSLayer "Regular" (beta)>, <GSLayer "Regular" (beta)>, <GSLayer "orphan" (beta)>, <GSLayer "orphan" (beta)>)
```

So it looks like that by default, an empty glyph has already two layers, and that when you add layers they are "orphan". I will investigate a bit more into what "orphan" layers are.

The code from your comment https://github.com/googlei18n/glyphsLib/pull/283#issuecomment-345544882 only makes sense if the id "yyy" for the associated master indeed refers to an existing master. I think it should raise if that's not the case, and "do the right thing" if there is a master with that id. I'm not exactly sure what "the right thing" is, I will investigate that too.

Maybe if some of the operations here don't really make sense from a business-logic point of view, we could just raise and prevent the user from performing them. That would be better than creating nonsense None layers.

@schriftgestalt Do you have advice on how to handle those layer manipulations?

madig:
(Raising on nonsensical id assignments would make testing this PR harder I think ;) )

I haven't had the time yet to find out where the None is coming from, but I think it shouldn't be there. I suppose an orphan is a layer not associated with an existing master, that's what I want to test for.

anthrotype:
what's the status of this?

madig:
I think we were waiting for @schriftgestalt to tell us about layer handling. That might end up in a different PR though, so I guess this can be mer— ah dang, a merge conflict.

anthrotype:
github is still saying there's a merge conflict

madig:
I think I messed up. Uh... new PR?

schriftgestalt:
I just made some tests. If you open a .glyphs file in Glyphs and there is no `associatedMasterId`, then the masterID from the first master is used. 
Empty names don't do anything, those layers are just keep around.

madig:
(I made a new PR as I messed up this one, https://github.com/googlei18n/glyphsLib/pull/291 -- please continue the discussion there.)

PR #284, opened 2017-11-14
Set hhea caretSlopeRise and caretSlopeRun based on italicAngle value

punchcutter:
It appears that Glyphs (or makeotf?) calculates `caretSlopeRise` and `caretSlopeRun` from the italicAngle, but glyphsLib doesn't do this anywhere. ufo2ft checks for `openTypeHheaCaretSlopeRise` and `openTypeHheaCaretSlopeRun` so all we need to do for the default is to add these to the ufo. 
I tested on a few files and this gives me the same hhea values that are generated when exported from Glyphs. 
@schriftgestalt does this look correct to you?

### DIFF
--- a/Lib/glyphsLib/builder/font.py
+++ b/Lib/glyphsLib/builder/font.py
@@ -17,6 +17,7 @@ from __future__ import (print_function, division, absolute_import,
 
 from collections import deque, OrderedDict
 import logging
+import math
 
 from .common import to_ufo_time
 from .constants import GLYPHS_PREFIX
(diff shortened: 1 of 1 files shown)


anthrotype:
@moyogo if this looks correct to you, shall we put something like that in ufo2ft fontInfoData as a fallback when the hhea carte values are not explicitly defined?

schriftgestalt:
In Glyphs, those values are still calculated by makeOTF.

punchcutter:
@schriftgestalt Ok, that's what it looked like. Here's the makeotf code from makeotf hhea.c which is basically the same thing. 
```
    if (g->font.ItalicAngle == 0) {
        h->tbl.caretSlopeRise       = 1;
        h->tbl.caretSlopeRun        = 0;
    }
    else {
        h->tbl.caretSlopeRise       = 1000;
        h->tbl.caretSlopeRun        =
            (short)RND(tan(FIX2DBL(-g->font.ItalicAngle) / RAD_DEG) * 1000);

```

anthrotype:
Zachary, could you please open a similar pull request for ufo2ft? I prefer to have this in there instead.
You’d would define a new special fallback function inside ufo2ft.fontInfoData module.
Thanks

moyogo:
+1 for having this in ufo2ft.

punchcutter:
Yes, ufo2ft makes way more sense. I was thinking it was done by Glyphs, but now I see it's all makeotf so I'll close this and open a new pull request for ufo2ft.

PR #285, opened 2017-11-17
Read colors and set UFO keys for Glyphs ColorIndex and public.markColor

punchcutter:
As mentioned here: https://github.com/googlei18n/glyphsLib/issues/116

With colors coming from outside Glyphs we were failing on `GLYPHS_COLORS[color_index]` when color_index wasn't in GLYPHS_COLORS. 
Here we should check for a tuple or a Glyphs color index and set the keys accordingly. Using the FontLab export script to make a Glyphs file results in tuples like (235,0,255,1), but the UFO spec requires integer or floats between 0 and 1 (http://unifiedfontobject.org/versions/ufo3/conventions/#colors). So check here to make sure we get what we expect. 

I'm not sure about how many decimal places we care about so put 4 for now.

### DIFF
--- a/Lib/glyphsLib/builder/glyph.py
+++ b/Lib/glyphsLib/builder/glyph.py
@@ -14,6 +14,8 @@
 
 from __future__ import (print_function, division, absolute_import,
                         unicode_literals)
+import logging
+logger = logging.getLogger(__name__)
 
 import glyphsLib
 from .common import to_ufo_time
@@ -35,9 +37,21 @@ def to_ufo_glyph(self, ufo_glyph, layer, glyph_data):
     if last_change is not None:
         ufo_glyph.lib[GLYPHLIB_PREFIX + 'lastChange'] = to_ufo_time(last_change)
     color_index = glyph_data.color
-    if color_index is not None and color_index >= 0:
+    if color_index is not None:
         ufo_glyph.lib[GLYPHLIB_PREFIX + 'ColorIndex'] = color_index
-        ufo_glyph.lib[PUBLIC_PREFIX + 'markColor'] = GLYPHS_COLORS[color_index]
+        color_tuple = None
+        if isinstance(color_index, list):
+            if not all(i in range(0, 256) for i in color_index):
+                logger.warn('Invalid color tuple {} for glyph {}. '
+                            'Values must be in range 0-255'.format(color_index, glyph_data.name))
+            else:
+                color_tuple = ','.join('{0:.4f}'.format(i/255) if i in range(1, 255) else str(i//255) for i in color_index)
+        elif isinstance(color_index, int) and color_index in range(len(GLYPHS_COLORS)):
+            color_tuple = GLYPHS_COLORS[color_index]
+        else:
+            logger.warn('Invalid color index {} for {}'.format(color_index, glyph_data.name))
+        if color_tuple is not None:
+            ufo_glyph.lib[PUBLIC_PREFIX + 'markColor'] = color_tuple
     export = glyph_data.export
     if not export:
         ufo_glyph.lib[GLYPHLIB_PREFIX + 'Export'] = export
--- a/tests/builder_test.py
+++ b/tests/builder_test.py
@@ -1131,5 +1131,40 @@ class DrawPathsTest(unittest.TestCase):
         self.assertEqual(first_segment_type, 'qcurve')
 
 
+class GlyphPropertiesTest(unittest.TestCase):
+
+    def test_glyph_color(self):
+        font = generate_minimal_font()
+        glyph = GSGlyph(name='a')
+        glyph2 = GSGlyph(name='b')
+        glyph3 = GSGlyph(name='c')
+        glyph4 = GSGlyph(name='d')
+        glyph.color = [244, 0, 138, 1]
+        glyph2.color = 3
+        glyph3.color = 88
+        glyph4.color = [800, 0, 138, 1]
+        font.glyphs.append(glyph)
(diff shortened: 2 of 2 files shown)


moyogo:
Cool! Could you add tests?

moyogo:
Oh, and could you remove the italic commits?

punchcutter:
Definitely. I totally forgot to rebase after moving the italic stuff to ufo2ft. I'll add some tests, too.

Lib/glyphsLib/builder/glyph.py
```diff
@@ -35,9 +35,16 @@ def to_ufo_glyph(self, ufo_glyph, layer, glyph_data):
     if last_change is not None:
         ufo_glyph.lib[GLYPHLIB_PREFIX + 'lastChange'] = to_ufo_time(last_change)
     color_index = glyph_data.color
-    if color_index is not None and color_index >= 0:
+    if color_index is not None:
         ufo_glyph.lib[GLYPHLIB_PREFIX + 'ColorIndex'] = color_index
-        ufo_glyph.lib[PUBLIC_PREFIX + 'markColor'] = GLYPHS_COLORS[color_index]
+        color_tuple = None
+        if isinstance(color_index, list):
+            color_tuple = ','.join(['{0:.4f}'.format(i/255.0)
```
anthrotype: Note that we are using python 3 division operator (or if we don’t we should, by importing it `from __future__`.
You need to use two slashes for integer division; one slash is always float division.

Lib/glyphsLib/builder/glyph.py
```diff
@@ -35,9 +35,16 @@ def to_ufo_glyph(self, ufo_glyph, layer, glyph_data):
...
+                                    if i in range(1, 255) else str(int(i/255)) for i in color_index])
+        elif isinstance(color_index, int) and color_index in range(len(GLYPHS_COLORS)):
+            color_tuple = GLYPHS_COLORS[color_index]
+        if color_tuple is not None:
```
anthrotype: Shouldn’t we at least log a warning that we dropped some data even though invalid or unsupported that we could not fit?

Lib/glyphsLib/builder/glyph.py
anthrotype: The join method of str accept an iterable, so you can pass it a generator expression directly without needing to create a temporary list (with a list comprehension).
“”.join(x for x in something)

Lib/glyphsLib/builder/glyph.py
```diff
@@ -35,9 +35,16 @@ def to_ufo_glyph(self, ufo_glyph, layer, glyph_data):
...
         ufo_glyph.lib[GLYPHLIB_PREFIX + 'lastChange'] = to_ufo_time(last_change)
     color_index = glyph_data.color
-    if color_index is not None and color_index >= 0:
+    if color_index is not None:
         ufo_glyph.lib[GLYPHLIB_PREFIX + 'ColorIndex'] = color_index
-        ufo_glyph.lib[PUBLIC_PREFIX + 'markColor'] = GLYPHS_COLORS[color_index]
+        color_tuple = None
+        if isinstance(color_index, list):
+            color_tuple = ','.join(['{0:.4f}'.format(i/255.0)
+                                    if i in range(1, 255) else str(int(i/255)) for i in color_index])
```
anthrotype: Curious how did you come up with this formula?

Lib/glyphsLib/builder/glyph.py
punchcutter: Well, the parser won't allow most invalid things to even get to this point, but it is possible to manually put in some bad values like a tuple of (344, -400, 23, 1) so we can warn for that.  A Glyphs color index of > 11 will fail because GLYPHS_COLORS only has 12 elements. I put in some warnings for either of those cases.

Lib/glyphsLib/builder/glyph.py
punchcutter: A little trial and error. We know it needs to be in the range 0-255, but if we don't adjust the precision then the UFO gets a bunch of ugly values like 0.0,0.952941176471,1.0,0.00392156862745
If we set precision to 4 places we get 0.0000,0.9529,1.0000,0.0039 which is still kind of ugly. That's why the weird formula to only give 4 place precision for anything between 0 and 1, but for 0 and 1 we don't need anything so ignore it on those.

punchcutter:
Ok, added some warnings and tests. There didn't seem to be a place for this kind of test so I added GlyphPropertiesTest in the builder tests which I figure could test other GSGlyph specific properties as well.

schriftgestalt:
I added the `color` to the glyphs file spec. 

The `glyph.color` can be set from the UI and it will get a int number. But you can also set it to any RGBA value by script or from a .ufo. That’s where the tuple comes from.

tests/builder_test.py
```diff
@@ -1131,5 +1131,40 @@ def test_to_ufo_draw_paths_qcurve(self):
         self.assertEqual(first_segment_type, 'qcurve')
 
 
+class GlyphPropertiesTest(unittest.TestCase):
```
anthrotype: you're testing `to_ufos` function (or that what you are calling in the test), so this `test_glyph_color` method should go to `ToUfosTest` class, I think.

@belluzj wdyt?

belluzj: To me the current class name is good because it spells out the intent of the tests (test glyph properties) rather than the technical details (the test calls to_ufos). Beyond that, I don't know what the best practice is for organizing tests in python...

PR #286, opened 2017-11-18
Add name setter so Glyphs 2.5b "name" property in fontMaster doesn't conflict

punchcutter:
https://github.com/googlei18n/fontmake/issues/374
In Glyphs 2.5b fontMaster has "name" and "iconName" properties. The name property conflicts with the GSFontMaster name property in the parser.

### DIFF
--- a/Lib/glyphsLib/classes.py
+++ b/Lib/glyphsLib/classes.py
@@ -399,7 +399,7 @@ class FontFontMasterProxy(Proxy):
             raise(KeyError)
 
     def __setitem__(self, Key, FontMaster):
-        FontMaster.parent = self._owner
+        FontMaster.font = self._owner
         if type(Key) is int:
             OldFontMaster = self.__getitem__(Key)
             if Key < 0:
@@ -414,7 +414,6 @@ class FontFontMasterProxy(Proxy):
         else:
             raise(KeyError)
 
-
     def __delitem__(self, Key):
         if type(Key) is int:
             if Key < 0:
@@ -428,7 +427,7 @@ class FontFontMasterProxy(Proxy):
         return self._owner._masters
 
     def append(self, FontMaster):
-        FontMaster.parent = self._owner
+        FontMaster.font = self._owner
         FontMaster.id = str(uuid.uuid4()).upper()
         self._owner._masters.append(FontMaster)
 
@@ -438,7 +437,6 @@ class FontFontMasterProxy(Proxy):
                 newLayer = GSLayer()
                 glyph._setupLayer(newLayer, FontMaster.id)
                 glyph.layers.append(newLayer)
-
 
     def remove(self, FontMaster):
 
@@ -451,7 +449,7 @@ class FontFontMasterProxy(Proxy):
         self._owner._masters.remove(FontMaster)
 
     def insert(self, Index, FontMaster):
-        FontMaster.parent = self._owner
+        FontMaster.font = self._owner
         self._owner._masters.insert(Index, FontMaster)
 
     def extend(self, FontMasters):
@@ -463,7 +461,7 @@ class FontFontMasterProxy(Proxy):
             values = list(values)
         self._owner._masters = values
         for m in self._owner._masters:
-            m.parent = self._owner
+            m.font = self._owner
 
 
 class FontGlyphsProxy(Proxy):
@@ -1157,10 +1155,10 @@ class GSFontMaster(GSBase):
         "descender": float,
         "guideLines": GSGuideLine,
         "horizontalStems": int,
+        "iconName": str,
         "id": str,
         "italicAngle": float,
         "name": str,
-        "iconName": str,
         "userData": dict,
         "verticalStems": int,
         "visible": bool,
@@ -1200,10 +1198,10 @@ class GSFontMaster(GSBase):
         "descender",
         "guideLines",
         "horizontalStems",
+        "iconName",
         "id",
         "italicAngle",
         "name",
-        "iconName",
         "userData",
         "verticalStems",
         "visible",
@@ -1215,8 +1213,8 @@ class GSFontMaster(GSBase):
     )
 
     def __init__(self):
-        self._font = None
         super(GSFontMaster, self).__init__()
+        self._font = None
         self._name = None
         self._customParameters = []
         self._weight = "Regular"
@@ -1240,9 +1238,9 @@ class GSFontMaster(GSBase):
             # Always write those values
             return True
         if key == "name":
-            if self.font and int(self.font.appVersion) >= 1075:
-                return True
-            return False
+            if getattr(self, key) == "Regular":
+                return False
+            return True
         return super(GSFontMaster, self).shouldWriteValueForKey(key)
 
     @property
@@ -1275,10 +1273,7 @@ class GSFontMaster(GSBase):
 
     @name.setter
     def name(self, value):
-        if self.font and int(self.font.appVersion) >= 1075:
-            self._name = value
-        else:
-            return
+        self._name = value
 
     customParameters = property(
         lambda self: CustomParametersProxy(self),
@@ -2771,9 +2766,8 @@ class GSFont(GSBase):
                 logger.info('Parsing .glyphs file into %r', self)
                 p.parse_into_object(self, fp.read())
             self.filepath = path
-
-        for master in self.masters:
-            master.font = self
+            for master in self.masters:
+                master.font = self
 
     def __repr__(self):
         return "<%s \"%s\">" % (self.__class__.__name__, self.familyName)
--- a/Lib/glyphsLib/classes.py
+++ b/Lib/glyphsLib/classes.py
@@ -1159,6 +1159,8 @@ class GSFontMaster(GSBase):
         "horizontalStems": int,
         "id": str,
         "italicAngle": float,
+        "name": str,
+        "iconName": str,
         "userData": dict,
         "verticalStems": int,
         "visible": bool,
@@ -1200,6 +1202,8 @@ class GSFontMaster(GSBase):
         "horizontalStems",
         "id",
         "italicAngle",
+        "name",
+        "iconName",
         "userData",
         "verticalStems",
         "visible",
@@ -1211,6 +1215,7 @@ class GSFontMaster(GSBase):
     )
 
     def __init__(self):
+        self._font = None
         super(GSFontMaster, self).__init__()
         self._name = None
         self._customParameters = []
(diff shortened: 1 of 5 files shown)


punchcutter:
Oops, not so simple. Need to do a couple things first:
To keep the parser roundtrip working we need to return True for the "name" key in shouldWriteValueForKey, but only for appVersion > 1075. But GSFontMaster doesn't have a font attribute like it does in Glyphs. 

So to handle both 2.5 and pre-2.5 Glyphs files we'd have to check for appVersion in some of the tests and return different things. Some other things that have changed:

GSNodeFromFileTest.test_userData GSNode now returns string instead of int
TTFStems custom parameter now also have "horizontal = 1;" in the Glyphs file  

This test is problematic:
```
    def test_font_master_is_name_not_writable(self):
        """Match the Glyphs python API"""
        with self.assertRaises(AttributeError):
            self.font.masters[0].name = "Test"
```
In Glyphs 2.5 we can set the name so this test fails if we are able to set the name.

punchcutter:
https://github.com/googlei18n/glyphsLib/issues/267

Lib/glyphsLib/classes.py
```diff
@@ -1234,8 +1239,20 @@ def shouldWriteValueForKey(self, key):
         if key in ("xHeight", "capHeight", "ascender"):
             # Always write those values
             return True
+        if key == "name":
+            if self.font and int(self.font.appVersion) >= 1075:
```
anthrotype: if the appVersion type is a string, instead of simply an integer, I guess the reason is that it may (or may have) contained some non-numeric characters in it. Unconditionally casting it to int may fail with ValueError if the string contains non-numeric characters.  
Question for @schriftgestalt: does "appVersion" string always only contain numbers? If so, why do you have to type it as a string? On the other hand, if it _may_ also contain non-numbers, what exactly is the format of "appVersion", so that we can have a function that converts it into a tuple in order to correctly do comparison between them?

Lib/glyphsLib/classes.py
```diff
@@ -1234,8 +1239,20 @@ def shouldWriteValueForKey(self, key):
...
+                return True
+            return False
```
anthrotype: I don't think we need to check the appVersion of the original file to decide whether to write the "name" attribute.
If the attribute is not None (because the file comes from new Glyphs), then we write it, otherwise we need to do what Glyphs.app does when opening and re-saving a file generated from an older version of the app. I don't know the answer to that (don't a mac at hand to try this out ATM).
Basically, we should "upgrade" the source files to whatever latest version of Glyphs.app that we currently match at any given time. Instead of attempting to ensure we roundtrip any file generated from previous versions. I think this only adds more complexity.

Lib/glyphsLib/classes.py
```diff
@@ -1253,8 +1270,16 @@ def name(self):
...
             name = " ".join(list(names))
+        self._name = name
         return name
 
+    @name.setter
+    def name(self, value):
+        if self.font and int(self.font.appVersion) >= 1075:
+            self._name = value
+        else:
+            return
```
anthrotype: same here. If I set a value, I do set that value, regardless of the input file's appVersion. 
Basically, the appVersion of the input file should only be relevant while parsing, and kept in the file for info only; but it should not (I believe) determine the internal structure of the data nor the output format.

belluzj:
I agree with the general comments from Cosimo, that glyphsLib should track the "latest and greatest" from Glyphs.app. It was discussed here: https://github.com/googlei18n/glyphsLib/pull/227#issuecomment-335513220

tests/classes_test.py
```diff
@@ -1399,7 +1399,11 @@ def test_name(self):
         self.assertEqual(self.node.name, 'Hello')
 
     def test_userData(self):
-        self.assertEqual(1, self.node.userData["rememberToMakeCoffee"])
+        if int(self.font.appVersion) >= 1075:
```
belluzj: The unit tests should not contain any logic like this. The input file is supposed to be fixed, and the test should only have code that test the current input file (GlyphsUnitTestSans). If we wanted to support several versions of Glyphs, I think we should have several version of GlyphsUnitTestSans and several tests files, one per version, that test specific aspects of that version. But anyway, I don't think we want to keep the API compatible with several versions of Glyphs.

Lib/glyphsLib/classes.py
```diff
@@ -1234,8 +1239,20 @@ def shouldWriteValueForKey(self, key):
         if key in ("xHeight", "capHeight", "ascender"):
             # Always write those values
             return True
+        if key == "name":
+            if self.font and int(self.font.appVersion) >= 1075:
+                return True
+            return False
         return super(GSFontMaster, self).shouldWriteValueForKey(key)
 
+    @property
```
belluzj: In Glyphs 1088, `font` is a regular method, not a property (the property access returns an "objective C native selector")

Lib/glyphsLib/classes.py
```diff
@@ -2747,6 +2772,9 @@ def __init__(self, path=None):
                 p.parse_into_object(self, fp.read())
             self.filepath = path
 
+        for master in self.masters:
+            master.font = self
```
belluzj: Actually there is already a mechanism that links the masters to the font using the `parent` property, which is set in `FontMasterProxy`here:
https://github.com/punchcutter/glyphsLib/blob/c089739b2b67a12436291df38f5126fe852f9f7a/Lib/glyphsLib/classes.py#L402

Unfortunately the `parent` property is not documented at http://docu.glyphsapp.com/#gsfontmaster and doesn't work in Glyphs.app:
```
>>> print(Glyphs.font.masters[0].font().masters[0])
<GSFontMaster "Regular" width 100.0 weight 100.0>
>>> print(Glyphs.font.masters[0].font.masters[0])
  File "<string>", line 5, in <module>
AttributeError: 'objc.native_selector' object has no attribute 'masters'
>>> print(Glyphs.font.masters[0].parent.masters[0])
  File "<string>", line 5, in <module>
AttributeError: 'NSKVONotifying_GSFontMaster' object has no attribute 'parent'
```

Maybe the correct way (in the sense that is matches Glyphs' behaviour and minimizes the amount of code that you add) would be to have a `font()` method that returns `self.parent`?

Lib/glyphsLib/classes.py
anthrotype: what do you mean a "method"? A callable as in `master.font()`?
we shouldn't expose or try to mimic the the objc interface here

belluzj:
In the end, if you remove the version-specific logic and the roundtrip test on GlyphsUnitTestSans crashes, you can open & save the GlyphsUnitTestSans with the latest version of Glyphs, so that we test with the latest version only.

Lib/glyphsLib/classes.py
belluzj: Yes a callable:
```python
>>> print(Glyphs.font.masters[0].font().masters[0])
<GSFontMaster "Regular" width 100.0 weight 100.0>
>>> print(Glyphs.font.masters[0].font.masters[0])
  File "<string>", line 5, in <module>
AttributeError: 'objc.native_selector' object has no attribute 'masters'
```

I agree that it would be much better to have a method, but then the scripts written for glyphsLib would not work in Glyphs.app (is that still a goal?)

Lib/glyphsLib/classes.py
punchcutter: I’m all for keeping it simple and just supporting the latest version.

Lib/glyphsLib/classes.py
punchcutter: Yes, FontFontMasterProxy has it in the setter, but that's not being set at all.  
```
>>> font.masters
(<GSFontMaster "Light" width 100.0 weight 17.0>, <GSFontMaster "Regular" width 100.0 weight 90.0>, <GSFontMaster "Bold" width 100.0 weight 220.0>)
>>> font.masters[0]
<GSFontMaster "Light" width 100.0 weight 17.0>
>>> font.masters[0].parent
AttributeError: 'GSFontMaster' object has no attribute 'parent'
```

punchcutter:
We could just add "name" and that fixes the problem with 2.5, but it seems userData on GSNode also changed so the roundtrip fails there. The newline before and after are no longer added in Glyphs

Old format: ```"352 147 LINE SMOOTH {\nname = Hello;\nrememberToMakeCoffee = \"1\";\n}",```
New format: ```"352 147 LINE SMOOTH {name = Hello;\nrememberToMakeCoffee = \"1\";}",```

schriftgestalt:
I adding a ‘parent’ property to the master was my mistake. The more specific ‘font’ property is better and I will add that to the documentation. So the fontMasterProxy should be changed.

Lib/glyphsLib/classes.py
```diff
@@ -1234,8 +1237,20 @@ def shouldWriteValueForKey(self, key):
...
+            return True
         return super(GSFontMaster, self).shouldWriteValueForKey(key)
 
+    @property
+    def font(self):
+        return self._font
+
+    @font.setter
+    def font(self, value):
+        self._font = value
```
belluzj: In the end, `font` is a very simple property. Maybe there is no need for a getter/setter?

punchcutter: That's what I thought at first, but then I wondered about how it's possible to create a GSFontMaster that's not yet attached to a font. Loading from a Glyphs file will set the font property. Appending or inserting into the font will set the font property. But if you just do new_master = GSFontMaster() it won't have a font property at all. Not sure how common that is.

belluzj: The case you describe can work if we replace `self._font = None` with `self.font = None` in `__init__`?

punchcutter: Yes, I guess that's simpler in this case.

anthrotype:
what is the status of this?

punchcutter:
@anthrotype I just removed the font getter/setter since we don't really need them on GSFontMaster. Tests pass and everything looks ok to me.

belluzj:
Looks ok to merge.

I don't like

belluzj:
Sorry for the close/reopen. I was saying that it looks ok to merge, although I don't like the changes about the userData in the node string. But that's a bit out of scope, and I will add more tests about that in another MR.

Lib/glyphsLib/types.py
```diff
@@ -376,7 +376,8 @@ def encode_dict_as_string_for_gsnode(value):
     encoded such that it can be included in the string representation
     of a GSNode."""
     value = value.replace('"', '\\"')
-    return value.replace("\n", "\\n")
+    value = value.replace('\n', '')
+    return value.replace(";", ";\\n", value.count(';') - 1)
```
belluzj: What if one of the dict's values contain some `\n` or `;`? I think it's out of scope for this MR though, and will add more tests somewhere else.

punchcutter: Yes, sorry, I don't like this userData thing either, but updating the Glyphs file to the latest version introduced that weirdness with newlines. Please make better! Thanks!

PR #287, opened 2017-11-19
add feature to generate glyphdata_generated.py from supplied xml files

medicalwei:
I am [packaging this into Debian][1] to build some of fonts made by Glyphs.app from source. However, per Debian Policy we have to generate the glyphdata_generated.py from source.

The pull request allows loading the files instead of downloading from the repo, if there are filenames supplied after the command.

However there's a bootstrap problem that the libraries needs to be able to load in order to run the script, which means glyphdata_generated.py needs to exist and loadable from python.  It can be workarounded by using an empty file in place when generating the xml instead of using the pre-generated file.

[1]: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=868005


anthrotype:
Thanks for working on packaging glyphsLib for Debian!
I’ll do all that is needed to make your task easier.
However, I still don’t quite get what is the point of you having to re-run this script when you build the deb package and regenerating exactly the same (because it has to be the same otherwise there’s a problem) module from the same data file (whether downloaded from the upstream git repo or taken from another Debian package).
Is it about trust? Sorry if my question may sound stupid to you.

anthrotype:
I’m interested in understanding how this issue of packaging source files generated from other this party data files works, because i’m adding some files generated from Unicode Character Database to fonttools.
So I want to get this right for the beginning instead of then you having to come with downstream patches later.

medicalwei:
The reason behind generating the file is that we have to follow Debian Free
Software Guideline. In §2 it says:

The program must include source code, and must allow distribution in source
code as well as compiled form.

We take it seriously and considers the file generated is not "preferred
form of modifications", and regenerating the file is necessary. This also
applies to fonts unless we found no way to generate it from source.

Upstream actually don't need to strip the file from the release. We can
strip the files in question from the package by defining the files needed
to be stripped defined in debian/copyright. It will generate a tarball
without specific file(s). Then, when building the package there should be a
override to the debian-helper (which can automatically do packaging of
various formats) of debian/rules (a Makefile for building and packaging) to
generate the needed file before actually do packaging.

Cosimo Lupo <notifications@github.com>於 2017年11月19日 週日，19:26寫道：

anthrotype:
I see, thanks for clarifying that.

PR #288, opened 2017-11-23
Test node user data

belluzj:
This add a test for #286 and integrates pending changes from #277

### DIFF
--- a/Lib/glyphsLib/classes.py
+++ b/Lib/glyphsLib/classes.py
@@ -1158,7 +1158,7 @@ class GSFontMaster(GSBase):
         "iconName": str,
         "id": str,
         "italicAngle": float,
-        "name": str,
+        "name": unicode,
         "userData": dict,
         "verticalStems": int,
         "visible": bool,
@@ -2717,7 +2717,7 @@ class GSFont(GSBase):
     }
     _wrapperKeysTranslate = {
         ".appVersion": "appVersion",
-        "fontMaster": "_masters",
+        "fontMaster": "masters",
         "unitsPerEm": "upm",
         "gridSubDivision": "gridSubDivisions"
     }
--- a/Lib/glyphsLib/classes.py
+++ b/Lib/glyphsLib/classes.py
@@ -16,18 +16,19 @@
 # limitations under the License.
 
 from __future__ import print_function, unicode_literals
-import re, math, inspect
+
+import re
+import math
+import inspect
 import traceback
 import uuid
 import logging
 import glyphsLib
 from glyphsLib.types import (
     transform, point, rect, size, glyphs_datetime, color, floatToString,
-    readIntlist, writeIntlist, needsQuotes, feature_syntax_encode, baseType,
-    encode_dict_as_string_for_gsnode, decode_dict_as_string_from_gsnode
-)
+    readIntlist, writeIntlist, baseType)
 from glyphsLib.parser import Parser
-from glyphsLib.writer import Writer
+from glyphsLib.writer import Writer, escape_string
 from collections import OrderedDict
 from fontTools.misc.py23 import unicode, basestring, UnicodeIO, unichr, open
 from glyphsLib.affine import Affine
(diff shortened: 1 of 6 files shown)


Lib/glyphsLib/classes.py
```diff
@@ -1297,8 +1298,8 @@ def width(self, value):
 
 
 class GSNode(GSBase):
-    _rx = '([-.e\d]+) ([-.e\d]+) (LINE|CURVE|QCURVE|OFFCURVE|n/a)'\
-          '(?: (SMOOTH))?(?: (\{.*\}))?'
+    _rx = '"([-.e\d]+) ([-.e\d]+) (LINE|CURVE|QCURVE|OFFCURVE|n/a)'\
+          '(?: (SMOOTH))?(?: (\{(?:.|\\n)*\}))?"'
```
anthrotype: `re.DOTALL` flag makes the dot also match newlines
https://docs.python.org/3.7/library/re.html#re.DOTALL

Lib/glyphsLib/classes.py
```diff
@@ -1418,6 +1418,36 @@ def connection(self):
...
+        encoded such that it can be included in the string representation
+        of a GSNode."""
+        # Strip the first and last newlines
+        if value.startswith('{\n'):
+            value = '{' + value[2:]
+        if value.endswith('\n}'):
+            value = value[:-2] + '}'
+        value = value.replace('"', '\\"')
+        value = value.replace('\n', '\\n')
+        return value
```
anthrotype: I wonder if instead of all this string copies and consecutive calls to `.replace()` would be more efficient to use the re.sub approach (with a callable `repl` function that takes the match object and returns a string, like the `_decode_dict_as_string` does)

Lib/glyphsLib/writer.py
```diff
@@ -166,3 +164,64 @@ def dumps(obj):
...
+    True, True, True, True, True, True, True, True,
+    True, True, True, False, False, False, False, False
+    )
+
+
+def _needs_quotes(string):
+    if len(string) == 0:
+        return True
+    if not isinstance(string, (str, unicode)):
+        return False
```
anthrotype: the argument name is "string". How can something that is called "string" not be an instance of (str, unicode)?
also, `_needs_quotes` is called below, after the very same isinstance test, so this additional check seems redundant to me. If anything should be an assert

Lib/glyphsLib/writer.py
```diff
@@ -166,3 +164,64 @@ def dumps(obj):
...
+    # Does it need quotes because it could be confused with a number?
+    i = None
+    try:
+        i = int(string)
+    except:
+        pass
+    if i is not None:
+        return True
+
+    return False
```
anthrotype: yes, it means that if a string only contains numbers, then i must be quoted to avoid confusing it with a number number.

this is better 

```python
try:
    int(string)
except ValueError:
    return False
else:
    return True
```

Lib/glyphsLib/writer.py
```diff
@@ -166,3 +164,64 @@ def dumps(obj):
...
+
+
+def escape_string(value):
+    if isinstance(value, (str, unicode)) and _needs_quotes(value):
+        value = value.replace("\\", "\\\\")
+        value = value.replace("\"", "\\\"")
+        value = value.replace("\n", "\\012")
+        value = '"%s"' % value
```
anthrotype: same thing here, maybe re.sub with a replace callback is better than multiple replace.
it's a pity that the signature of the `translate` method of built-in strings changed between python2 and 3, because that would have been handy in these situations.

Lib/glyphsLib/classes.py
belluzj: It looks like re.sub is slower (maybe because I tested with a small input string, but I guess that most strings in that case will have only a name for the node)

https://gist.github.com/belluzj/9f7281384190a9988e17d98d5e3d1686

anthrotype: It’s good that you timed it!
Does it make any difference if instead of calling re.sub you first re.compile and then call sub method on the compiled regex object?

anthrotype: ok good!

PR #291, opened 2017-12-01
Skip dangling layers and layers without a name on UFO building

madig:
This change detects and skips layers that are not associated with an
existing master and layers without a name. The former are a sort of data
corruption. Empty layer names are an error as per the UFO specification
and choke ufoLib. Both may happen when you copy glyphs from a different
font in Glyphs, Glyphs simply ignores them.

All invalid layers now get a warning printed and skipped. Before, they
could have been silently ignored.

Add tests for both types of bogus layers.

### DIFF
--- a/Lib/glyphsLib/builder/builders.py
+++ b/Lib/glyphsLib/builder/builders.py
@@ -90,6 +90,13 @@ class UFOBuilder(object):
             return self._ufos.values()
         kerning_groups = {}
 
+        # Store set of actually existing master (layer) ids. This helps with
+        # catching dangling layer data that Glyphs may ignore, e.g. when
+        # copying glyphs from other fonts with, naturally, different master
+        # ids. Note: Masters have unique ids according to the Glyphs
+        # documentation and can therefore be stored in a set.
+        master_layer_ids = {m.id for m in self.font.masters}
+
         # stores background data from "associated layers"
         supplementary_layer_data = []
 
@@ -123,9 +130,10 @@ class UFOBuilder(object):
 
                 assoc_id = layer.associatedMasterId
                 if assoc_id != layer.layerId:
-                    if layer_name is not None:
-                        supplementary_layer_data.append(
-                            (assoc_id, glyph_name, layer_name, layer))
+                    # Store all layers, even the invalid ones, and just skip
+                    # them and print a warning below.
+                    supplementary_layer_data.append(
+                        (assoc_id, glyph_name, layer_name, layer))
                     continue
 
                 ufo = self._ufos[layer_id]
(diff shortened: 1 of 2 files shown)


schriftgestalt:
I just made some tests. If you open a .glyphs file in Glyphs and there is no associatedMasterId, then the masterID from the first master is used.
Empty names don't do anything, those layers are just keep around.

madig:
(This is a new PR as I messed up the old one: https://github.com/googlei18n/glyphsLib/pull/283 -- see there for context)

anthrotype:
> Empty names don't do anything, those layers are just keep around.

so if Glyphs.app keeps them, maybe we should also keep them when exporting to UFO? (e.g. by giving them a special "__no_name__" name, maybe with a incrmenting digit suffix to make them unique so they don't clash on the filesystem). Just a thought, not saying we should do this here right now.

anthrotype:
> If you open a .glyphs file in Glyphs and there is no associatedMasterId, then the masterID from the first master is used

... whereas in this PR, they are dropped. shouldn't we do the same as Glyphs.app does, according to what Georg sais (i.e. implicitly associate them with first master)?

I wonder, is it possible for multiple layers to be associated with the same master? and what does that actually mean? :confused:

madig:
> I wonder, is it possible for multiple layers to be associated with the same master?
I think so, just make multiple copies of the same master layer in Glyphs, they're all named by date by default :)

schriftgestalt:
>  maybe with a incrementing digit suffix to make them unique so they don't clash on the filesystem

that would be a general problems. Layer names are not required to be unique.

anthrotype:
> Layer names are not required to be unique

but the layers' folder names inside a UFO must be unique

anthrotype:
so for an un-named layer in glyphs. the corresponding layer name in the UFO would be `__no_name__`, and the folder name that this is mapped to in the `layercontents.plist` could be `__no_name__`, `__no_name__#1`, etc.

anthrotype:
just for reference, this is the issue that triggered this PR
https://github.com/googlei18n/glyphsLib/issues/282

I just noticed one of Noto fonts (noto-source/src/NotoSansTamil-MM.glyphs) also has one of these layers without a name. The glyphsLib < 2.0 would not choke on these, probably because it was not creating UFO3 layers but UFO2 robofont-style layers-inside-the-glyph-lib.

It's still not clear to me how these nameless layers actually happen, whether it's the user that deletes the name and leaves an empty string, or whether this is the result of some other action like pasting from a different font, like someone reported.

In any case, I believe for now it's OK if we warn and drop them from the UFO3, so I'm going to merge this.

Later we can see how to preserve them for the sake of roundtrippability.

madig:
If empty layer names are indeed user error, I'd petition @schriftgestalt to add guards against them in Glyphs.app.

schriftgestalt:
>I wonder, is it possible for multiple layers to be associated with the same master? and what does that actually mean?

Yes. You can have as many layers associated with a master. There should only be one "Master" layer. 

Each layer needs to know certain value that are defined in the masters. That is where the associatedMasterID is used.

anthrotype:
thanks for the explanation!

PR #298, opened 2017-12-13
adjust the API to Glyphs.app

schriftgestalt:


anthrotype:
thanks georg, sorry for getting back to this so late.
you added a property getter for gridLength, but since there isn't a corresponding setter, it is a read-only property and the tests fail with AttributeError.
If this is meant to be writable, could you please add a setter as well?

schriftgestalt:
The `gridLength` property is not writeable. Only `grid` and `gridSubDivisions`. The `gridLenght` is just the combination of the other two.

anthrotype:
good to know, then the tests need to be changed accordingly

schriftgestalt:
This is a bit confusing as there was a writeable `gridLenght` property before ;(

anthrotype:
ah, ok so the API change was to make it read-only?

schriftgestalt:
the value that was accessible in `gridLength` is now in `grid`. And then there is a new property that is called (confusingly) `gridLength`. That change changes the API to match Glyphs.app. There it developed to be like that over the years.

anthrotype:
I'm ok with sync-ing with the latest Glyphs.app API, whatever you say that is. But we need to address the test failure before we can merge.

mashabow:
Thanks Georg!

@anthrotype This is the original thread on the Glyphs forum, FYI: https://forum.glyphsapp.com/t/grid-properties-inconsistency-between-python-api-and-glyphs-file/7779

anthrotype:
ok I see now, thanks @mashabow. I'll merge this and then I fix the test (removing the line that attempts to modify gridLength, as it's now read-only).

PR #299, opened 2017-12-13
allow a date string without time zone

schriftgestalt:
Glyphs might change to not write the tz if it is +0000. That saves a
few KB on file size.

### DIFF
--- a/Lib/glyphsLib/types.py
+++ b/Lib/glyphsLib/types.py
@@ -228,7 +228,12 @@ class glyphs_datetime(baseType):
         """Parse a datetime object from a string."""
         # parse timezone ourselves, since %z is not always supported
         # see: http://bugs.python.org/issue6641
-        string, tz = src.rsplit(' ', 1)
+        srcs = src.split(' ')
+        string = srcs[0] + " " + srcs[1]
+        if len(srcs) > 2:
+            tz = srcs[2]
+        else:
+            tz = "+0000";
         if 'AM' in string or 'PM' in string:
             datetime_obj = datetime.datetime.strptime(
                 string, '%Y-%m-%d %I:%M:%S %p'


Lib/glyphsLib/types.py
```diff
@@ -228,7 +228,12 @@ def read(self, src):
         """Parse a datetime object from a string."""
         # parse timezone ourselves, since %z is not always supported
         # see: http://bugs.python.org/issue6641
-        string, tz = src.rsplit(' ', 1)
+        srcs = src.split(' ')
+        string = srcs[0] + " " + srcs[1]
+        if len(srcs) > 2:
+            tz = srcs[2]
```
anthrotype: you're assuming that the third item in the list is the time zone. Couldn't it also be the optional 'AM'/'PM' string?

Lib/glyphsLib/types.py
```diff
@@ -228,7 +228,12 @@ def read(self, src):
         """Parse a datetime object from a string."""
         # parse timezone ourselves, since %z is not always supported
         # see: http://bugs.python.org/issue6641
-        string, tz = src.rsplit(' ', 1)
+        srcs = src.split(' ')
```
anthrotype: the variable name "srcs" is not very clear to me. Maybe something like "chunks" or "words" would make more sense?

Lib/glyphsLib/types.py
schriftgestalt: That was already like this. I just adapted it.

anthrotype: but the tests are failing now because of that
https://travis-ci.org/googlei18n/glyphsLib/jobs/315772812#L756

anthrotype:
implemented differently (also fixed the tests)
https://github.com/googlei18n/glyphsLib/commit/fa9cdc8264809b5d37b821aef150ca6d3e96b92d

This only changes the parser. @schriftgestalt let us know when Glyphs.app will default to not writing the extra +0000 so we can modify our own writer accordingly.

PR #301, opened 2017-12-18
[builder.names] set empty weight string if 'Regular'

anthrotype:
same as we do for the 'width'.

Fixes https://github.com/googlei18n/glyphsLib/issues/300

### DIFF
--- a/Lib/glyphsLib/builder/names.py
+++ b/Lib/glyphsLib/builder/names.py
@@ -26,7 +26,7 @@ def to_ufo_names(self, ufo, master, family_name):
 
     styleName = build_style_name(
         width if width != 'Regular' else '',
-        weight,
+        weight if weight != 'Regular' else '',
         custom,
         is_italic
     )


schriftgestalt:
I solution is to put all pieces in an OrderedSet and remove "Regular" if it has more than 1 elements. Because you normally what to keep the "Regular" if there are not other elements.

anthrotype:
yeah, but that is already handled by the build_style_name function

```python
def build_style_name(width='', weight='', custom='', is_italic=False):
    """Build style name from width, weight, and custom style strings
    and whether the style is italic.
    """

    return ' '.join(
        s for s in (custom, width, weight, 'Italic' if is_italic else '') if s
    ) or 'Regular'
```

anthrotype:
I could bake the logic of "treat the string 'Regular' as a special one but only for the width and weight attributes" inside the `build_style_name` function itself. 
However, I prefer to leave it to the caller to pass correct input (and that is what we were already doing for the width -- where Regular was special-cased by the caller - but not for the weight).
Garbage in, garbage out.

PR #330, opened 2018-03-23
Change all uses of logger.warn to logger.warning

belluzj:


Lib/glyphsLib/builder/kerning.py
```diff
@@ -37,15 +37,15 @@ def _to_ufo_kerning(self, ufo, kerning_data):
         if left_is_class:
             left = 'public.kern1.%s' % match.group(1)
             if left not in ufo.groups:
-                # self.logger.warn(warning_msg % left)
+                # self.logger.warning(warning_msg % left)
```
anthrotype: is there a reason why this is left in but commented-out?

belluzj: Only a bad one: that GlyphsUnitTestSans has a lot of those so they used to show up in tests. And also this PR is incomplete, I just saw that I forgot to hit "Save all" in my text editor...

PR #333, opened 2018-03-26
Generate the name of a master using family, width, weight, custom order

madig:
Fixes https://github.com/googlei18n/glyphsLib/issues/323.

### DIFF
--- a/Lib/glyphsLib/classes.py
+++ b/Lib/glyphsLib/classes.py
@@ -1347,7 +1347,7 @@ class GSFontMaster(GSBase):
             self.customParameters['Master Name'] = name
 
     def _joinName(self):
-        names = [self.weight, self.width, self.customName]
+        names = [self.width, self.weight, self.customName]
         names = [n for n in names if n]  # Remove None and empty string
         # Remove all occurences of 'Regular'
         while len(names) > 1 and "Regular" in names:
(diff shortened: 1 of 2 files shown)


anthrotype:
does the `_splitName` function also need to be adjusted accordingly now that you changed the order returned by `_joinName`?

also, what does Glyphs.app when generating UFO masters? Does it write the width before the weight?

madig:
Jany says that `_splitNames` does not expect order.

anthrotype:
ok thanks. So was change this unintentional, or has something else changed in Glyphs.app recently?

anthrotype:
> Jany says that _splitNames does not expect order.

perhaps for consistency with `_joinNames`, `_splitNames` should return the tuple as `(width, weight, custom)`

https://github.com/googlei18n/glyphsLib/pull/333/files#diff-2486ae3fe6d67c56b1c1b22402d779c1R1386

anthrotype:
can we merge this?

belluzj:
> So was this change unintentional, or has something else changed in Glyphs.app recently?

Unintentional, I was confused by the fact the elsewhere in the code, the weight is handled before the width, but didn't think of how the stylenames should look in the end.

> perhaps for consistency with _joinNames, _splitNames should return the tuple as (width, weight, custom)

Bof, I don't know, both those methods are supposed to be private so it does not really matter

> can we merge this?

Yes

PR #335, opened 2018-03-26
Make to target UFOs filenames unique if masters have the same style name

madig:
Found this when a colleague made a six master glyphs file, where Glyphs provides just five different weight names; he just made the two boldest ones "Bold". Before, we would  overwrite the first with the second.

### DIFF
--- a/Lib/glyphsLib/builder/sources.py
+++ b/Lib/glyphsLib/builder/sources.py
@@ -53,6 +53,15 @@ def _to_designspace_source(self, master, is_regular):
         source.filename = build_ufo_path(
             '', source.familyName, source.styleName)
 
+        # Make sure UFO filenames are unique, lest we overwrite masters that
+        # happen to have the same weight name.
+        n = "_"
+        while any(s != source and s.filename == source.filename
+                  for s in self._sources.values()):
+            source.filename = os.path.basename(
+                build_ufo_path('.', source.familyName, source.styleName + n))
+            n += "_"
+
     location = {}
     for axis_def in get_axis_definitions(self.font):
         location[axis_def.name] = axis_def.get_design_loc(master)
--- a/Lib/glyphsLib/builder/sources.py
+++ b/Lib/glyphsLib/builder/sources.py
@@ -15,6 +15,7 @@
 from __future__ import (print_function, division, absolute_import,
                         unicode_literals)
 
+import logging
 import os
 
 from glyphsLib.util import build_ufo_path
@@ -22,6 +23,9 @@ from glyphsLib.util import build_ufo_path
 from .masters import UFO_FILENAME_KEY
 from .axes import (get_axis_definitions, get_regular_master,
                    font_uses_new_axes, interp)
+
+
+logger = logging.getLogger(__name__)
 
 
 def to_designspace_sources(self):
@@ -61,6 +65,12 @@ def _to_designspace_source(self, master, is_regular):
             source.filename = os.path.basename(
                 build_ufo_path('.', source.familyName, source.styleName + n))
             n += "_"
+            logger.warn("The master with id {} has the same style name ({}) "
+                        "as another one. All masters should have distinctive "
+                        "(style) names. Use the 'Master name' custom parameter"
+                        " on a master to give it a unique name. Proceeding "
+                        "with an unchanged name, but appending '_' to the file"
+                        " name on disk.".format(master.id, source.styleName))
 
     location = {}
     for axis_def in get_axis_definitions(self.font):
--- a/Lib/glyphsLib/builder/sources.py
+++ b/Lib/glyphsLib/builder/sources.py
@@ -60,10 +60,10 @@ def _to_designspace_source(self, master, is_regular):
         # Make sure UFO filenames are unique, lest we overwrite masters that
         # happen to have the same weight name.
         n = "_"
-        while any(s != source and s.filename == source.filename
+        while any(s is not source and s.filename == source.filename
                   for s in self._sources.values()):
             source.filename = os.path.basename(
-                build_ufo_path('.', source.familyName, source.styleName + n))
+                build_ufo_path('', source.familyName, source.styleName + n))
             n += "_"
             logger.warn("The master with id {} has the same style name ({}) "
                         "as another one. All masters should have distinctive "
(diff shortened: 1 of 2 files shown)


anthrotype:
hm. Wouldn't it be better to assert the master names are indeed unique and let the user fix the sources?
Is there a valid use case for having two masters with the same name in Glyphs?
Would setting the "Master Name" custom parameter be the recommended workaround?

madig:
Not sure... But then glyphslib should at least blow up violently.

anthrotype:
> glyphslib should at least blow up violently.

yes, that's what personally i would expect it to do in this case

moyogo:
+1 for making "Master Name" the correct solution, and raise if it’s missing in those cases.

schriftgestalt:
The changes in handling the master name aims at solving that problem. Before, with the popup, it was easy to get the same master names. In Glyphs prior to 2.5, the user can just add anything in the custom name field.

madig:
Alright, then I'll change the code to blow up instead :)

belluzj:
sorry; misclicked

I was going to say that since it works fine in Glyphs to have several masters with the same name, maybe users will be unhappy to have their perfectly fine .glyphs files rejected by glyphsLib? In particular if someone is using glyphsLib to edit the file with something else than Glyphs, it might be because they don't have access to Glyphs to easily change the master names. So IMO it would be more user-friendly to still generate UFOs and print a warning about two of them having the same name.

anthrotype:
What happens if one exports a glyphs file with two masters having the same name to UFOs from within Glyphs.app? Does it create a unique filename, or it silently overwrites one with the other, or shows an error?

schriftgestalt:
> What happens if one exports a glyphs file with two masters having the same name to UFOs from within Glyphs.app? Does it create a unique filename, or it silently overwrites one with the other, or shows an error?

It overwrites without warning. I have a look.

anthrotype:
> It overwrites without warning

good, same as glyphsLib. Let's coordinate on the fix then. I think it'd be better to force the user to rename the masters with a "Master Name" custom parameter.
Btw, why does it appear in grey and is not listed in the master's custom parameters drop-down menu? It does change a master's name as shown on the left panel after applied, but I don't see it listed.

schriftgestalt:
The "Master Name" parameter is not directly supported any more. The master has a `.name` property and the parameter is only used as fallback if it was set in an older file.

anthrotype:
oh I didn't notice now a master's "Name" field can be edited from the GUI! :)

madig:
So, the blow up can stay?

Lib/glyphsLib/builder/sources.py
```diff
@@ -53,6 +53,16 @@ def _to_designspace_source(self, master, is_regular):
             # FIXME: (jany) have this function not write the dot
             build_ufo_path('.', source.familyName, source.styleName))
 
+        # Make sure UFO filenames are unique, lest we overwrite masters that
+        # happen to have the same weight name.
+        if any(s != source and s.filename == source.filename
```
anthrotype: what type are `s` and `source` here? do they implement `__eq__`? Maybe you meant `is not` for identity checking, rather than equality

anthrotype:
needs rebasing

Lib/glyphsLib/builder/sources.py
madig: `s` is the same type as `source`, both are in the same list. I'm filtering out the element in question while looking at all other elements in the list. Will try `is not` :)

madig:
This now breaks some tests :)

Lib/glyphsLib/builder/sources.py
```diff
@@ -53,6 +53,16 @@ def _to_designspace_source(self, master, is_regular):
...
+               for s in self._sources.values()):
+            raise ValueError(
+                "The source Glyphs file has duplicate master "
+                "names or styles. A master name must be unique. Below "
+                "Glyphs 2.5, you can add custom names in the custom name "
+                "field.")
```
anthrotype: the error message should at least tell which master names are duplicate.
Also I don't understand when it says "below Glyphs 2.5, you can add custom names in the custom name field."?
Do you mean the "Master Name" custom parameter? If so, this needs to be made clear.
And what about equal or above 2.5?

Again, I think you want to check `s is not source` to make it explicit that your are checking that objects are non identical, not equivalent. Even if the SourceDescriptor class doesn't implement `__eq__` (yet?) and thus falls back on `object.__eq__` which compares objects' id, explicit is better than implicit.

anthrotype:
Fair enough. Let's make it a ERROR logging message then, but continue to make the UFO masters with the unique name.

madig:
So I guess it's fine to log an error/warn but proceed to write out `#1` files?

anthrotype:
yes, makes sense that the GUI app forces you to make the edits, but the glyphsLib library still allows to generate UFOs in case one doesn't have the app available and thus cannot easily edit the glyphs file.

madig:
Commit Agreeable?

Lib/glyphsLib/builder/sources.py
```diff
@@ -53,6 +57,21 @@ def _to_designspace_source(self, master, is_regular):
         source.filename = build_ufo_path(
             '', source.familyName, source.styleName)
 
+        # Make sure UFO filenames are unique, lest we overwrite masters that
+        # happen to have the same weight name.
+        n = "_"
+        while any(s != source and s.filename == source.filename
```
anthrotype: please replace `s != source` with `s is not source`

Lib/glyphsLib/builder/sources.py
```diff
@@ -53,6 +57,21 @@ def _to_designspace_source(self, master, is_regular):
...
+                  for s in self._sources.values()):
+            source.filename = os.path.basename(
+                build_ufo_path('.', source.familyName, source.styleName + n))
```
anthrotype: why the "." here? A few lines before you called `build_ufo_path("", ...)` with an empty directory string.
You don't need to compute the filename again, just add the suffix to the current one `source.filename += n`

Lib/glyphsLib/builder/sources.py
anthrotype: I wonder if instead of a while loop called each time `for master in self.font.masters` (O(n^2)), we could store the set of filenames used so far in a private attribute of self and use that to check if each new master is given a unique filename..
But I guess the number of masters is unlikely to be huge.

madig: > please replace s != source with s is not source

Argh forgot about that.

> I wonder if instead of a while loop called each time

This would have to be benchmarked... I'd say leave for now.

Lib/glyphsLib/builder/sources.py
madig: Well, I wanted to compute `Something-Bold_.ufo` instead of `Something-Bold.ufo_`. I vaguely remember something choking when the suffix isn't ".ufo".

PR #338, opened 2018-03-27
Writer: string that looks like float needs quotes

### LINKED ISSUE #337: float written un-quoted, loading back the file raises ValueError
Take this file:
https://raw.githubusercontent.com/googlei18n/noto-source/master/src/NotoSansCanadianAboriginal/NotoSansCanadianAboriginal-MM.glyphs

Load it with glyphsLib, save it back to file with no modifications, and try to re-load the new copy.
A ValueError is raised while attempting to parse some unquoted float numbers in a userData dictionary, that in the original font were wrapped in quotes:

```python
>>> from glyphsLib import GSFont
>>> f1 = GSFont("src/NotoSansCanadianAboriginal/NotoSansCanadianAboriginal-MM.glyphs")
>>> f1.save("src/NotoSansCanadianAboriginal/NotoSansCanadianAboriginal-MM_2.glyphs")
>>> f2 = GSFont("src/NotoSansCanadianAboriginal/NotoSansCanadianAboriginal-MM_2.glyphs")
  File "/Users/clupo/Github/glyphsLib/Lib/glyphsLib/classes.py", line 3013, in __init__
    p.parse_into_object(self, fp.read())
  File "/Users/clupo/Github/glyphsLib/Lib/glyphsLib/parser.py", line 65, in parse_into_object
    i = self._parse_dict_into_object(res, text, 1)
  File "/Users/clupo/Github/glyphsLib/Lib/glyphsLib/parser.py", line 175, in _parse_dict_into_object
    result = self._parse(text, i)
  File "/Users/clupo/Github/glyphsLib/Lib/glyphsLib/parser.py", line 103, in _parse
    return self._parse_list(text, i)
  File "/Users/clupo/Github/glyphsLib/Lib/glyphsLib/parser.py", line 203, in _parse_list
    list_item, i = self._parse(text, i)
  File "/Users/clupo/Github/glyphsLib/Lib/glyphsLib/parser.py", line 97, in _parse
    return self._parse_dict(text, i)
  File "/Users/clupo/Github/glyphsLib/Lib/glyphsLib/parser.py", line 156, in _parse_dict
    i = self._parse_dict_into_object(res, text, i)
  File "/Users/clupo/Github/glyphsLib/Lib/glyphsLib/parser.py", line 175, in _parse_dict_into_object
    result = self._parse(text, i)
  File "/Users/clupo/Github/glyphsLib/Lib/glyphsLib/parser.py", line 97, in _parse
    return self._parse_dict(text, i)
  File "/Users/clupo/Github/glyphsLib/Lib/glyphsLib/parser.py", line 156, in _parse_dict
    i = self._parse_dict_into_object(res, text, i)
  File "/Users/clupo/Github/glyphsLib/Lib/glyphsLib/parser.py", line 175, in _parse_dict_into_object
    result = self._parse(text, i)
  File "/Users/clupo/Github/glyphsLib/Lib/glyphsLib/parser.py", line 132, in _parse
    value = self.current_type(value)
ValueError: invalid literal for int() with base 10: '22.0'
```

diffing the original font with the dumped font, reveals:

```diff
@@ -893,7 +893,7 @@
 GSOffsetVertical = 3;
 noodleExtremesAndInflections = 1;
 noodleRemoveOverlap = 1;
-noodleThickness = "22.0";
+noodleThickness = 22.0;
 };
 verticalStems = (
 26,
@@ -981,7 +981,7 @@
 userData = {
 noodleExtremesAndInflections = 1;
 noodleRemoveOverlap = 0;
-noodleThickness = "190.0";
+noodleThickness = 190.0;
 };
 verticalStems = (
 78,
@@ -1068,7 +1068,7 @@
 userData = {
 noodleExtremesAndInflections = 1;
 noodleRemoveOverlap = 1;
-noodleThickness = "84.0";
+noodleThickness = 84.0;
 };
 verticalStems = (
 146,
@@ -1157,7 +1157,7 @@
 GSOffsetVertical = 1;
 noodleExtremesAndInflections = 1;
 noodleRemoveOverlap = 0;
-noodleThickness = "106.0";
+noodleThickness = 106.0;
 };
 verticalStems = (
 193,
```

madig:
Fixes #337.

### DIFF
--- a/Lib/glyphsLib/parser.py
+++ b/Lib/glyphsLib/parser.py
@@ -76,11 +76,8 @@ class Parser(object):
             return unicode
         if parsed[-1] != '"':
             try:
-                float_val = float(value)
-                if float_val.is_integer():
-                    current_type = int
-                else:
-                    current_type = float
+                v = float(value)
+                current_type = lambda _: v if not v.is_integer() else int(v)
             except:
                 current_type = unicode
         else:
@@ -128,12 +125,6 @@ class Parser(object):
             if self.current_type == bool:
                 value = bool(int(value))  # bool(u'0') returns True
                 return value, i
-
-            if self.current_type == int:
-                if "." in value:
-                    decimals, fractionals = value.split(".")
-                    if all(f == "0" for f in fractionals):
-                        value = decimals
 
             value = self.current_type(value)
 
--- a/Lib/glyphsLib/parser.py
+++ b/Lib/glyphsLib/parser.py
@@ -129,6 +129,12 @@ class Parser(object):
                 value = bool(int(value))  # bool(u'0') returns True
                 return value, i
 
+            if self.current_type == int:
+                if "." in value:
+                    decimals, fractionals = value.split(".")
+                    if all(f == "0" for f in fractionals):
+                        value = decimals
+
             value = self.current_type(value)
 
             return value, i
--- a/Lib/glyphsLib/writer.py
+++ b/Lib/glyphsLib/writer.py
@@ -209,6 +209,13 @@ def _needs_quotes(string):
     try:
         int(string)
     except ValueError:
+        pass
+    else:
+        return True
+
+    try:
+        float(string)
+    except ValueError:
         return False
     else:
         return True
--- a/tests/parser_test.py
+++ b/tests/parser_test.py
@@ -130,6 +130,12 @@ class ParserTest(unittest.TestCase):
             [('key', b'value')]
         )
 
+    def test_parse_stringy_floats(self):
+        self.run_test(
+            b'{noodleThickness = "106.0";}',
+            [('noodleThickness', '106.0')]
+        )
+
 
 class ParserGlyphTest(unittest.TestCase):
     def test_parse_empty_glyphs(self):
--- a/tests/parser_test.py
+++ b/tests/parser_test.py
@@ -136,6 +136,18 @@ class ParserTest(unittest.TestCase):
             [('noodleThickness', '106.0')]
         )
 
+    def test_parse_float_no_frac_as_int(self):
+        self.run_test(
+            b'{noodleThickness = 106.0;}',
+            [('noodleThickness', 106)]
+        )
+
+    def test_parse_float_as_float(self):
+        self.run_test(
+            b'{noodleThickness = 106.1;}',
+            [('noodleThickness', 106.1)]
+        )
+
 
 class ParserGlyphTest(unittest.TestCase):
     def test_parse_empty_glyphs(self):
--- a/tests/writer_test.py
+++ b/tests/writer_test.py
@@ -123,6 +123,7 @@ class WriterTest(unittest.TestCase, test_helpers.AssertLinesEqual):
             'a': 'test',
             'b': [1, {'c': 2}],
             'd': [1, "1"],
+            'noodleThickness': "106.0",
         }
         # grid -> gridLength
         font.grid = 35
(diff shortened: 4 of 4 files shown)


anthrotype:
i'm confused. Is it invalid to write the float without the quotes, or is it valid and the parser is wrong in interpreting the float as an integer:
see https://github.com/googlei18n/glyphsLib/issues/337#issuecomment-376568878

anthrotype:
or even worse, is that thing `noodleThickness = "22.0"` really a _string_, and thus should be kept with the quotes to disambiguate it from a real _float_?

(this is one of the situations where a file format that's clear about the types like json would make a difference)

madig:
Errr... @schriftgestalt?

anthrotype:
it seems like it's being interpreted as a string

```python
>>> type(f.masters[0].userData['noodleThickness'])
str
```

madig:
Jany and Denis say: It's really a string that happens to look like a float. Changed title 😁

anthrotype:
ok, this PR may fix the string-that-looks-like-a-float case, but doesn't explain why I also get the same ValueError if I remove the quotes and write the float as a float, like so:

```diff
diff --git a/tests/data/GlyphsUnitTestSans.glyphs b/tests/data/GlyphsUnitTestSans.glyphs
index 529c601..e89c18f 100644
--- a/tests/data/GlyphsUnitTestSans.glyphs
+++ b/tests/data/GlyphsUnitTestSans.glyphs
@@ -117,7 +117,7 @@ GSRoughenSegmentLength = 15;
 GSRoughenVertical = 10;
 noodleExtremesAndInflections = 1;
 noodleRemoveOverlap = 0;
-noodleThickness = "106.0";
+noodleThickness = 106.0;
 };
 verticalStems = (
 17,
```

this breaks both the parser_main and the rountrip tests

anthrotype:
is that a contrived case, and is it technically impossible (as in illegal) to even write an unquoted float in this ascii plist format?
in which case, once would never be able to disambiguate a real float from a quoted-string-that-looks-like-a-float...

anthrotype:
yeah, I keep forgetting that

> numbers are handled as strings in property lists

or rather, in "old style" property lists:
https://developer.apple.com/library/content/documentation/Cocoa/Conceptual/PropertyLists/OldStylePlists/OldStylePLists.html

anthrotype:
> The quotation marks can be omitted if the string is composed strictly of alphanumeric characters and contains no white space

the "." in the float representation is non-alphanumeric? I would think so. In which case, decimal floats would always need to be written out as quoted strings.

I see in the writer module there's a `NSPropertyListNameSet` array that has at index 46 (the result of `ord(".")`) the value `True`; the writer's `_needs_quotes` function returns True when a character ordinal is _not_ in that set (has value `False`).

I wonder if that should be flipped... 🤔

anthrotype:
I tried from within Glyphs.app to set `Glyphs.font.masters[0].userData["noodleThickness"] = 106.1` i.e. to a python float, saved the file and I can see that the value does _not_ have quotes.

So, whatever the old-style ascii plist says, Glyphs.app is using the quotes to distinguish between strings (quoted sometimes, e.g. when they may be confused for an integer or float numbers) and numbers (never quoted, including floats with their non-alphanumeric "." in them).

Our writer was "wrong" in that it was writing a string that looked like a float number ("106.0") without quotes, and this PR addresses that.

The parser still can't parse a float-that-is-actually-an-integer, (e.g. `106.0` without quotes), but that never seems to happen in that, both glyphs and glyphsLib always write these as integers without the redundant decimal part.

let's merge this and move on

schriftgestalt:
I explained it in the other thread.

madig:
I thoroughly confused but does this PR now make your roundtrip diff-less?

PR #339, opened 2018-03-29
allow to apply_instance_data to selected instances

anthrotype:
as discussed yesterday

PTAL

### DIFF
--- a/Lib/glyphsLib/builder/instances.py
+++ b/Lib/glyphsLib/builder/instances.py
@@ -28,6 +28,8 @@ from .axes import (get_axis_definitions, is_instance_active, interp,
                    WEIGHT_AXIS_DEF, WIDTH_AXIS_DEF)
 from .custom_params import to_ufo_custom_params
 
+import defcon
+
 EXPORT_KEY = GLYPHS_PREFIX + 'export'
 WIDTH_KEY = GLYPHS_PREFIX + 'width'
 WEIGHT_KEY = GLYPHS_PREFIX + 'weight'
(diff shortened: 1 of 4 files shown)


Lib/glyphsLib/builder/instances.py
```diff
@@ -308,30 +310,39 @@ def set_width_class(ufo, designspace, instance):
...
-
-        glyphs_instance = InstanceDescriptorAsGSInstance(instance)
-        # to_ufo_custom_params(self, ufo, data.parent)  # FIXME: (jany) needed?
+    for designspace_instance in designspace.instances:
+        fname = designspace_instance.filename
+        assert fname is not None, ("instance %r missing required filename" %
+                getattr(designspace_instance, "name", designspace_instance))
+        if include_filenames is not None:
+            fname = os.path.normcase(os.path.normpath(fname))
+            if fname not in include_filenames:
```
belluzj: Should you process `include_filenames` with `normpath` and `normcase` as well?

tests/builder/instances_test.py
```diff
@@ -0,0 +1,71 @@
...
+    if instance_names is None:
+        # generate all instances
+        builder.process()
+        include_filenames = None
+    else:
+        # generate only selected instances
+        for name in instance_names:
+            builder.readInstance(("stylename", name))
+        # make relative filenames from paths returned by MutatorMath
+        include_filenames = {os.path.normcase(os.path.normpath(
```
belluzj: I think that if you want the callers of `apply_instance_data` to normpath/normcase themselves, you should document that (or have apply_intance_data take care of it)

anthrotype:
ok, maybe it's better we do the normalisation of the  `include_filenames` argument in here, instead of requiring the caller to do it

Lib/glyphsLib/builder/instances.py
anthrotype: > Should you process include_filenames with normpath and normcase as well?

done

belluzj:
One last comment: is it ok to break compatibility with older fontmake like this? After this is merged, some people who might want the new version of glyphsLib will have to update fontmake as well?

anthrotype:
hm, I see what you mean. Maybe we could check if the first positional argument of apply_instance_data is an instance of `InstanceData`, and take the `designspace` from there

anthrotype:
Ok, I made the changes backward compatible to make it work with current fontmake (without https://github.com/googlei18n/fontmake/pull/410)

PR #340, opened 2018-03-31
mute "layer without name" warning if minimize_glyphs_diff=False

anthrotype:
it can be very chatty, especially on some fonts which have tons of these
https://github.com/googlei18n/noto-source/issues/114

How often does this situation happen? It's unlikely that such name-less layers are one of the master-linked layers -- which are the ones that fontmake cares about for building fonts.


schriftgestalt:
Master layers should not have name on their own (sometimes they have, that is when there where made a master manually). None master layers can have an empty name. Then they can be safely ignored.

Issue #341, opened 2018-04-02
designspace axis map must not contain same keys mapping to different values


anthrotype:
it's called "map" because it works like a mapping of keys (user locations) to values (design space locations). Just like in python dicts, one key cannot map to multiple values.

this is how varLib is translating from the "input" and "output" attributes of map elements to a python dictionary:
https://github.com/fonttools/fonttools/blob/a523a697164a4cdc721a06dec80791d4b1496f12/Lib/fontTools/varLib/__init__.py#L626

So if more than one "input" with same value is present, the last one wins and overwrites the previous ones.

Now, because of the way we map Thin, ExtraLight and UltraLight all to 250 weight class (see classes.py WEIGHT_CODES), we end up writing things like this (this is GlyphsUnitTestSans.glyphs, exported to designspace):

```xml
<?xml version='1.0' encoding='utf-8'?>
<designspace format="3">
    <axes>
        <axis default="400" maximum="900" minimum="250" name="Weight" tag="wght">
            <labelname xml:lang="en">Weight</labelname>
            <map input="250" output="17" />
            <map input="250" output="30" />
            <map input="300" output="55" />
            <map input="400" output="75" />
            <map input="400" output="90" />
```

In this font, there are two Thin and Extra Light instances that are assigned a 250 weight class. And there is even a "Web" instance that has default 400 weight class but has a different design location (75) than the Regular instance. In the latter case, one would expect the user to add a weightClass custom parameter (e.g. 350, in between Light and Regular) to make this unique and avoid the clash.
 
Note that this test font does not use the new Axes and Axis Location custom parameters, therefore is using the user location of the instances to make the axis map elements.

Even if the Axis Location was used we should raise an error if two masters have the same Axis Location but different design space locations, to prevent such invalid map elements.

anthrotype:
there may be good legacy reasons (e.g. support some old MS Word) to have OS/2 weight class set to 250 for thin and extra light styles, but for the designspace axis maps -- which are primarily meant to be used to build a non-linear normalization mapping in the avar table -- we don't want that to happen.
The user location of a Thin named instance in a variable font is 100, not 250.

I just compared opening GlyphsUnitTestSans.glyphs in latest stable Glyphs.app (2.4.4 1075) and in latest beta (2.5b 1126) and it appears that in the latest beta, the default weight class for Thin and Extra Light is respectively 100 and 200, instead of the legacy 250 value.

[...]

PR #343, opened 2018-04-02
Fixes for designspace axes, and for UFO instances weight/width class

anthrotype:
- don't write axis map elements if all input == output 
- don't write conflicting mappings (the same input maps to different outputs); if these occur (either because "Axis Location" parameters point different masters to same user-space location, or different instances have the same user-space location), then use the last occurrence but also warn the user about the mapping been redefined (would it be better to raise? see #341)
- Drop the legacy `250` value for default Thin, ExtraLight and UltraLight OS/2 weight classes. The latest Glyphs.app beta uses 100 and 200, so we do the same here.
- when setting OS/2 weight/width of UFO instances in apply_instance_data, look up weight and width axes by their tag rather than the name (which may not always be "Weight" or "Width")

### DIFF
--- a/Lib/glyphsLib/builder/instances.py
+++ b/Lib/glyphsLib/builder/instances.py
@@ -293,18 +293,17 @@ def _set_class_from_instance(ufo, designspace, instance, axis_tag):
     try:
         design_loc = instance.location[axis_def.name]
     except KeyError:
-        # The location does not have this axis. Use default value
-        design_loc = axis_def.default_user_loc
-
-    if mapping:
-        # Retrieve the user location (weightClass/widthClass)
-        # by going through the axis mapping in reverse.
-        reverse_mapping = sorted({dl: ul for ul, dl in mapping}.items())
-        user_loc = interp(reverse_mapping, design_loc)
-        axis_def.set_ufo_user_loc(ufo, user_loc)
+        user_loc = axis_def.default_user_loc
     else:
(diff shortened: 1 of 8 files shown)


Lib/glyphsLib/builder/instances.py
```diff
@@ -274,44 +274,51 @@ def __init__(self, descriptor):
                 self.customParameters.append(GSCustomParameter(name, value))
 
 
-def _set_class_from_instance(ufo, designspace, instance, axis_def):
+def _set_class_from_instance(ufo, designspace, instance, axis_tag):
     # FIXME: (jany) copy-pasted from above, factor into method?
-    design_loc = None
+    assert axis_tag in ("wght", "wdth")
+
+    factory = AxisDefinitionFactory()
```
belluzj: This factory is meant to be used by calling `get` exactly once per axis in the same order as in the designspace/font, so that the Glyphs properties `weightValue`, `widthValue`, `customValue`, etc are "aligned" with the axes. It may not matter here but still, I would rather have the factory stay hidden in axes.py and provide a method like `get_designspace_axis_defs` for use outside of axes.py

What do you think?

anthrotype: sure I can do that

anthrotype:
@belluzj i'm going to merge this as is for now. feel free to do any clean ups on top of it

PR #345, opened 2018-04-02
Omit default labelnames

### LINKED ISSUE #342: labelname only needed for non-registered axes, or when different from "name"
for weight, width, etc. it's not necessary to write the axis labelname element, as varLib will use the default registered name when writing the fvar.

the lablename is only needed if:

1) one wants to name one of the registered axes with a different labelname from the standard "en" (english);
2) one wants to have a "labelname" for an axis (be it registered or custom) that's different from the latter "name" (used for internally referencing axes in location elements). VarLib uses the axis "name" attribute when no labelname elements are present. 

See varLib code:
https://github.com/fonttools/fonttools/blob/a523a697164a4cdc721a06dec80791d4b1496f12/Lib/fontTools/varLib/__init__.py#L628-L638

anthrotype:
Fixes #342

### DIFF
--- a/Lib/glyphsLib/builder/axes.py
+++ b/Lib/glyphsLib/builder/axes.py
@@ -140,11 +140,19 @@ def to_designspace_axes(self):
     regular_master = get_regular_master(self.font)
     assert isinstance(regular_master, classes.GSFontMaster)
 
+    # TODO also include the other registered axes?
+    standard_axes_names = {
+        WEIGHT_AXIS_DEF.tag: WEIGHT_AXIS_DEF.name,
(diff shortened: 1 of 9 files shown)


anthrotype:
actually, we could completely stop writing English "labelname" elements for all the axes, not just the weight and width, and simply use the "name" attribute.
If "labelname" is not present in an axis, varLib will use the axis name attribute anyway.
And with the Glyphs.app "Axes" custom parameter one can only specifcy a single (presumably English) Name for a given axis tag, which we already translate into the axis "name" attribute in the designspace.

BTW, support for localized labelnames are still not enabled in varLib (but that's a separate issue https://github.com/fonttools/fonttools/issues/930#issuecomment-329979235).

PR #346, opened 2018-04-05
Add tests for disabled and automatic features, fix roundtripping them

madig:
The code was trying to subscript an `ast.Comment` object before. Get the `.text` property instead for subscription.

When are we going to use type hints + mypy to catch these kinds of errors sooner? 😁


madig:
Don't merge yet, need to investigate something.

madig:
Ok now.

PR #352, opened 2018-04-19
Handle hex binary data

madig:
As encountered in a Glyphs 2.5 file.

### DIFF
--- a/Lib/glyphsLib/parser.py
+++ b/Lib/glyphsLib/parser.py
@@ -22,6 +22,7 @@ import re
 import logging
 import sys
 import base64
+import binascii
 
 import glyphsLib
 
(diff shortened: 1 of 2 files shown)


anthrotype:
did you check this?
https://github.com/googlei18n/glyphsLib/issues/119

madig:
Nope.

anthrotype:
Basically we need to figure out from which Glyphs.app version the serialisation of the binary data changed from one format to the other and decode accordingly

schriftgestalt:
Can you check how many file even have binary data? I suspect it to be none or only very few.

madig:
One turned up the other day, it may have come from on of our FontLab scripts.

schriftgestalt:
The change was in the 2.5 branch around build 995, but the build number is not really helping here as there where similar numbers in the 2.4 branch ;(. I don’t have a list of all 2.5 versions and their numbers.

madig:
Hm. Merge PR as is then? Maybe not worth the effort to list all version numbers that may be affected?

anthrotype:
maybe we should only support binary data encoded as hexadecimal string, and remove support for base64 encoded strings, now that Glyphs.app only writes it as hexadecimal.

madig:
I just tried to access the binary data we found in one file, hex format. Neither 939 nor 1103 could access the data, both just ignore it. I also couldn't make any binary data from the macro window -- what would produce binary? 

So... Let's drop base64?

anthrotype:
> Let's drop base64?

yeah. We never released a glyphsLib version which included support for it anyway.

anthrotype:
I'm gonna remove base64 support

PR #354, opened 2018-04-25
Drop base64; always use hexadecimal strings for binary data

anthrotype:
The parser will no longer be able to parse and decode base64 encoded strings in .glyphs files.
And the builder will always treat the userData as hexadecimal strings when decoding to/encoding from UFO data files.

See discussion in #352 

Related issue #119  

I think if DaMa has UFO files floating around with base64 data, they can easily write a simple script to re-encode them as hexadecimal strings.

### DIFF
--- a/Lib/glyphsLib/builder/user_data.py
+++ b/Lib/glyphsLib/builder/user_data.py
@@ -110,10 +110,11 @@ def to_glyphs_master_user_data(self, ufo, master):
 
     # Save UFO data files
     if ufo.data.fileNames:
+        from glyphsLib.types import BinaryData
         ufo_data = {}
         for os_filename in ufo.data.fileNames:
             filename = posixpath.join(*os_filename.split(os.path.sep))
-            ufo_data[filename] = bytearray(ufo.data[os_filename])
+            ufo_data[filename] = BinaryData(ufo.data[os_filename])
         master.userData[UFO_DATA_KEY] = ufo_data
 
 
--- a/Lib/glyphsLib/builder/user_data.py
+++ b/Lib/glyphsLib/builder/user_data.py
@@ -15,7 +15,6 @@
 from __future__ import (print_function, division, absolute_import,
                         unicode_literals)
 
-import binascii
 import os
 import posixpath
 
@@ -51,7 +50,7 @@ def to_ufo_master_user_data(self, ufo, master):
     if UFO_DATA_KEY in master.userData:
         for filename, data in master.userData[UFO_DATA_KEY].items():
             os_filename = os.path.join(*filename.split('/'))
-            ufo.data[os_filename] = binascii.unhexlify(data)
+            ufo.data[os_filename] = bytes(data)
 
 
 def to_ufo_glyph_user_data(self, ufo, glyph):
@@ -114,10 +113,7 @@ def to_glyphs_master_user_data(self, ufo, master):
         ufo_data = {}
         for os_filename in ufo.data.fileNames:
             filename = posixpath.join(*os_filename.split(os.path.sep))
-            data_bytes = binascii.hexlify(ufo.data[os_filename])
-            # FIXME: (jany) The `decode` is here because putting bytes in
-            # userData doesn't work in Python 3. (comes out as `"b'stuff'"`)
-            ufo_data[filename] = data_bytes.decode()
+            ufo_data[filename] = bytearray(ufo.data[os_filename])
         master.userData[UFO_DATA_KEY] = ufo_data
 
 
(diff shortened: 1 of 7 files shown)


tests/data/GlyphsUnitTestSans.glyphs
```diff
@@ -115,6 +115,9 @@ GSOffsetVertical = 9;
 GSRoughenHorizontal = 15;
 GSRoughenSegmentLength = 15;
 GSRoughenVertical = 10;
+com.schriftgestaltung.Glyphs.ufoData = {
```
anthrotype: can this occur in real life?

Lib/glyphsLib/parser.py
```diff
@@ -135,17 +134,9 @@ def _parse(self, text, i, _parsing_unicodes=False):
         m = self.hex_re.match(text, i)
         if m:
             parsed, value = m.group(0), m.group(1)
-            decoded = binascii.unhexlify(value)
+            decoded = bytearray(binascii.unhexlify(value))
```
anthrotype: I thought about returning something else than mere `bytes`, e.g. a bytearray object, so that in the Writer we can detect that, and serialize it as a unquoted, hexadecimal string within `<` and `>` (see below). If we use generic bytes, then it's tricky to distinguish between python2 `str` values containing ascii text, which the user may have set in a userData dict, and actual binary data which is meant to be hexlified. 
I know, bytearrays are mutable. If that's a problem, we may wish to subclass the built-in `bytes`. But bytearray is handy because it's built-in. And it's unequivocally binary data.

anthrotype:
FYI, I'll be away until Wednesday next week so I won't be able to finish this. Feel free to pick it up where I left.

Lib/glyphsLib/builder/user_data.py
```diff
@@ -114,10 +113,7 @@ def to_glyphs_master_user_data(self, ufo, master):
         ufo_data = {}
         for os_filename in ufo.data.fileNames:
             filename = posixpath.join(*os_filename.split(os.path.sep))
-            data_bytes = base64.b64encode(ufo.data[os_filename])
-            # FIXME: (jany) The `decode` is here because putting bytes in
-            # userData doesn't work in Python 3. (comes out as `"b'stuff'"`)
```
anthrotype: > FIXME: (jany) The `decode` is here because putting bytes in userData doesn't work in Python 3. (comes out as `"b'stuff'"`)

the reason of course was that the current Writer did not explicitly handle binary data as hex strings (unlike the Parser which Nikolaus and Jany had extended to support parsing these). It was simply calling `unicode(value)` and then escaping the quotes.

anthrotype:
I'm no longer sure about using `bytearray` here.

When I open a .glyphs file containing such binary user data in Glyphs.app, and I print the type of the loaded object from the Macro Panel, I get `_NSInlineData`, a subclass of `NSData` (as wrapped by pyobjc). This is an immutable byte buffer, more or less equivalent to python3 `bytes` object.

I'm thinking that maybe we could be strict with `str` vs `bytes` in glyphsLib parser and writer, and:

- always return (unicode) `str` from the parser (which I think we already do, no?)
- always interpret `bytes` objects as binary data (that must be serialized as hexadecimal strings) in the writer

If some python2 user sets some `str` bytes strings instead of using the correct `unicode` type, we could ensure they are decoded in the classes' setters.

anthrotype:
It would be interesting to know where in the glyphs file format this hex-encoded binary data can happen. I mean, what fields expect it or allow it. Is it only in the userData values? If that's the case, then we could enforce the use of unicode strings for the userData values that are meant to contain text, and let any `bytes` values be encoded as hexadecimal strings; whereas, in other parts of the glyphs file, we could be more relaxed and allow ascii bytes str be implicity coerced to unicode, as the writer currently does.

anthrotype:
I ended up defining a new `BinaryData` class in `glyphsLib.types` that subclasses `bytes`, and adds two additional methods: `fromHex` (classmethod constructor, used by the parser) and `plistValue` used by the writer.

@belluzj PTAL

tests/data/GlyphsUnitTestSans.glyphs
belluzj: We've seen binary data in our Glyphs files that was a remain of old Robofont plugin data... I don't think the data in there was worth anything or taken into account by our build tools, but it did appear and caused a crash of glyphsLib for one or two projects, that's why we want glyphsLib to handle it. (was that your question?)

anthrotype: kind of, yeah. Ideally I would like to know all the places where this binary data can (or is likely to) happen in a glyphs file. I guess only @schriftgestalt can tell for certain.

schriftgestalt: Binary data can occur in userData and in custom parameters (if some script or plugin puts it there). In custom parameters it is not officially supported but it uses the same code path as the storage of user data.

PR #358, opened 2018-05-07
Put UFO instances in instance_ufos/ subdirectory by default

madig:
[TestFont.zip](https://github.com/googlei18n/glyphsLib/files/1980923/TestFont.zip)
```
>>> import glyphsLib
>>> a=glyphsLib.load_to_ufos("TestFont.glyphs")
>>> a=glyphsLib.loads("TestFont.glyphs")
>>> d=glyphsLib.to_designspace(a)
>>> [x.filename for x in d.instances]
['instance_ufos/newFont-Regular.ufo', 'instance_ufos/newFont-Medium.ufo', 'instance_ufos/newFont-Bold.ufo']
>>> [x.filename for x in d.sources]
['newFont-Regular.ufo', 'newFont-Bold.ufo']
```

### DIFF
--- a/Lib/glyphsLib/builder/instances.py
+++ b/Lib/glyphsLib/builder/instances.py
@@ -83,7 +83,7 @@ def _to_designspace_instance(self, instance):
             fname = self.instance_dir + '/' + os.path.basename(fname)
         ufo_instance.filename = fname
     if not ufo_instance.filename:
-        instance_dir = self.instance_dir or ''
+        instance_dir = self.instance_dir or 'instance_ufos'
         ufo_instance.filename = build_ufo_path(
             instance_dir, ufo_instance.familyName, ufo_instance.styleName)
 


anthrotype:
hm... so you're assuming fontmake's current defaults (which we are basically trying to get rid of).

anthrotype:
note that the filename attributes are always relative to the directory of the designspace itself. Fontmake by default places the designspace in the same directory as the master_ufo, so the instance_ufo folder is actually a sibling `../instance_ufo`

anthrotype:
anyway, I'm not sure this goes in the right direction. I'll have to think about it. Please give more info about the issue this is trying to solve, thanks.

madig:
Well, the idea is to prevent instances overwriting masters if they have the same name. Where would this be better kept? Parameter in `to_designspace` or something?

anthrotype:
this is related somewhat https://github.com/googlei18n/glyphsLib/issues/325

PR #359, opened 2018-05-10
Accept os.PathLike objects in GSFont.__init__

madig:
So you can pass in `Path` and friends.

### DIFF
--- a/Lib/glyphsLib/classes.py
+++ b/Lib/glyphsLib/classes.py
@@ -3003,6 +3003,11 @@ class GSFont(GSBase):
         self._userData = None
 
         if path:
+            # Support os.PathLike objects. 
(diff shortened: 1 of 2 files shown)


anthrotype:
got a test?
of course, if should `@pytest.mark.skipif(sys.version_info < (3,4)`

madig:
Once we drop Python < 3.6, a better solution might be to copy the approach of the 3.6 stdlib: https://github.com/python/cpython/blob/a6a4dc816d68df04a7d592e0b6af8c7ecc4d4344/Lib/bz2.py#L91

madig:
Anything missing?

anthrotype:
I already approved, you can do the merge, no?

madig:
I see no merge button :O

anthrotype:
> I see no merge button :O

@madig sorry, I now added you as collaborator

PR #361, opened 2018-05-12
glyphs2ufo: focus on roundtripping to designspace plus master UFOs

madig:
* Add options to preserve metadata and reduce roundtrip differences when going to UFOs.
* Always generate a .designspace file.
* Remove option to interpolate instances. Use fontmake instead.

This is meant to be a base for discussing the direction. I want to add a ufo2glyphs command also, which can take either a bunch of UFOs or a designspace file.

I'm thinking of adding a mark color clamper of some sorts. We often get annoying and unprovoked markColor diffs due to rounding errors or something... So I'd clamp all colors to the nearest Glyphs color mark UI value. That should be enough color for anyone unless someone goes and modifies the color through a script.


schriftgestalt:
> Remove option to interpolate instances. Use fontmake instead.

That will implementing brace layers more difficult, isn't it?

madig:
What do you mean? I intend to change only the glyphs2ufo command-line script and make its primary use roundtripping between Glyphs and UFOs. I don't want to change the library itself.

Lib/glyphsLib/__init__.py
```diff
@@ -72,35 +78,55 @@ def build_masters(filename, master_dir, designspace_instance_dir=None,
...
-    else:
-        return ufos
+        if normalize_ufos:
+            try:
+                import ufonormalizer
+                ufonormalizer.normalizeUFO(ufo_path, writeModTimes=False)
+            except ImportError:
+                logger.warning("ufonormalizer not found, UFOs may contain "
+                               "more differences than necessary.")
+            except TypeError:
```
anthrotype: I understand the intention to catch the unsupported keyword argument here, but a TypeError might be raised from within `normalizeUFO` and that would be trigger the warning message and the exception would be passed over, which is not good.
I think it's ok to let the outdated ufonormalizer raise the error and stop here. The user should be able to upgrade it and then retry. And normalization is optional anyway, one can always not pass the option. Also the import error is obvious, just do the import and let it fail if not present. We have been asked to normalize and there is no ufonormalizer. The user should be able to take the necessary actions.

Lib/glyphsLib/__init__.py
```diff
@@ -72,35 +78,55 @@ def build_masters(filename, master_dir, designspace_instance_dir=None,
...
+                               "contain more differences than necessary.")
+
+    designspace_path = os.path.join(master_dir, designspace.filename)
+    designspace.write(designspace_path)
+
+    # All the instance data should be in the designspace. That's why for
+    # now we return the full designspace in place of `instance_data`.
+    # However, other functions still expect the instance data to have
+    # a list of (path, something) tuples, hence the class `InstanceData`.
+    return ufos, designspace_path, InstanceData(designspace)
```
anthrotype: note that fontmake (master) no longer uses this deprecated InstanceData list-of-tuples. You can drop it, unless it is being used by other internal glyphsLib functions.

Lib/glyphsLib/__main__.py
```diff
@@ -20,42 +20,63 @@
...
+def main(args=None):
+    """Converts a Glyphs.app source file into UFO masters or UFO instances and
+    MutatorMath designspace."""
+    parser = argparse.ArgumentParser(description=main.__doc__)
+    parser.add_argument(
+        "--version",
+        action="version",
+        version="glyphsLib %s" % (glyphsLib.__version__))
+    parser.add_argument(
+        "glyphs_file",
```
anthrotype: so you dropped the -g. Okay.

anthrotype:
@madig maybe also take a look at fontmake's `build_master_ufos`, which is very similar to this `build_masters` and see if there's anything to be added
https://github.com/googlei18n/fontmake/blob/da849fb44070459ddb5816582c4e649f151f1061/Lib/fontmake/font_project.py#L76-L113

madig:
Disabled instances are still written to disk and built. Would it make sense to teach glyphsLib/ufo2ft/designspaceLib/fontmake to read the export flag and delete it from the instances? And if so, where should that flag be handled?

madig:
Should we remove glyphsLib.build_instances() with associated functions and direct people to fontmake? It seems to be used only by the current glyphs2ufo.

anthrotype:
yes, and then we can drop dependency on mutatormath here in glyphsLib at least

belluzj:
> yes, and then we can drop dependency on mutatormath here in glyphsLib at least

What about smart components? At some point you were going to work on them, right? Do you think they should be handled outside of glyphsLib/not using mutatorMath?

madig:
Aren't smart components constructed at display/compile-time? We just shuffle the underlying metadata around here, no? Unless we want to decompose/bake those components into the UFO.

madig:
Woops. Jany pointed out that not only is the test is now broken (misuse DesignSpaceDocument.__init__, MutatorMath actually wrote the instances to disk), it also doesn't test that something is changed in the instances. What's more, apply_instance_data doesn't complain about the missing UFOs? Need to stare at the code harder.

madig:
I don't think there's anything that should be ported to fontmake, as we basically just post-process the UFOs to make round-tripping to Glyphs less annoying? Unless fontmake wants to make use of Glyphs metadata.

anthrotype:
> Unless fontmake wants to make use of Glyphs metadata

well, only in so far as it is relevant for compilation step

anthrotype:
if this glyphsLib.build_masters does all the things that fontmake build_master_ufos does (plus additional optional stuff to ensure roundtripping in version control), then I'd gladly use this in fontmake and remove the duplication

madig:
Then I'd need to rewrite the function to not save to disk so you can keep control, no?

madig:
This breaks because glyphs2ufo-ing GlyphsUnitTestSans.glyphs writes lib keys without values into bold "A". Ugh...

madig:
Okay, so `A` has the following keys in the Glyphs file:
```
script = "";
category = "";
subCategory = "";
```
The lib values are set correctly in `to_ufo_glyph()`:
```
(Pdb) ufo_glyph.lib[GLYPHLIB_PREFIX + 'category']
''
(Pdb) ufo_glyph.lib[GLYPHLIB_PREFIX + 'subCategory']
''
(Pdb) ufo_glyph.lib[GLYPHLIB_PREFIX + 'script']
''
```
So something must cause these values to not be properly written to the plist.

Doing
```
>>> import defcon
>>> u=defcon.Font()
>>> g=u.newGlyph("A")
>>> g.lib
{}
>>> g.lib["djksdjksdjksdkdjskdjskdjskjsdjks.category"] = ''
>>> g.lib["djksdjksdjksdkdjskdjskdjskjsdjks.category"]
''
>>> u.save("t.ufo")
```
works:
```
$ cat t.ufo/glyphs/A_.glif
<?xml version="1.0" encoding="UTF-8"?>
<glyph name="A" format="2">
  <outline>
  </outline>
  <lib>
    <dict>
      <key>djksdjksdjksdkdjskdjskdjskjsdjks.category</key>
      <string></string>
    </dict>
  </lib>
</glyph>
```

madig:
AAAAAHHH! I found the culprit: ufonormalizer deletes `<string></string>`!

anthrotype:
> ufonormalizer deletes <string></string>

hm that's bad, it shouldn't do that

madig:
Hrm. So this works, but I should leave the instance filename alone if it's set already.

madig:
@anthrotype under what condition can this be merged? Not immediately, still have to solve https://github.com/googlei18n/glyphsLib/pull/361#issuecomment-391093449.

Lib/glyphsLib/__main__.py
```diff
@@ -14,48 +14,185 @@
...
+        minimize_ufo_diffs=options.no_preserve_glyphsapp_metadata)
+
+    # Make the Glyphs file more suitable for roundtrip:
+    font.customParameters["Disable Last Change"] = options.enable_last_change
+    font.disablesAutomaticAlignment = options.enable_automatic_alignment
+
+    if options.output_path:
+        font.save(options.output_path)
+    else:
+        if locals().get("designspace_file"):
```
anthrotype: I don't like this trick. Set the variable to None, upfront

Lib/glyphsLib/__main__.py
```diff
@@ -14,48 +14,185 @@
...
+        font.save(filename_to_write)
+
+    return 0
+
 
-if __name__ == '__main__':
-    main(sys.argv[1:])
+if __name__ == "__main__":
+    print("Invoke either through glyphs2ufo or ufo2glyphs.", file=sys.stderr)
+    sys.exit(1)
```
anthrotype: the purpose of `__main__` is to be able to run a package (i.e. folder with `__init__.py`) as a module, like `python -m glyphsLib`. I would like to keep this feature.
I suggest you move the two entry points to an importable module, `glyphsLib.cli`, and in `glyphsLib.__main__` you reintroduce a `main` function that calls either `glyphs2ufo` or `ufo2glyphs` based on the first argument, like a sub-command, so one can still do `glyphs -m glyphsLib [glyphs2ufo|ufo2glyphs] ...`

tests/builder/interpolation_test.py
```diff
@@ -1,535 +0,0 @@
...
-        path = self.write_to_tmp_path(designspace, 'varfontorig.designspace')
-        doc = etree.parse(path)
-        medium = doc.find('sources/source[@stylename="Medium"]')
-        self.assertEqual(medium.find("lib").attrib["copy"], "1")
-        weightAxis = doc.find('axes/axis[@tag="wght"]')
-        self.assertEqual(weightAxis.attrib["default"], "444")
-
-        self.expect_designspace_roundtrip(designspace)
-
-    def test_designspace_name(self):
```
anthrotype: why did you remove the whole `interpolation_test.py` module? there are many tests that still apply to the current code. you can rename the test module (or move the relevant test elsewhere) if the lib module that it was originally meant to be testing is no longer present

Lib/glyphsLib/__main__.py
```diff
@@ -14,48 +14,185 @@
...
+        help=("Skip preserving Glyphs metadata in master UFOs and designspace "
+              "file, which would be used to minimize differences when "
+              "roundtripping between Glyphs and UFOs."),
+    )
+    group.add_argument(
+        "--propagate-anchors",
+        action="store_true",
+        help=("Copy anchors from underlying components to actual "
+              "glyph. Glyphs would do this implicitly, only use if you need "
+              "100% control over all anchors."),
```
khaledhosny: The `%` needs to be escaped here, otherwise running `glyph2ufo -h` fails for me with:
```
  File "/home/khaled/.local/bin/glyphs2ufo", line 11, in <module>
    sys.exit(glyphs2ufo())
  File "/home/khaled/.local/lib/python3.6/site-packages/glyphsLib/__main__.py", line 75, in glyphs2ufo
    options = parser.parse_args(args)
  File "/usr/lib/python3.6/argparse.py", line 1730, in parse_args
    args, argv = self.parse_known_args(args, namespace)
  File "/usr/lib/python3.6/argparse.py", line 1762, in parse_known_args
    namespace, args = self._parse_known_args(args, namespace)
  File "/usr/lib/python3.6/argparse.py", line 1968, in _parse_known_args
    start_index = consume_optional(start_index)
  File "/usr/lib/python3.6/argparse.py", line 1908, in consume_optional
    take_action(action, args, option_string)
  File "/usr/lib/python3.6/argparse.py", line 1836, in take_action
    action(self, namespace, argument_values, option_string)
  File "/usr/lib/python3.6/argparse.py", line 1020, in __call__
    parser.print_help()
  File "/usr/lib/python3.6/argparse.py", line 2362, in print_help
    self._print_message(self.format_help(), file)
  File "/usr/lib/python3.6/argparse.py", line 2346, in format_help
    return formatter.format_help()
  File "/usr/lib/python3.6/argparse.py", line 282, in format_help
    help = self._root_section.format_help()
  File "/usr/lib/python3.6/argparse.py", line 213, in format_help
    item_help = join([func(*args) for func, args in self.items])
  File "/usr/lib/python3.6/argparse.py", line 213, in <listcomp>
    item_help = join([func(*args) for func, args in self.items])
  File "/usr/lib/python3.6/argparse.py", line 213, in format_help
    item_help = join([func(*args) for func, args in self.items])
  File "/usr/lib/python3.6/argparse.py", line 213, in <listcomp>
    item_help = join([func(*args) for func, args in self.items])
  File "/usr/lib/python3.6/argparse.py", line 519, in _format_action
    help_text = self._expand_help(action)
  File "/usr/lib/python3.6/argparse.py", line 606, in _expand_help
    return self._get_help_string(action) % params
TypeError: %c requires int or char
```

Lib/glyphsLib/__main__.py
```diff
@@ -14,48 +14,185 @@
...
+        designspace = fontTools.designspaceLib.DesignSpaceDocument()
+        designspace.read(designspace_file)
+        object_to_read = designspace
+    elif all(
+            source.endswith(".ufo") and os.path.isdir(source)
+            for source in sources):
+        ufos = [defcon.Font(source) for source in sources]
+        ufos.sort(
+            key=lambda ufo: [  # Order the masters by weight and width
+                ufo.info.openTypeOS2WeightClass, ufo.info.openTypeOS2WidthClass],
```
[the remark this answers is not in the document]
madig:
```
In [11]: a=[1,2,3]

In [12]: a.sort(key=lambda ufo: [None, None])

In [13]: a
Out[13]: [1, 2, 3]
```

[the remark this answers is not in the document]
anthrotype: it will fail if one of the two is not None...  😛 

```python
>>> from defcon import Font
>>> f1 = Font()
>>> f2 = Font()
>>> f1.info.openTypeOS2WeightClass = 400
>>> f1.info.openTypeOS2WidthClass = 5
>>> # f2.info.openTypeOS2WidthClass = 500
>>> f2.info.openTypeOS2WidthClass = 4
>>> lst = [f2, f1]
>>> lst.sort(key=lambda ufo: [ufo.info.openTypeOS2WeightClass, ufo.info.openTypeOS2WidthClass])
TypeError: '<' not supported between instances of 'int' and 'NoneType'
>>> f2.info.openTypeOS2WeightClass = 500
>>> lst.sort(key=lambda ufo: [ufo.info.openTypeOS2WeightClass, ufo.info.openTypeOS2WidthClass])
>>> lst
[<defcon.objects.font.Font object at 0x10342a470>, <defcon.objects.font.Font object at 0x103432828>]
```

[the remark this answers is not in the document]
anthrotype: just use `lambda ufo: [ufo.info.openTypeOS2WeightClass or 400, ufo.info.openTypeOS2WidthClass or 5]`

Lib/glyphsLib/__main__.py
anthrotype: to escape, add another `%` sign after `100%`, as in `100%%`

madig: I replaced `100%` by `full` in my buffer.

tests/builder/interpolation_test.py
madig: Oops, I was overeager.

I noticed that fontmake relies on build_instances, though :(

anthrotype: > fontmake relies on build_instances

no, where? You mean "apply_instance_data" perhaps?

madig: Err, yeah, that one.

Lib/glyphsLib/__main__.py
```diff
@@ -14,48 +14,45 @@
...
+import glyphsLib.cli
 
 
-description = """\n
-Converts a Glyphs.app source file into UFO masters
-or UFO instances and MutatorMath designspace.
-"""
+def main():
+    if len(sys.argv) >= 2:
+        # Relay to subcommand.
```
anthrotype: clever, doesn't argparse already supports subcommands?

anthrotype:
merge conflict, needs to be rebased

Lib/glyphsLib/__main__.py
madig: Mh, yes. So I'm going to do the argparsing in main instead of the `*2*` functions.

madig: The separate ufo2glyphs and glyphs2ufo scripts need to go then I guess, as there is only main() and no parsing is done in the functions as such?

madig:
Merges again now. Can break up into smaller chunks again if desired.

setup.py
```diff
@@ -178,15 +178,14 @@ def run(self):
     packages=find_packages("Lib"),
     entry_points={
         "console_scripts": [
-            "glyphs2ufo = glyphsLib.__main__:main"
+            "glyphs-converter = glyphsLib.cli:main",
```
anthrotype: can't we just remove this console script entry point altogether and simply use `python -m glyphsLib [glyphs2ufo|ufo2glyphs]`? It's not much more typing after all.

Lib/glyphsLib/builder/instances.py
```diff
@@ -337,31 +336,24 @@ def apply_instance_data(designspace_path, include_filenames=None,
...
-        assert fname is not None, ("instance %r missing required filename" %
-                getattr(designspace_instance, "name", designspace_instance))
+        assert fname is not None, (
+            "instance %r missing required filename" % getattr(
+                designspace_instance, "name", designspace_instance))
         if include_filenames is not None:
             fname = normcase(normpath(fname))
             if fname not in include_filenames:
                 continue
-        logger.debug("Appling instance data to %s", fname)
```
anthrotype: oops sorry for the typo :stuck_out_tongue:

Lib/glyphsLib/cli.py
```diff
@@ -0,0 +1,226 @@
...
+def main(args=None):
+    if not args:
+        args = sys.argv[1:]
+
+    parser = argparse.ArgumentParser()
+    subparsers = parser.add_subparsers()
+
+    parser_glyphs2ufo = subparsers.add_parser(
+        "glyphs2ufo",
+        help="Converts a Glyphs.app source file into UFO masters or UFO instances and MutatorMath designspace.",
```
anthrotype: we don't make "UFO instances" in glyphsLib any more, do we?

Lib/glyphsLib/cli.py
```diff
@@ -0,0 +1,226 @@
...
+        options.glyphs_file,
+        options.output_dir,
+        options.instance_dir,
+        designspace_path=options.designspace_path,
+        minimize_glyphs_diffs=options.no_preserve_glyphsapp_metadata,
+        propagate_anchors=options.propagate_anchors,
+        normalize_ufos=options.no_normalize_ufos,
+    )
+
+    return 0
```
anthrotype: this is not required, since `sys.exit(None)` is the same as `sys.exit(0)`, but doesn't do any harm

tests/builder/instances_test.py
```diff
@@ -45,26 +45,35 @@ def test_apply_instance_data(tmpdir, instance_names):
...
     write_designspace_and_UFOs(designspace, path)
-    builder = DesignSpaceDocumentReader(designspace.path, ufoVersion=3)
+
+    test_designspace = DesignSpaceDocument()
+    test_designspace.read(designspace.path)
     if instance_names is None:
-        # generate all instances
-        builder.process()
-        include_filenames = None
+        # Generate all instances.
```
anthrotype: hm the comment says "generate all instances" but we aren't calling mutatormath any more.. where are these UFO instances coming from now? you sure this test code is even running?

setup.py
madig: Hm. Somewhat hesitant, but okay.

anthrotype: i'm okay if you want to have two separate console scripts `glyphs2ufo` and `ufo2glyphs`, but `glyphs-coverter [glyphs2ufo|ufo2glyphs]` seems too much typing for nothing, especially since `python -m glyphsLib [glyphs2ufo|ufo2glyphs]` would do the same (and I want to keep the latter).

maybe the original method you were using of parsing only the first argument in the top-level glyphsLib main, and then pass through all the following unparsed arguments to the respective sub-main functions (glyphs2ufo and ufo2glyphs) which do their own argparsing, works better than the argparse sub-parsers approach?
Then we could keep both a generic `python -m glyphsLib` invocation method (with `glyphs2ufo` and `ufo2glyphs` as subcommands of the latter), and two specific console scripts, respectively `glyphs2ufo` and `ufo2glyphs` for those who prefer less typing

madig: Sigh, yeah I'm not entirely happy with the current approach either... okay, so I'll revert back to the previous way.

madig: Wait. I got a better idea.
```
def _glyphs2ufo_entry_point():
    args = sys.argv[1:]
    args.insert(0, "glyphs2ufo")
    return main(args)
[...]
```
And in setup.py:
```
        "console_scripts": [
            "ufo2glyphs = glyphsLib.cli:_ufo2glyphs_entry_point",
            "glyphs2ufo = glyphsLib.cli:_glyphs2ufo_entry_point",
        ],
```

tests/builder/instances_test.py
madig: All tests pass:
```
$ tox -e py36 -- -k apply_instance_data
[...]
tests/builder/instances_test.py::test_apply_instance_data[default] PASSED [ 33%]
tests/builder/instances_test.py::test_apply_instance_data[include_1] PASSED [ 66%]
tests/builder/instances_test.py::test_apply_instance_data[include_2] PASSED [100%]
```
I think the comment is wrong, all the code does is iterate over the instances defined in the generated designspace file.

Lib/glyphsLib/cli.py
```diff
@@ -25,7 +25,7 @@ def main(args=None):
     if not args:
         args = sys.argv[1:]
 
-    parser = argparse.ArgumentParser()
+    parser = argparse.ArgumentParser(prog="{} -m glyphsLib".format(sys.executable))
```
anthrotype: maybe basename is better than the full sys.executable?

PR #371, opened 2018-06-11
Changing guideline angle handling to reflect change in UFO3 spec

jvgaultney:
The UFO3 spec has just been changed (https://github.com/unified-font-object/ufo-spec/issues/26) to acknowledge that most people and tools (specifically Glyphs and Robofont) interpret guideline angle to be CCW, despite what the spec originally stated. This provides the necessary fix and also updates appropriate tests.

### DIFF
--- a/Lib/glyphsLib/builder/guidelines.py
+++ b/Lib/glyphsLib/builder/guidelines.py
@@ -36,7 +36,6 @@ def to_ufo_guidelines(self, ufo_obj, glyphs_obj):
         new_guideline = {}
         x, y = guideline.position
         angle = guideline.angle
-        angle = (360 - angle) % 360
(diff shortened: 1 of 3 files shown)


Lib/glyphsLib/builder/guidelines.py
```diff
@@ -36,7 +36,6 @@ def to_ufo_guidelines(self, ufo_obj, glyphs_obj):
         new_guideline = {}
         x, y = guideline.position
         angle = guideline.angle
-        angle = (360 - angle) % 360
```
moyogo: The angle must be between 0 and 360 in UFO.
So you could change the previous line to `angle = guideline.angle % 360`.

PR #378, opened 2018-06-20
custom_params: support 'Reencode Glyphs'

anthrotype:
fixes https://github.com/googlei18n/glyphsLib/issues/205

There is no point in attempting to roundtrip this custom parameter, since it only applies to instances (which are not meant to be convertible back to Glyphs, but only for export).

I'll add a test case later. Just want to check with @belluzj @madig or @moyogo if they think this in the right direction.

It's an issue that affect the UI variant of Noto Sans Bengali, Malayalam and Devanagari. These use a combination of "Reencode Glyphs" and "Keep Glyphs" to deal with some UI glyph variants (more compressed vertically).

### DIFF
--- a/Lib/glyphsLib/builder/custom_params.py
+++ b/Lib/glyphsLib/builder/custom_params.py
@@ -564,6 +564,52 @@ class ReplaceFeatureParamHandler(AbstractParamHandler):
 register(ReplaceFeatureParamHandler())
 
 
+class ReencodeGlyphsParamHandler(AbstractParamHandler):
+    """ The "Reencode Glyphs" custom parameter contains a list of
+    'glyphname=unicodevalue' strings: e.g., ["smiley=E100", "logo=E101"].
+    It only applies to specific instance (not to master or globally) and is
+    meant to assign Unicode values to glyphs with the specied name at export
+    time.
+    When the Unicode value in question is already assigned to another glyph,
+    the latter's Unicode value is deleted.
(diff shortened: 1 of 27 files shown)


Lib/glyphsLib/builder/custom_params.py
```diff
@@ -564,6 +564,52 @@ def to_glyphs(self, glyphs, ufo):
 register(ReplaceFeatureParamHandler())
 
 
+class ReencodeGlyphsParamHandlder(AbstractParamHandler):
```
belluzj: Small typo in the name: Handler

belluzj:
Looks good to me (I mean implementation, I don't know about functionality-wise)

Lib/glyphsLib/builder/custom_params.py
anthrotype: fixed, thanks

anthrotype:
I added a test case with a designspace containing the 'Reencode Glyphs' custom parameter and then I run `apply_instance_data` and verify that the UFOs are re-encoded accordingly.
I'm going on holidays for a couple of weeks and would like to deliver this before I leave.
Sorry the rush, I'm going to merge this as soon as the bots go green.
Thanks for quick review Jany!

punchcutter:
A quick check on Bengali and Devanagari Sans UI looks good.

anthrotype:
Awesome, thanks for double-checking.

PR #379, opened 2018-07-11
Fix a roundtrip bug where the Regular would be renamed to Normal

belluzj:
@moyogo

### DIFF
--- a/Lib/glyphsLib/builder/axes.py
+++ b/Lib/glyphsLib/builder/axes.py
@@ -342,27 +342,41 @@ class AxisDefinition(object):
 
     def set_user_loc(self, master_or_instance, value):
         """Set the user location of a Glyphs master or instance."""
-        # Try to set the key if possible, i.e. if there is a key, and
-        # if there exists a code that can represent the given value, e.g.
-        # for "weight": 600 can be represented by SemiBold so we use that,
-        # but for 550 there is no code so we will have to set the custom
-        # parameter as well.
-        if (self.user_loc_key is not None and
-                hasattr(master_or_instance, self.user_loc_key)):
-            code = user_loc_value_to_instance_string(self.tag, value)
-            value_for_code = user_loc_string_to_value(self.tag, code)
-            setattr(master_or_instance, self.user_loc_key, code)
-            if self.user_loc_param is not None and value != value_for_code:
-                try:
-                    class_ = user_loc_value_to_class(self.tag, value)
-                    master_or_instance.customParameters[
-                        self.user_loc_param] = class_
-                except:
-                    pass
-
-        # Only masters can have an 'Axis Location' parameter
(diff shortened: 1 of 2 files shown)


moyogo:
LGTM. Thanks!

Lib/glyphsLib/builder/axes.py
```diff
@@ -342,28 +342,42 @@ def get_user_loc(self, master_or_instance):
...
+                code = user_loc_value_to_instance_string(self.tag, value)
+                value_for_code = user_loc_string_to_value(self.tag, code)
+                setattr(master_or_instance, self.user_loc_key, code)
+                if self.user_loc_param is not None and value != value_for_code:
+                    try:
+                        class_ = user_loc_value_to_class(self.tag, value)
+                        master_or_instance.customParameters[
+                            self.user_loc_param] = class_
+                    except:
+                        pass
```
anthrotype: usually `except: pass` smells.. what’s it for here?

belluzj: I think it's because `user_loc_value_to_class` raises `NotImplementedError` for axes other than weight/width.
https://github.com/googlei18n/glyphsLib/blob/baad5f67a9ffc14ffc005912003278e04743bd3b/Lib/glyphsLib/builder/axes.py#L96-L112

I could do have an `if` or `except NotImplementedError` if that's better

Lib/glyphsLib/builder/axes.py
```diff
@@ -342,28 +342,43 @@ def get_user_loc(self, master_or_instance):
...
+                code = user_loc_value_to_instance_string(self.tag, value)
+                value_for_code = user_loc_string_to_value(self.tag, code)
+                setattr(master_or_instance, self.user_loc_key, code)
+                if self.user_loc_param is not None and value != value_for_code:
+                    try:
+                        class_ = user_loc_value_to_class(self.tag, value)
+                        master_or_instance.customParameters[
+                            self.user_loc_param] = class_
+                    except NotImplementedError:
+                        # user_loc_value_to_class only works for weight & width
```
anthrotype: Much better, thanks

Lib/glyphsLib/builder/axes.py
```diff
@@ -342,28 +342,43 @@ def get_user_loc(self, master_or_instance):
...
+        # For masters, set directly the custom parameter (old way)
+        # and also the Axis Location (new way).
+        # Only masters can have an 'Axis Location' parameter.
+        if self.user_loc_param is not None:
+            try:
+                class_ = user_loc_value_to_class(self.tag, value)
+                master_or_instance.customParameters[
+                    self.user_loc_param] = class_
+            except:
+                pass
```
anthrotype: Oh, another one!

PR #382, opened 2018-07-19
Fix crash when round-tripping a GSFeaturePrefix containing only a comment

belluzj:


belluzj:
Thanks for the fast review!

anthrotype:
reviews are faster when a test accompanies them ;)

anthrotype:
feel free to tag 3.0.2
"continuous delivery" FTW! :+1:

Issue #386, opened 2018-07-30
Glyph data: glyph variants not properly categorized.


madig:
I stumbled over this bug: https://gitlab.gnome.org/GNOME/cantarell-fonts/issues/24. Inspecting the glyphsLib/fontmake generated binaries showed that several combining accents (in this case, dieresiscomb.case) had a non-zero advance width. Turns out that glyph variants are miscategorized:

```
In [1]: import glyphsLib

In [2]: glyphsLib.glyphdata.get_glyph("dieresiscomb")
Out[2]: Glyph(name='dieresiscomb', production_name='uni0308', unicode='̈', category='Mark', subCategory='Nonspacing')

In [3]: glyphsLib.glyphdata.get_glyph("dieresiscomb.case")
Out[3]: Glyph(name='dieresiscomb.case', production_name='dieresiscomb.case', unicode=None, category='Letter', subCategory=None)
```

madig:
Glyphs.app:
```
print(Glyphs.glyphInfoForName("brevecomb").subCategory)
=> Nonspacing

print(Glyphs.glyphInfoForName("brevecomb.case").subCategory)
=> Nonspacing
```

madig:
Apparently not that easy. Sigh.
```
Glyphs.glyphInfoForName("one_two.foo")
-> (u'Number', u'Ligature')
```

madig:
Groan.
```
Glyphs.glyphInfoForName("a_parallel.circled").productionName or .name
=> uni00612225.circled

Glyphs.glyphInfoForName("a_Dboldscript-math").productionName or .name
=> a_Dboldscript-math
```
So uXXXXX glyphs are rejected and uniXXXX with a .suffix in the name get the suffix appended when I don't think they should.

madig:
```
n=["a_a", "a_idotaccent", "a_brevecomb", "idotaccent_a", "idotaccent_brevecomb", "dieresiscomb", "dieresiscomb_a", "a_dieresiscomb"]

for name in n:
  i = Glyphs.glyphInfoForName(name)
  print(i.productionName or i.name)
```
```
a_a
a_i.loclTRK
uni00610306
i.loclTRK_a
idotaccent_brevecomb
uni0308
uni03080061
uni00610308
```
Uh...

schriftgestalt:
What would be the preferred name for `a_Dboldscript-math`? `a_u1D4D3`?

[...]

PR #387, opened 2018-07-30
Properly handle categorization of glyph variants

madig:
See #386.

### DIFF
--- a/Lib/glyphsLib/glyphdata.py
+++ b/Lib/glyphsLib/glyphdata.py
@@ -72,20 +72,22 @@ def get_glyph(glyph_name, data=glyphdata_generated):
 
     # Next, derive the actual characters from the production name, e.g.
     # "uni0414" -> "Д". Two caveats:
-    # 1. For some glyphs, Glyphs does not have a mapped character even when one 
+    # 1. For some glyphs, Glyphs does not have a mapped character even when one
     #    could be derived.
     # 2. For some others, Glyphs has a different idea than the agl module.
-    characters = None
+    unicode_characters = None
     if base_name not in data.MISSING_UNICODE_STRINGS:  # 1.
-        characters = data.IRREGULAR_UNICODE_STRINGS.get(base_name)  # 2.
-        if characters is None:
-            characters = agl.toUnicode(production_name) or None
+        unicode_characters = data.IRREGULAR_UNICODE_STRINGS.get(base_name)  # 2.
+        if unicode_characters is None:
+            unicode_characters = agl.toUnicode(production_name) or None
 
     # Lastly, generate the category in the sense of Glyph's
     # GSGlyphInfo.category and .subCategory.
-    category, sub_category = _get_category(base_name, characters, data)
+    category, sub_category = _get_category(base_name, unicode_characters, data)
 
-    return Glyph(glyph_name, production_name, characters, category, sub_category)
+    return Glyph(
+        glyph_name, production_name, unicode_characters, category, sub_category
+    )
 
 
 def _lookup_production_name(glyph_name, data=glyphdata_generated):
@@ -199,7 +201,7 @@ def _lookup_production_name(glyph_name, data=glyphdata_generated):
     return final_production_name
 
 
-def _get_unicode_category(character):
+def _get_unicode_category(unicode_characters):
     """Return the Unicode general category for a character (or the first
     character of a string).
 
@@ -210,15 +212,15 @@ def _get_unicode_category(character):
     thousand entries.
     """
 
-    if not character:
+    if not unicode_characters:
         return None
 
     if NARROW_PYTHON_BUILD:
-        utf32_str = character.encode("utf-32-be")
+        utf32_str = unicode_characters.encode("utf-32-be")
         nchars = len(utf32_str) // 4
         first_char = unichr(struct.unpack('>%dL' % nchars, utf32_str)[0])
     else:
-        first_char = character[0]
+        first_char = unicode_characters[0]
 
     return unicodedata.ucd_3_2_0.category(first_char)
 
--- a/Lib/glyphsLib/glyphdata.py
+++ b/Lib/glyphsLib/glyphdata.py
@@ -89,20 +89,26 @@ def get_glyph(name, data=glyphdata_generated):
     return Glyph(base_name, production_name, character, category, sub_category)
 
 
-def _get_unicode_category(unistr):
-    # We use data for a fixed Unicode version (3.2) so that our generated
-    # data files are independent of Python runtime that runs the rules.
-    # By switching to current Unicode data, we could save some entries
-    # in our exception tables, but the gains are not very large; only
-    # about one thousand entries.
-    if not unistr:
+def _get_unicode_category(character):
+    """Return the Unicode general category for a character.
+
+    We use data for a fixed Unicode version (3.2) so that our generated
+    data files are independent of Python runtime that runs the rules. By
+    switching to current Unicode data, we could save some entries in our
+    exception tables, but the gains are not very large; only about one
+    thousand entries.
+    """
+
+    if not character:
         return None
+
     if NARROW_PYTHON_BUILD:
-        utf32_str = unistr.encode("utf-32-be")
-        nchars = len(utf32_str)//4
+        utf32_str = character.encode("utf-32-be")
+        nchars = len(utf32_str) // 4
         first_char = unichr(struct.unpack('>%dL' % nchars, utf32_str)[0])
     else:
-        first_char = unistr[0]
+        first_char = character[0]
+
     return unicodedata.ucd_3_2_0.category(first_char)
 
 
--- a/Lib/glyphsLib/glyphdata.py
+++ b/Lib/glyphsLib/glyphdata.py
@@ -106,6 +106,12 @@ def _lookup_production_name(glyph_name, data=glyphdata_generated):
     - Suffix is e.g. "case".
     """
 
+    # The OpenType feature file specification says it's 63, the AGL says it's 31. We
+    # settle on 63. makeotf uses 63 as explained by Read Roberts from Adobe in
+    # https://github.com/fontforge/fontforge/pull/2500#issuecomment-143263393
+    # (Sep 25, 2015).
+    MAX_GLYPH_NAME_LENGTH = 63
+
     def is_unicode_u_value(name):
         return name.startswith("u") and all(
             part_char in "0123456789ABCDEF" for part_char in name[1:]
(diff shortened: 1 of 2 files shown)


madig:
Alright, this works. Now to clean it up.

@moyogo @belluzj comments?

tests/glyphdata_test.py
```diff
@@ -32,6 +32,10 @@ def test_production_name(self):
         self.assertEqual(prod("s_t"), "s_t")
         self.assertEqual(prod("Gcommaaccent"), "uni0122")
         self.assertEqual(prod("o_f_f_i.foo"), "o_f_f_i.foo")
+        self.assertEqual(prod("brevecomb"), "uni0306")
+        self.assertEqual(prod("brevecomb.case"), "uni0306.case")
+        self.assertEqual(prod("brevecomb_acutecomb"), "u'uni03060301")
```
schriftgestalt: > "u'uni03060301"
Is this correct?

madig:
@schriftgestalt any more thoughts on this?

Lib/glyphsLib/glyphdata.py
```diff
@@ -31,69 +31,224 @@
...
+
+    Note:
+    - Glyph name is the full name, e.g. "brevecomb_acutecomb.case".
+    - Base name is the base part, e.g. "brevecomb_acutecomb"
+    - Suffix is e.g. "case".
+    """
+
+    def is_unicode_u_value(name):
+        return name.startswith("u") and all(
+            part_char in "0123456789ABCDEF" for part_char in name[1:]
```
anthrotype: use a set of chars instead of a string literal here maybe

Lib/glyphsLib/glyphdata.py
```diff
@@ -31,69 +31,224 @@
...
+        or base_name in agl.AGL2UV
+        or glyph_name in (".notdef", ".null")
+    ):
+        return glyph_name
+
+    if glyph_name in data.PRODUCTION_NAMES:  # e.g. ain_alefMaksura-ar.fina -> uniFD13
+        return data.PRODUCTION_NAMES[glyph_name]
+    if base_name in data.PRODUCTION_NAMES:
+        final_production_name = data.PRODUCTION_NAMES[base_name] + dot + suffix
+        if len(final_production_name) > 31:
```
anthrotype: are you sure about the length limit of 31? I remember it being 64, but maybe i'm wrong.

Lib/glyphsLib/glyphdata.py
```diff
@@ -31,69 +31,224 @@
...
+        return None
+
+    # So we have a ligature that is not mapped in PRODUCTION_NAMES. Split it up and
+    # look up the individual parts.
+    base_name_parts = base_name.split("_")
+
+    # If all parts are in the AGLFN list, the glyph name is our production
+    # name already.
+    if all(part in agl.AGL2UV for part in base_name_parts):
+        if len(glyph_name) > 31:
```
anthrotype: again, we should confirm the 31 length limit

Lib/glyphsLib/glyphdata.py
```diff
@@ -31,69 +31,224 @@
...
+                    _character_outside_BMP = True
+
+            else:
+                return None
+
+    # Some names Glyphs uses resolve to other names that are not uniXXXX names and may
+    # contain dots (e.g. idotaccent -> i.loclTRK). If there is any name with a "." in
+    # it before the last element, punt. We'd have to introduce a "." into the ligature
+    # midway, which is invalid according to the AGL. Example: "a_i.loclTRK" is valid,
+    # but "a_i.loclTRK_a" isn't.
```
anthrotype: that's correct according to AGLFN, I just hope that Glyphs.app's names also always conform to that rule.. Do they?

Lib/glyphsLib/glyphdata.py
```diff
@@ -31,69 +31,224 @@
...
-
-
-def _get_unicode_category(unistr):
-    # We use data for a fixed Unicode version (3.2) so that our generated
-    # data files are independent of Python runtime that runs the rules.
-    # By switching to current Unicode data, we could save some entries
-    # in our exception tables, but the gains are not very large; only
-    # about one thousand entries.
-    if not unistr:
+        final_production_name = "{}".format("_".join(production_names)) + dot + suffix
```
anthrotype: the `"{}".format(...)` here seems redundant, no? The `"_".join(production_names)` already returns a `str`.

Lib/glyphsLib/glyphdata.py
```diff
@@ -31,69 +31,224 @@
...
+    thousand entries.
+    """
+
+    if not character:
+        return None
+
     if NARROW_PYTHON_BUILD:
-        utf32_str = unistr.encode("utf-32-be")
-        nchars = len(utf32_str)//4
+        utf32_str = character.encode("utf-32-be")
```
anthrotype: why did you rename `unistr` argument to `character`? It's technically a string as it can contain multiple characters. The previous name made more sense I think.

tests/glyphdata_test.py
```diff
@@ -44,6 +70,10 @@ def test_unicode(self):
         self.assertIsNone(uni("s_t"))  # no 'unicode' in GlyphsData
         self.assertEqual(uni("Gcommaaccent"), "Ģ")
         self.assertEqual(uni("o_f_f_i.foo"), "offi")
+        self.assertEqual(uni("brevecomb"), "̆")
```
anthrotype: maybe use the `\u0123` escape instead of a literal unicode string here, as it's not really legible

Lib/glyphsLib/glyphdata.py
madig: https://github.com/adobe-type-tools/agl-specification

> Length and character set limitations on names Glyphs from Western OpenType/CFF and TrueType fonts must still be referenced in many workflows as name-keyed font data. As a result, glyph names are still subject to the length and character set limitations that are imposed by the Type 1 specification and PostScript interpreter implementations. Both of these require that a glyph name be no longer than 31 characters in length [...]

Lib/glyphsLib/glyphdata.py
madig: Glyphs turns "a_idotaccent_a" into "a_i.loclTRK_a".

Lib/glyphsLib/glyphdata.py
anthrotype: ok.. but the Adobe Feature File spec says 63...
https://github.com/adobe-type-tools/afdko/blob/develop/docs/OpenTypeFeatureFileSpecification.html

anthrotype: I guess the 64th is a C-string NULL terminating byte :smile:

anthrotype: I mean, if makeotf allows 64, who are we to restrict it to 31?!

madig: Hmm... so, which one should it be? 😕 

@miguelsousa ?

anthrotype:
some of this logic is generic and not-glyphs.app specific, it would be nice to integrate it in ufo2ft PostProcessor which creates the production names (when an explicit postscriptNames in fontinfo.plist isn't provided).
There is also this https://github.com/googlei18n/fontmake/issues/455

Lib/glyphsLib/glyphdata.py
madig: Ok, let's just assume everything is broken anyway and if makeotf is doing it, so should we 😁👍

Lib/glyphsLib/glyphdata.py
anthrotype: that's not good. What does your code do in that case?

Lib/glyphsLib/glyphdata.py
anthrotype: yes, let's use 63 and paste a link to that Feature File spec paragraph in an inline comment, thanks

moyogo:
> are you sure about the length limit of 31? I remember it being 64, but maybe i'm wrong.

@anthrotype 

The [OpenType Feature File Specification]() does say.
> A glyph name may be up to 63 characters in length [...].

The [AGL-specification](https://github.com/adobe-type-tools/agl-specification) says:
> All glyph names a subject to a length-limit of 31 characters in some current software implementations, [...]. This length limit can be an issue with glyph names for ligature and ornaments.

Lib/glyphsLib/glyphdata.py
madig: https://github.com/googlei18n/glyphsLib/pull/387/files#diff-969a14671a2b5301f379a9ffe3bb13f6R56 😁

tests/glyphdata_test.py
```diff
@@ -32,6 +33,31 @@ def test_production_name(self):
...
+        self.assertEqual(prod("Dboldscript-math_Dboldscript-math.f"), "u1D4D3_u1D4D3.f")
+        self.assertEqual(prod("Dboldscript-math_a"), "u1D4D3_a")
+        self.assertEqual(prod("a_Dboldscript-math"), "a_u1D4D3")  # a_Dboldscript-math
+        self.assertEqual(prod("Dboldscript-math_a_aa"), "u1D4D3_a_uniA733")  # Dboldscript-math_a_aa
+        self.assertEqual(prod("Dboldscript-math_a_aaa"), "Dboldscript-math_a_aaa")
+        self.assertEqual(prod("brevecomb_Dboldscript-math"), "uni0306_u1D4D3")  # brevecomb_Dboldscript-math
+        self.assertEqual(prod("Dboldscript-math_brevecomb"), "u1D4D3_uni0306")  # Dboldscript-math_brevecomb
+        self.assertEqual(prod("idotaccent"), "i.loclTRK")
+        self.assertEqual(prod("a_idotaccent"), "a_i.loclTRK")
+        self.assertEqual(prod("a_idotaccent_a"), "a_idotaccent_a")  # a_i.loclTRK_a
```
anthrotype: makes sense

moyogo:
For me, the OT fea spec defines what goes in the feature file, before anything is converted to production names. The production names that go in the post table need to follow the AGL spec.

anthrotype:
anyway, it doesn't look like you're truncating them to 31. you just exit with None when it doesn't fit. What if the name that you didn't rename (because its production names exceeded 31) also exceeds the 31 char limit. That would also be invalid according to that aglfn spec

madig:
Yes. I punt on that one and hope that someday, some tool is going to rename stuff that doesn't fit to glyphXXXXX ^_^ or just not write it to the `post` table.

Lib/glyphsLib/glyphdata.py
madig:
```
In [5]:  hex_characters = {
   ...:    "0", "1", "2", "3", "4", "5", "6", "7", "8", "9", "A", "B", "C", "D", "E", "F",
   ...:  }

In [10]: %timeit all(part_char in hex_characters for part_char in uname[1:])
623 ns ± 3.89 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)

In [11]: %timeit all(part_char in "0123456789ABCDEF" for part_char in uname[1:])
567 ns ± 3.34 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
```
Let's keep the string literal :)

madig:
Regarding name length: maybe let's wait for clarification from Adobe.

Lib/glyphsLib/glyphdata.py
schriftgestalt: I check the ‘a_i.loclTRK_a’ name. That is clearly wrong.

Issue #388, opened 2018-08-06
incorrect handling of openTypeCodePageRanges custom parameter


BoldMonday:
Hi,

I am trying to convert a Glyphs file that only has `Thai` set in OS/2 CodePageRanges.

<img width="383" alt="thai" src="https://user-images.githubusercontent.com/13793769/43706845-c1c2d264-9966-11e8-8b6b-153941adb879.png">

But when using `glyphsLib.build_masters()` I get the following traceback:

```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/code/glyphslib/lib/python3.5/site-packages/glyphsLib/__init__.py", line 102, in build_masters
    minimize_glyphs_diffs=minimize_glyphs_diffs)
  File "/code/glyphslib/lib/python3.5/site-packages/glyphsLib/builder/__init__.py", line 90, in to_designspace
    return builder.designspace
  File "/code/glyphslib/lib/python3.5/site-packages/glyphsLib/builder/builders.py", line 201, in designspace
    ufos = list(self.masters)  # Make sure that the UFOs are built
  File "/code/glyphslib/lib/python3.5/site-packages/glyphsLib/builder/builders.py", line 139, in masters
    self.to_ufo_font_attributes(self.family_name)
  File "/code/glyphslib/lib/python3.5/site-packages/glyphsLib/builder/font.py", line 84, in to_ufo_font_attributes
    self.to_ufo_custom_params(ufo, font)
  File "/code/glyphslib/lib/python3.5/site-packages/glyphsLib/builder/custom_params.py", line 619, in to_ufo_custom_params
    handler.to_ufo(glyphs_proxy, ufo_proxy)
  File "/code/glyphslib/lib/python3.5/site-packages/glyphsLib/builder/custom_params.py", line 209, in to_ufo
    self._write_to_ufo(glyphs, ufo, ufo_value)
  File "/code/glyphslib/lib/python3.5/site-packages/glyphsLib/builder/custom_params.py", line 247, in _write_to_ufo
    ufo.set_info_value(self.ufo_name, value)
  File "/code/glyphslib/lib/python3.5/site-packages/glyphsLib/builder/custom_params.py", line 145, in set_info_value
    setattr(self._owner.info, name, value)
  File "/code/glyphslib/lib/python3.5/site-packages/defcon/objects/info.py", line 38, in setter
    raise ValueError("Invalid value ({0}) for attribute {1}.".format(repr(value), name))
[...]

PR #390, opened 2018-08-07
Correct handling of codePageRanges

madig:
#388

### DIFF
--- a/tests/builder/custom_params_test.py
+++ b/tests/builder/custom_params_test.py
@@ -143,7 +143,7 @@ class SetCustomParamsTest(unittest.TestCase):
         self.assertEqual(self.ufo.info.openTypeOS2CodePageRanges, [0, 1])
 
     def test_set_openTypeOS2CodePageRanges(self):
-        self.font.customParameters['openTypeOS2CodePageRanges'] = [0, 1]
+        self.font.customParameters['openTypeOS2CodePageRanges'] = [1252, 1250]
         self.set_custom_params()
         self.assertEqual(self.ufo.info.openTypeOS2CodePageRanges, [0, 1])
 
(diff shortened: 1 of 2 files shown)


tests/builder/custom_params_test.py
```diff
@@ -143,7 +143,7 @@ def test_set_codePageRanges(self):
         self.assertEqual(self.ufo.info.openTypeOS2CodePageRanges, [0, 1])
 
     def test_set_openTypeOS2CodePageRanges(self):
-        self.font.customParameters['openTypeOS2CodePageRanges'] = [0, 1]
+        self.font.customParameters['openTypeOS2CodePageRanges'] = [1252, 1250]
```
belluzj: Can you add a test that round-trips one of the "unsupported bits"? E.g. you put the bit in UFO, go to Glyphs, check that there is "codePageRangesUnsupportedBits", then go back to UFO and check that it's still here?

madig:
Empty list should stay empty list? Or become None?

moyogo:
> Empty list should stay empty list? Or become None?

Stay as an empty list, no?

madig:
I just noticed that openTypeOS2Selection gets tossed on a round-trip when it's an empty list.

madig:
Ok 😢

madig:
Ok, so I think this is ready for a review.

anthrotype:
regarding my comment above about not mixing unittest-style tests with pytest-style ones, I'm happy if we convert them all to pure pytest ones.

madig:
I might try my hand at this at one point. https://docs.pytest.org/en/latest/unittest.html has a few pointers on that, including mentioning https://pypi.org/project/unittest2pytest/.

schriftgestalt:
I just added handling of unsupported bits. They are stored as `bit XX` (where XX is the bit number)

Issue #391, opened 2018-08-09
Empty openTypeOS2Selection list is removed on round-trip


madig:
Round-tripping UFO -> Glyphs -> UFO with an empty openTypeOS2Selection list makes it disappear in the final UFO.

PR #392, opened 2018-08-09
Fix openTypeOS2Selection handling

madig:
See #391.


anthrotype:
test looks good.. so there's issue somewhere

madig:
I think enforcing sortedness is ok?

anthrotype:
I doesn't hurt if they are sorted

PR #393, opened 2018-08-09
Round-trip italicAngle == zero

madig:


madig:
Hm... this might be more tricky than expected. Glyphs deletes the property if it's zero.

madig:
Does this make more sense? 🤔

anthrotype:
yeah

Issue #402, opened 2018-08-21
Updating GlyphsData.xml fails


madig:
```
glyphsLib (master) $ venv/bin/python MetaTools/generate_glyphdata.py 
Traceback (most recent call last):
  File "MetaTools/generate_glyphdata.py", line 282, in <module>
    test_data(glyphs, data)
  File "MetaTools/generate_glyphdata.py", line 185, in test_data
    assert prod == g.production_name, (name, prod, g.production_name)
AssertionError: ('A-osage', 'u104B0', 'A-osage')
```

madig:
It actually fails on a lot of things: [glyphdata-fail.txt](https://github.com/googlei18n/glyphsLib/files/2308016/glyphdata-fail.txt)

madig:
Duh. I forgot to pass the `data` parameter in https://github.com/googlei18n/glyphsLib/blob/master/Lib/glyphsLib/glyphdata.py#L55. Still some failures.

PR #403, opened 2018-08-21
Fix glyphsdata updating

madig:
#402.

Still some failures: [glyphdata-fail.txt](https://github.com/googlei18n/glyphsLib/files/2308156/glyphdata-fail.txt)

What to do about Unicode mismatches? We come up with a Unicode string whereas GlyphsData.xml seems to specify only a single Unicode point for a glyph or None?

### DIFF
--- a/Lib/glyphsLib/glyphdata.py
+++ b/Lib/glyphsLib/glyphdata.py
@@ -83,7 +83,7 @@ def get_glyph(glyph_name, data=glyphdata_generated):
 
     # Lastly, generate the category in the sense of Glyph's
     # GSGlyphInfo.category and .subCategory.
-    category, sub_category = _get_category(base_name, unicode_characters, data)
+    category, sub_category = _get_category(base_name, unicode_characters, data=data)
 
     return Glyph(
         glyph_name, production_name, unicode_characters, category, sub_category
--- a/Lib/glyphsLib/glyphdata.py
+++ b/Lib/glyphsLib/glyphdata.py
@@ -106,10 +106,7 @@ def _lookup_production_name(glyph_name, data=glyphdata_generated):
     - Suffix is e.g. "case".
     """
 
-    # The OpenType feature file specification says it's 63, the AGL says it's 31. We
-    # settle on 63. makeotf uses 63 as explained by Read Roberts from Adobe in
-    # https://github.com/fontforge/fontforge/pull/2500#issuecomment-143263393
-    # (Sep 25, 2015).
+    # The AGLFN has been amended to allow a maximum of 63 characters in a glyph name.
     MAX_GLYPH_NAME_LENGTH = 63
 
     def is_unicode_u_value(name):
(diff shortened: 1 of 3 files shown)


madig:
The update script goes through the parsed database and asserts that what we made of it matches what's actually in the database (https://github.com/googlei18n/glyphsLib/blob/master/MetaTools/generate_glyphdata.py#L184).

E.g. `('caa-kannada', None, 'Halfform')` means that caa-kannada's subcategory should be None according to the database but is "Halfform" according to Unicode.

madig:
While you're here, should doChada-thai.short have the production name "uni0E0E.short"?
```
<glyph unicode="0E0E" name="doChada-thai" category="Letter" script="thai" production="uni0E0E" altNames="dochadathai" description="THAI CHARACTER DO CHADA" />
<glyph name="doChada-thai.short" category="Letter" script="thai" description="THAI CHARACTER DO CHADA SHORT" />
```

madig:
Georg says that no subCategory implies "Other", maybe that should be the fallback. :thinking:

schriftgestalt:
How do you find a unicode-category for `caa-kannada` that doesn't has a unicode? It might be a conjunct?

schriftgestalt:
> While you're here, should doChada-thai.short have the production name "uni0E0E.short"?

I fixed that.

madig:
Ah, found it. In generate_glyphdata.py, build_categories(), default_categories[None] == ("Letter", "Halfform"). So it's a generation bug?

schriftgestalt:
I’m not sure what you are doing there exactly but that seems dangerous.

madig:
I'm trying to change the generation script so I can remove some special casing in glyphsdata.py. I'm doing random stuff until I find something that works :+1:

madig:
Hm.. maybe need to change _get_category instead to build irregular map...

madig:
Alright, this correctly updates the database now, except for the following production names:
```
('caroncomb.alt', 'caroncomb.alt', 'uni030C.alt')
('doChada-thai.short', 'doChada-thai.short', 'uni0E0E.short')
('karan-lao.small', 'karan-lao.small', 'uni0ECC.small')
('loChula-thai.short', 'loChula-thai.short', 'uni0E2C.short')
('maiCatawa-lao.small', 'maiCatawa-lao.small', 'uni0ECB.small')
('maiEk-lao.small', 'maiEk-lao.small', 'uni0EC8.small')
('maiTho-lao.small', 'maiTho-lao.small', 'uni0EC9.small')
('maiTi-lao.small', 'maiTi-lao.small', 'uni0ECA.small')
('niggahita-lao.small', 'niggahita-lao.small', 'uni0ECD.small')
('thoThan-thai.less', 'thoThan-thai.less', 'uni0E10.less')
('toPatak-thai.short', 'toPatak-thai.short', 'uni0E0F.short')
('yoYing-thai.less', 'yoYing-thai.less', 'uni0E0D.less')
```
(middle is what Glyphs would use as a production name, right what we output)

schriftgestalt:
caroncomb.alt should only ever be used as a component and should never be "typed".

schriftgestalt:
I will fix the rest in Glyphs.

madig:
So I have to special case it...

PR #405, opened 2018-08-25
fix glyph's single unicode should be parsed as string, but integer

shengdoushi:
'{unicode = 0008;}'

 is parsed to:

            [('unicode', 8)]

but should be:

            [('unicode', '0008')]

### DIFF
--- a/Lib/glyphsLib/parser.py
+++ b/Lib/glyphsLib/parser.py
@@ -31,7 +31,7 @@ class Parser(object):
     """Parses Python dictionaries from Glyphs source files."""
 
     value_re = r'(".*?(?<!\\)"|[-_./$A-Za-z0-9]+)'
-    unicode_list_re = re.compile(r'\s*([0-9a-fA-F]+(,[0-9a-fA-F]+)+)')
+    unicode_list_re = re.compile(r'\s*([0-9a-fA-F]+(,[0-9a-fA-F]+)*)')
     start_dict_re = re.compile(r'\s*{')
     end_dict_re = re.compile(r'\s*}')
     dict_delim_re = re.compile(r'\s*;')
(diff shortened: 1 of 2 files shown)


shengdoushi:
I signed it!

Lib/glyphsLib/parser.py
```diff
@@ -105,7 +105,7 @@ def _parse(self, text, i, _parsing_unicodes=False):
                 parsed = m.group(0)
                 i += len(parsed)
                 unicode_list = m.group(1).split(",")
-                return unicode_list, i
+                return len(unicode_list)==1 and unicode_list[0] or unicode_list, i
```
anthrotype: thank god black will take care of these missing spaces around binary operator from now on

belluzj:
LGTM, thanks for the test!

Lib/glyphsLib/parser.py
belluzj: also it should be an inline/ternary if-else, right?

anthrotype: i think it would be better if this always returned a list, even if it's only length one.
@belluzj ?

belluzj: I'm not sure what happens after this, I'll have to look if there's more special casing for a single unicode, maybe we use it for roundtripping correctly or something like that :(

moyogo: Always a list would make more sense.

belluzj: Yes it would. Also if you always return a list, you can probably delete those two lines that deal with a string:
https://github.com/googlei18n/glyphsLib/blob/a1a0ab59cb40ff689de2faf574259c80ee8c9b20/Lib/glyphsLib/types.py#L382-L383

belluzj: Maybe we can merge and fix this ourselves? I can commit a fix. I'm also going to look into the issue about bad axis mappings (because I made the suspicious code for axes)

anthrotype: that would be great, thanks guys!

PR #407, opened 2018-08-28
Fix a roundtrip failure with weird kerning rules

belluzj:
We encountered a case where glyphsLib's logic to remove "ambiguous" kerning rules would actually change the outcome of the kerning rules.

Here I fixed the bug, but maybe it would also be an opportunity to review the usefulness of that kerning logic? If I understand correctly, the "ambiguity" that the code removes is not considered ambiguous since the UFO3 spec says:

> The two second level pairs, public.kern1.O + F and Q + public.kern2.E define two possible values for the pair Q + F. This is an unresolvable ambiguity.
> 
> To resolve this, glyph + group pairs are given higher priority than group + glyph pairs when these contradictions occur. In the example above, the Q + F pair value would be -250.

So maybe we could get rid of that logic in glyphsLib? What do you think?


moyogo:
LGTM

Lib/glyphsLib/builder/kerning.py
```diff
@@ -54,7 +54,12 @@ def _to_ufo_kerning(self, ufo, kerning_data):
             ufo.kerning[left, right] = kerning_val
 
     seen = {}
-    for classname, glyph, is_left_class in reversed(class_glyph_pairs):
```
anthrotype: why was it `reversed` before?

belluzj: Blaming gives this: https://github.com/googlei18n/glyphsLib/commit/439676094ce3fc155ffdc4228447aa09695a25c0

anthrotype: > kerning conflicts are checked and resolved in reverse order of how they appear in Glyphs source.

hm...  but isn't `kerning_data.items()` an unsorted collection?

anthrotype: no, it's an OrderedDict apparently

anthrotype:
looking at the 4396760 commit message, it appears that James was trying to reproduce Glyphs.app's behavior in resolving kerning conflicts. The assumption is that kerning dictionary is ordered in the glyphs file, the order is the order the pairs appear in the file itself. What is this order, by the way?
In UFO this assumption does not hold. Kerning doesn't have any arbitrary order intrinsically. It's only sorted alphabetically for serialization purposes.

Lib/glyphsLib/builder/kerning.py
```diff
@@ -54,7 +54,12 @@ def _to_ufo_kerning(self, ufo, kerning_data):
...
+    # Sort by the 3rd tuple member: `is_left_class`, so that pairs with
+    # a class on the right are seen first, and thus have priority when
+    # solving ambiguous cases, as per the UFO spec at:
+    # http://unifiedfontobject.org/versions/ufo3/kerning.plist/#exception-conflict-resolution
+    for classname, glyph, is_left_class in sorted(
+            class_glyph_pairs, key=lambda pair: pair[2]):
```
anthrotype: maybe the lambda function here could return a tuple with the first item being `pair[2]`, and the second and third items being `pair[0], pair[1]`. So that within the group that have the same `is_left_class` pairs are sorted alphabetically.

anthrotype: alternatively we could change `class_glyph_pairs` list's tuples to already be in the correct order, so instead of

```
for classname, glyph, is_left_class in sorted(..., key=lambda ...)
```

we could have

```
for is_left_class, classname, glyph in sorted(...)
```

belluzj:
@anthrotype That's why I'm raising the question of the relevance of this whole logic.
* On the UFO side there are rules such that kerning is never ambiguous/there are no "conflicts"
* On the Glyphs side, we can ask @schriftgestalt or test again to see if it still behaves like in 2016

Best case scenario: Glyphs now agrees with the UFO spec, we dump the whole "conflict resolution" from glyphsLib

Other scenario: Glyphs still does like in 2016, and then we only need to resolve "conflicts" when Glyphs' behaviour gives a different result than UFO

WDYT?

anthrotype:
i would be fine to remove the conflict resolution logic altogether honestly...

anthrotype:
in the ufo2ft kernFeatureWriter we already collect kerning pairs in the right UFO order, by `is_left_class` flag: i.e. `[(False, False), (False, True), (True, False), (True, True)]`

https://github.com/googlei18n/ufo2ft/blob/70a73822832f71516a095bd7595b4fff26e33c1e/Lib/ufo2ft/featureWriters/kernFeatureWriter.py#L321

and feaLib also takes care of conflicting `enum pos` rules here (with a warning):
https://github.com/fonttools/fonttools/blob/32aef4fc880f4c8e9c564ea1f48849fe1f06cbe7/Lib/fontTools/feaLib/builder.py#L1525-L1532

belluzj:
The risk I see with removing that logic entirely is if someone uses those "ambiguous" rules in Glyphs and actually wants the result that Glyphs gives, and then when making UFOs get a different result. I'm not sure how often those rules appear in the wild, but they did happen on one of our projects somehow...

anthrotype:
the problem is we don't know how Glyphs.app actually resolves these conflicts. We only know what's in the UFO and the Feature File spec.

anthrotype:
I need to go into a meeting now, i'll get back to you in half an hour or so.
What we need to do before merging this or dropping the logic altogether is ascertain what Glyphs.app actually does with kerning conflicts.

schriftgestalt:
I did explain this before somewhere. I don't do active conflict resolution. Glyphs builds a dict for class+class pairs and glyph+glyph pairs (including flattering class+glyph or glyph+class). Because it is a dict, the last rule will win.

belluzj:
> So it somehow depends on the sorting order.

@schriftgestalt Yes :) in the UFO spec, there is a section which gives an example of ambiguous kerning and explains how to solve it by (arbitrarily) always choosing the glyph+class pairs over the class+glyph pairs. With your approach, if would be nice if Glyphs would always flatten first class+glyph pairs, then glyph+class pairs, so that the same pair always "wins" in Glyphs and in UFOs, and then we can remove the code here.
http://unifiedfontobject.org/versions/ufo3/kerning.plist/#exception-conflict-resolution

From what you're saying, I understand that Glyph currently relies on the order of kerning pairs in the file or something like that (I haven't investigated). If that's the case, then I think it would be beneficial also for Glyphs to have a more "predictable" output, ideally following the same rules as UFOs.

anthrotype:
and in what order are the kerning rules written out to the .glyphs file?

madig:
[weird-kerning.zip](https://github.com/googlei18n/glyphsLib/files/2360725/weird-kerning.zip)

`weird-kerning like ufo.glyphs` contains the kerning order that happened after we input the example given in http://unifiedfontobject.org/versions/ufo3/kerning.plist/#exception-conflict-resolution, saved and reloaded (see included screenshot). In `weird-kerning edited copy.glyphs`, we edited the kerning by hand and Glyphs reordered it on opening it.

madig:
```
kerning = {
"7A09CD0E-70BC-49AB-97E3-EC297BE3BC37" = {
Q = {
"@MMK_R_E" = -250;
};
"@MMK_L_O" = {
"@MMK_R_E" = -100;
F = -200;
};
D = {
F = -300;
};
};
};
```

madig:
Saving back results in what the screenshot shows and the other glyphs file contains.

anthrotype:
I don't see what's the logic in that reordering.

schriftgestalt:
The pairs are ordered alphabetically by classID or glyph name. The classID uses the `@MMK_L_XX` and thous the classes are always first.

schriftgestalt:
And that way the glyph+class pair always wins over the class+glyph pair.

belluzj:
> And that way the glyph+class pair always wins over the class+glyph pair.

Yay! I'll remove the conflict resolution code right now

PR #414, opened 2018-09-06
Reformat code (2)

anthrotype:
let's work on #413 together on a shared branch on the upstream repo

### DIFF
--- a/Lib/glyphsLib/builder/font.py
+++ b/Lib/glyphsLib/builder/font.py
@@ -14,11 +14,10 @@
 
 from __future__ import print_function, division, absolute_import, unicode_literals
 
-from collections import deque, OrderedDict
 import logging
 
 from .common import to_ufo_time, from_ufo_time
-from .constants import GLYPHS_PREFIX, GLYPHLIB_PREFIX
+from .constants import GLYPHS_PREFIX
 
 logger = logging.getLogger(__name__)
 
--- a/Lib/glyphsLib/builder/font.py
+++ b/Lib/glyphsLib/builder/font.py
@@ -47,7 +47,8 @@ def to_ufo_font_attributes(self, family_name):
     designer_url = font.designerURL
     manufacturer = font.manufacturer
     manufacturer_url = font.manufacturerURL
-    note = font.note
+    # XXX note is unused?
+    # note = font.note
     glyph_order = list(glyph.name for glyph in font.glyphs)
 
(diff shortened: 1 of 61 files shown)


belluzj:
Before you blacken everything, would you want to review and merge the two other open PRs?
* https://github.com/googlei18n/glyphsLib/pull/405
* https://github.com/googlei18n/glyphsLib/pull/407

I'm afraid they might have conflicts after that, and your blackening thing is all automated right?

anthrotype:
@belluzj good point Jany. I'm just setting things up so that we are ready to reformat everything. I'll wait for the other unfinished stuff before merging this.

anthrotype:
I also added flake8 and started fixing some stuff but the list is too long, will finish in next days

Lib/glyphsLib/builder/font.py
```diff
@@ -47,7 +47,8 @@ def to_ufo_font_attributes(self, family_name):
     designer_url = font.designerURL
     manufacturer = font.manufacturer
     manufacturer_url = font.manufacturerURL
-    note = font.note
+    # XXX note is unused?
+    # note = font.note
```
madig: Note to self: test this

tests/writer_test.py
```diff
@@ -780,7 +780,8 @@ def test_write_layer(self):
         # bounds: read-only, computed
         # selectionBounds: read-only, computed
         # background
-        bg = layer.background
+        # XXX bg is unused?
```
madig: Note to self: Investigate.

madig:
PyCharm has even more checks. Let's run a few.

Lib/glyphsLib/classes.py
```diff
@@ -65,6 +65,7 @@
     "GSHint",
     "GSBackgroundImage",
     # Constants
+    "__all__",
```
anthrotype: Not sure adding __all__ to __all__ is a good idea..

madig:
Oh boy... there are quite a few warnings about attributes being defined outside `__init__`, e.g. https://github.com/googlei18n/glyphsLib/blob/master/Lib/glyphsLib/classes.py#L1338

anthrotype:
We can’t fix everything now. Let’s start with making flake8 pass

madig:
Ok, so I'll stop with PyCharm now :) The rest looks okay to me.

PR #415, opened 2018-09-06
Make UFO normalization optional

### LINKED ISSUE #412: glyphs2ufo script requires ufonormalizer
but ufonormalizer is not explicitly listed in install_requires, so installing glyphsLib in a fresh environment leads to import error (unless one passes a negative --no-normalize-ufos option).

I would prefer if the option was --normalize-ufos (or just --normalize), while keeping ufonormalizer optional.

madig:
Fixes #412.

### DIFF
--- a/Lib/glyphsLib/cli.py
+++ b/Lib/glyphsLib/cli.py
@@ -81,6 +81,7 @@ def main(args=None):
         ),
     )
     group.add_argument(
+        "-N",
         "--normalize-ufos",
         action="store_true",
         help=(
--- a/Lib/glyphsLib/cli.py
+++ b/Lib/glyphsLib/cli.py
@@ -81,10 +81,10 @@ def main(args=None):
         ),
     )
     group.add_argument(
-        "--no-normalize-ufos",
-        action="store_false",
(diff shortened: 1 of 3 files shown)


Lib/glyphsLib/cli.py
```diff
@@ -81,10 +81,10 @@ def main(args=None):
         ),
     )
     group.add_argument(
-        "--no-normalize-ufos",
-        action="store_false",
+        "--normalize-ufos",
```
anthrotype: can we add a short option as well? `-N` comes to mind

setup.py
```diff
@@ -55,6 +55,9 @@
         "fonttools>=3.24.0",
         "defcon>=0.3.0",
     ],
+    extras_require = {
+        "ufo_normalization":  ["ufonormalizer"]
```
anthrotype: yeah I thought about that, then I thought... hm, the name of the extra is longer than the name of the only extra requirement, "ufonormalizer" :D
But it's ok, leave it in. I doubt anyone will ever use it, but at least it's an explicit signal that there is an optional requirement.

Issue #280, opened 2017-11-08
Use `Axes` property


schriftgestalt:
There is a custom parameter `Axes`. It contains axis names and tags. It is required for variable font export and does limit the available coordinate fields in the master and instance settings.

anthrotype:
any documentation available yet?

anthrotype:
@schriftgestalt in #279 you also mentioned a "Virtual Master". What's that? You may have already explained this to me in Montreal but I forgot :grimacing:

schriftgestalt:
the format of the Axes parameter:
```
{
	name = Axes;
	value = (
		{
			Name = Weight;
			Tag = wght;
		},
		{
			Name = Width;
			Tag = wdth;
		}
	);
}
```

anthrotype:
and the axis location for each master?

schriftgestalt:
That is defined in the masters settings. They have coordinates for each axis.

m4rc1e:
> It contains axis names and tags. It is required for variable font export

As far as I am aware, this only happens in Glyphs

![screen shot 2017-11-08 at 10 55 45](https://user-images.githubusercontent.com/7525512/32545361-655eed14-c473-11e7-8b38-c99b25c28949.png)


In Fontmake, it will automatically gen axises if non are specified. I love this because I am dealing with many families which do not contain an axis property. Could Fontmake and its supporting libs such as this one still retain this functionality?


Apologies if this is slightly off topic.

schriftgestalt:
Keeping the old mechanism as a fallback might be a good idea. But optimise it to omit "empty" axes.

anthrotype:
@m4rc1e I think we will keep that even as we add support for the new parameters.

@schriftgestalt I mean, the user-space locations for registered axes that have predefined rangee like weight (1-1000) or width (percentage). We are currently getting these from the instances to generate the axes map elements in the designspace (used to build avar table) because in the old Glyphs.app masters only had design-space locations (any arbitary number used by the designers).

schriftgestalt:
There is master parameter for that:
```
{
	name = "Axis Location";
	value = (
		{
			Axis = Weight;
			Location = 100;
		},
		{
			Axis = Width;
			Location = 200;
		}
	);
}
```
You can’t remap Virtual masters. But those are most likely used for non standard axes that don't have a predefined range so it is fine using the internal numbers. (I think of something if someone presents a valid use case).

anthrotype:
[...]

Issue #411, opened 2018-09-05
to_designspace might be producing incorrect axis maps for Fah Kwang


m4rc1e:
Whilst attempting to gen a variable font for [Fah Kwang](https://github.com/m4rc1e/Fah-Kwang/tree/vf), using fontmake 1.7.2. I found the vf instances do not match the static fonts.

The glyphs source consists of six masters:

![screen shot 2018-09-05 at 19 05 10](https://user-images.githubusercontent.com/7525512/45112102-a179c180-b13e-11e8-8aa9-43c54e8dfd88.png)

In the source, the masters axis weight coordinates are [200, 400, 700] (ExtraLight to Bold)

In the genned designspace we get the following weight axis

```
        <axis default="400" maximum="700" minimum="300" name="Weight" tag="wght">
            <map input="300" output="200" />
            <map input="400" output="400" />
            <map input="700" output="700" />
        </axis>
```

If I overlay the vf instances against the same family as static fonts, the Light instance doesn't match. In fact, it has the same weight as the ExtraLight instance.

![screen shot 2018-09-05 at 19 16 15](https://user-images.githubusercontent.com/7525512/45112664-2dd8b400-b140-11e8-9fb1-f60e0e4b5276.png)
*Cyan: VF Light Instance, Black: Light static*

If I change the first map input from 300 to 200. I get the correct result.

![screen shot 2018-09-05 at 19 29 37](https://user-images.githubusercontent.com/7525512/45113356-0edb2180-b142-11e8-9807-eb2f3a499e92.png)
*Cyan: VF Light Instance, Black: Light static*

I suspect the issue is lingering [here](https://github.com/googlei18n/glyphsLib/blob/master/Lib/glyphsLib/builder/axes.py#L155-L156)

Btw, even though maps exist in the designspace, no avar table is being genned.

anthrotype:
Looks similar to this other issue @madig raised the other day:
https://github.com/googlei18n/glyphsLib/issues/409

I’ll have a look tomorrow.

PR #416, opened 2018-09-06
Fix axes mapping when using Axes custom param, fixes #409, #411

### LINKED ISSUE #409: OS/2 weight classes jumbled when having explicit Axes parameter
In this file: [Cantarell.glyphs.zip](https://github.com/googlei18n/glyphsLib/files/2346190/Cantarell.glyphs.zip) I have different weights and instances for them. When I leave the global "Axes" parameter in, the weight ranges get jumbled and e.g. Thin starts at 300 instead of 100 (The designspace file defines the axis from 300 to 700 even though it should be from 100 to 800). Deleting the parameter makes the weight classes assign as expected.

belluzj:
I tested on Fah-Kwang, the new axes are like that:

```xml
    <axes>
        <axis default="400" maximum="700" minimum="200" name="Weight" tag="wght" />
        <axis default="0" maximum="10" minimum="0" name="Italic" tag="ital" />
    </axes>
```

The code wrongly assumed that the "Axes" custom parameter was always used in conjunction with the "Axis Location" parameters for masters.

### DIFF
--- a/Lib/glyphsLib/builder/axes.py
+++ b/Lib/glyphsLib/builder/axes.py
@@ -208,7 +208,13 @@ def to_designspace_axes(self):
 
 
 def font_uses_new_axes(font):
-    return bool(font.customParameters['Axes'])
+    # It's possible for fonts to have the 'Axes' parameter but to NOT specify
+    # the master locations using 'Axis Location', in which case we have to
+    # resort to using instances or other old tricks to get the mapping.
+    # https://github.com/googlei18n/glyphsLib/issues/409
+    # https://github.com/googlei18n/glyphsLib/issues/411
+    return font.customParameters['Axes'] and all(
+        master.customParameters['Axis Location'] for master in font.masters)
 
 
 def to_glyphs_axes(self):
--- a/tests/builder/axes_test.py
+++ b/tests/builder/axes_test.py
@@ -203,3 +203,35 @@ def test_master_user_location_goes_into_os2_classes():
 
     assert black.info.openTypeOS2WeightClass == 920
     assert black.info.openTypeOS2WidthClass == 1
+
+
+def test_mapping_is_same_regardless_of_axes_custom_parameter():
+    # https://github.com/googlei18n/glyphsLib/issues/409
+    # https://github.com/googlei18n/glyphsLib/issues/411
+
+    # First, try without the custom param
+    font = to_glyphs([defcon.Font(), defcon.Font(), defcon.Font()])
+    font.masters[0].name = 'ExtraLight'
+    font.masters[0].weightValue = 200
+    font.masters[1].name = 'Regular'
+    font.masters[1].weightValue = 400
+    font.masters[2].name = 'Bold'
+    font.masters[2].weightValue = 700
+
+    doc = to_designspace(font)
+    assert doc.axes[0].minimum == 200
+    assert doc.axes[0].maximum == 700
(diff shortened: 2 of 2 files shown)


madig:
That's funny. I thought I specified the axis location in the master dialog...

belluzj:
Can you test the branch with Cantarell? I didn't

Lib/glyphsLib/builder/axes.py
```diff
@@ -208,7 +208,13 @@ def to_designspace_axes(self):
...
 
 def font_uses_new_axes(font):
-    return bool(font.customParameters['Axes'])
+    # It's possible for fonts to have the 'Axes' parameter but to NOT specify
+    # the master locations using 'Axis Location', in which case we have to
+    # resort to using instances or other old tricks to get the mapping.
+    # https://github.com/googlei18n/glyphsLib/issues/409
+    # https://github.com/googlei18n/glyphsLib/issues/411
+    return font.customParameters['Axes'] and all(
+        master.customParameters['Axis Location'] for master in font.masters)
```
anthrotype: and what happens in Glyphs.app if only some of the masters have Axis Location?

belluzj: I can't test in Glyphs at the moment

madig:
The version in question is https://gitlab.gnome.org/GNOME/cantarell-fonts/raw/801f581de296a1d2d0ddca3db6565d4bca5044d3/src/Cantarell.glyphs. This PR makes it work.

Lib/glyphsLib/builder/axes.py
madig: You could do a all() test and log a warning if some don't have a location and fall back to the old behavior. Or blow up.

madig:
I don't understand why it didn't work before. Axis locations were all specified?

belluzj:
In your Cantarell file, none of the masters have the Axis Location custom parameter

madig:
But "Axis coordinates" are all specified?!

anthrotype:
oh gosh.. what's "Axis Coordinates" now?!

madig:
I suppose a new way to specify coordinates :O

belluzj:
"Axis Location" is something else than this number box, it's a way to map master design locations (the number box in the screenshot) to user locations. It was explained by Georg here: https://github.com/googlei18n/glyphsLib/issues/280#issuecomment-342786012

Your Cantarell files doesn't have the Axis Location custom param as seen here:
image


The axes coordinates above are nothing new, they actually end up in the glyphs file in the same fields as `weightValue`, `widthValue`, `customValue` etc, as confirmed by Georg there: https://github.com/googlei18n/glyphsLib/issues/280#issuecomment-374912389

madig:
So, how to proceed? Does your fix check all the boxes?

belluzj:
I think it does, it fixes your two problems and there are no regressions. By the way we could also probably close #280 now? Are you and @m4rc1e are building variable fonts with success?

madig:
Building a VF works, as do static instances.

schriftgestalt:
The UI calls it it "Axis Locations" but the underlying data structure in the file has not changes. There are still the `weightValue`, `widthValue`, `customValue`, `customValue1` ... for the first axis, use the `weightValue`, for the second axis use the `widthValue` and so on. The UI will show the proper axis names but under the hood, nothing has changed. I couldn’t change the file structure for v2.5.

PR #418, opened 2018-09-10
Simplify glyphOrder handling

madig:
Glyphs.app handles the glyphOrder parameter like UFOs: it's management is left to the user (new glyphs don't trigger a modification of the list) and non-existant glyphs are ignored (tested in Glyphs.app 2.4.1 and 2.5.2).

Remove making it a special custom parameter.

### DIFF
--- a/Lib/glyphsLib/builder/custom_params.py
+++ b/Lib/glyphsLib/builder/custom_params.py
@@ -25,6 +25,7 @@ from .constants import (
     UFO2FT_USE_PROD_NAMES_KEY,
     CODEPAGE_RANGES,
     REVERSE_CODEPAGE_RANGES,
+    PUBLIC_PREFIX,
 )
 from .features import replace_feature
 
@@ -134,6 +135,11 @@ class GlyphsObjectProxy(object):
         for param in self._owner.customParameters:
             if param.name not in self._handled:
                 yield param
+
+    def is_font(self):
+        """Returns whether we are looking at a top-level GSFont object as
+        opposed to a master or instance."""
+        return hasattr(self._owner, "glyphs")
 
 
 class UFOProxy(object):
@@ -562,8 +568,34 @@ class OS2SelectionParamHandler(AbstractParamHandler):
 
 register(OS2SelectionParamHandler())
 
-# Do NOT use public.glyphOrder
-register(ParamHandler("glyphOrder", ufo_prefix=GLYPHS_PREFIX))
+
+class GlyphOrderParamHandler(AbstractParamHandler):
+    """Translate between Glyphs.app's glyphOrder parameter and UFO's
+    public.glyphOrder.
+
+    See the GlyphOrderTest class for a thorough explanation.
+    """
+
+    def to_glyphs(self, glyphs, ufo):
+        if glyphs.is_font():
+            glyphs_glyphOrder = ufo.get_lib_value(GLYPHS_PREFIX + "glyphOrder")
+            if isinstance(glyphs_glyphOrder, list):
+                glyphs.set_custom_value("glyphOrder", glyphs_glyphOrder)
+            elif glyphs_glyphOrder is None:
+                ufo_glyphOrder = ufo.get_lib_value(PUBLIC_PREFIX + "glyphOrder")
+                if ufo_glyphOrder:
+                    glyphs.set_custom_value("glyphOrder", ufo_glyphOrder)
+
+    def to_ufo(self, glyphs, ufo):
+        if glyphs.is_font():
+            glyphs_glyphOrder = glyphs.get_custom_value("glyphOrder")
+            if glyphs_glyphOrder:
+                ufo.set_lib_value(PUBLIC_PREFIX + "glyphOrder", glyphs_glyphOrder)
+            else:
+                ufo.set_lib_value(GLYPHS_PREFIX + "glyphOrder", False)
+
+
+register(GlyphOrderParamHandler())
 
 
 # See https://github.com/googlei18n/glyphsLib/issues/214
--- a/tests/builder/builder_test.py
+++ b/tests/builder/builder_test.py
@@ -1144,20 +1144,61 @@ class SkipDanglingAndNamelessLayers(unittest.TestCase):
 class GlyphOrderTest(unittest.TestCase):
     """Check that the glyphOrder data is persisted correctly in all directions.
 
-    In Glyphs, there are two pieces of information to save:
-     * The custom parameter 'glyphOrder', if provided
-     * The actual order of glyphs in the font.
+    Problem: Glyphs.app (tested 2.4.1 and 2.5.2) does not import and export a
+    UFO's public.glyphOrder lib key verbatim. To prevent the glyph order from
+    changing when exporting to UFO from within Glyphs.app, working on the UFO
+    and re-opening it in Glyphs.app, it will:
 
-    We need both because:
-     * There can be glyphs in the font that are not in the glyphOrder
-     * Those extraneous glyphs are ordered automatically by Glyphs.app but
-       we don't know how, so if we want to reproduce the original ordering
-       with glyphsLib, we must store it.
-       FIXME: ask Georg how/whether we can order like Glyphs.app in glyphsLib
+    1. if a glyphOrder parameter is present, write it verbatim to the
+       `com.schriftgestaltung.glyphOrder` as a list and write a
+       cleaned-up version of it (e.g. only existing glyphs plus ones not covered
+       in it but present in the font) to `public.glyphOrder`.
+    2. if a glyphOrder parameter is not present, write the implicit order of the
+       font to `public.glyphOrder` and write the
+       `com.schriftgestaltung.glyphOrder` lib key with the value `False`.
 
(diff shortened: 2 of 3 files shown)


madig:
Todo: determine if we should always place a glyphOrder parameter in the .glyphs files (if so, adjust GlyphsUnitTestSans.glyphs). I'm leaning towards yes to keep diffs small... at least diffs after the first Glyphs->UFO->Glyphs trip. 🤔

madig:
I think I'm simply going to always put a glyphOrder parameter in the .glyphs files to keep it simple. Can change if it turns out to be a bad idea later I guess.

anthrotype:
> I think I'm simply going to always put a glyphOrder parameter in the .glyphs files to keep it simple.

hm, why? What if the UFOs do not have any `public.glyphOrder`?

madig:
When would that happen?

anthrotype:
It’s an optional attribute, the UFOs may come from anywhere, no?

madig:
Hm!!! Gonna add a test for that.. as soon as I figure out why my roundtrip test fails.

schriftgestalt:
> I think I'm simply going to always put a glyphOrder parameter in the .glyphs files to keep it simple. Can change if it turns out to be a bad idea later I guess.

That is not what Glyphs is doing. 
When writing a ufo, Glyphs always writes a `public.glyphOrder`. If the Glyphs file has a `glyphOrder` parameter, it puts it into `com.schriftgestaltung.glyphOrder`. If not it still writes `com.schriftgestaltung.glyphOrder` but with a value of `False`.

On import, it checks for the `com.schriftgestaltung.glyphOrder` key and sets the `glyphOrder`parameter. If it doesn’t find it, it puts `public.glyphOrder` into the `glyphOrder parameter.

madig:
Hm! Here's a reference .glyphs file and UFO dump of 2.5.2: [Archive.zip](https://github.com/googlei18n/glyphsLib/files/2367583/Archive.zip)

madig:
@schriftgestalt this is actually unwanted behavior for us. We frequently use Fontlab 5 in our workflow and part of that is making encoding files to sort and line glyphs up in a certain way, using placeholder glyphs that don't actually exist in the font as empty slots/"spacers" of sorts (the glyph12345 thingies in the example fonts above). They get stored as part of `public.glyphOrder` and round-trip back and forth. By writing the true glyph order with the spacers somewhere else, our designers frequently have to wade through a lot of Git noise and more often than we'd like just commit it wholesale, leading to problems down the line.

schriftgestalt:
For round tripping I need to know it the UFO was opened in Glyphs before and thous it should use the default glyph ordering (not reading the glyphOrder from the UFO). Because if you safe a UFO from Glyphs to do a few things elsewhere, you don’t expect to get a glyphOrder parameter when you open the file in Glyphs, later. 

And I need a place to store the actual ‘glyphOrder’ parameter as it can contain more or less entries than you have glyphs in the file. 

If you need like to change the sorting outside of Glyphs, either update both lists or just remove the Glyphs specific glyphOrder parameter from the UFO. That way, the conversion to .glyphs (either by Glyphs or glyphsLib) should put the public glyphOrder into a glyphOrder parameter. So there shouldn’t be any fit noise.

anthrotype:
> the default glyph ordering

That is, the glyph order when there is no `glyphOrder` custom parameter.
Where is that specified? Shouldn't glyphsLib also use this?

madig:
This sounds like Glyphs is doing more work than necessary here... why not make "Ignore UFO glyph order" an default-true option like "Disable automatic alignment", i.e. a boolean lib key? That way you don't need to copy lists around and we can go for  less git noise by default?

madig:
I'd think that since glyphsLib is explicitly not using Glyphs' import/export mechanisms, we don't need to do the same thing. How about this:

1. Glyphs -> UFO: If a .glyphs file has... 
    1. a `glyphOrder` parameter set, it gets copied to `public.glyphOrder`
    2. no `glyphOrder` parameter set, the implicit order of glyphs becomes `public.glyphOrder`
2. UFO -> Glyphs: If a UFO has...
    1. a `com.schriftgestaltung.glyphOrder` lib key, it gets preferentially copied to the `glyphOrder` parameter
    2. just a `public.glyphOrder` lib key, it gets copied to the `glyphOrder` parameter
    3. no `public.glyphOrder` lib key, do nothing and let Glyphs order things implicitly

madig:
@anthrotype proposed something else:
1. Glyphs -> UFO: If a .glyphs file has... 
    1. a `glyphOrder` parameter set, it gets copied to `public.glyphOrder`, `com.schriftgestaltung.glyphOrder` is not written.
    2. no `glyphOrder` parameter set, the implicit order of glyphs becomes `public.glyphOrder` and `com.schriftgestaltung.glyphOrder` is set to `False`.
2. UFO -> Glyphs: If a UFO has...
    1. a `public.glyphOrder` and `instanceof(com.schriftgestaltung.glyphOrder, list)` lib key, `com.schriftgestaltung.glyphOrder` gets copied to the `glyphOrder` parameter.
    2. a `public.glyphOrder` and `com.schriftgestaltung.glyphOrder is False` lib key, do not set a `glyphOrder` parameter.
    3. a `public.glyphOrder` and no `com.schriftgestaltung.glyphOrder` key, copy `public.glyphOrder` to the `glyphOrder` parameter.
    4. no `public.glyphOrder` lib key, do not set the `glyphOrder` parameter.

Side-note: if `com.schriftgestaltung.glyphOrder is False`, `public.glyphOrder` must not contain non-existant glyphs as Glyphs won't sanitize it and display everything in a hacked-up way.

anthrotype:
This sounds good to me, thanks Nikolaus!

madig:
@schriftgestalt Uh oh. I turned UbuntuMono R and B UFOs into a .glyphs file with this code and Glyphs.app crashes on trying to open it. Seems to be related to the hex data representing the data/ folder through: 
[UbuntuMono.glyphs.zip](https://github.com/googlei18n/glyphsLib/files/2371725/UbuntuMono.glyphs.zip)

schriftgestalt:
> a glyphOrder parameter set, it gets copied to the public.glyphOrder lib key, the com.schriftgestaltung.glyphOrder lib key is not written.

the glyphOrder parameter does not necessarily reflect the actually glyphs in the font. Thats why I always write it to `com.schriftgestaltung.glyphOrder` (and use false if not present) and `public.glyphOrder` is always implied from the glyphs.

schriftgestalt:
> @schriftgestalt Uh oh. I turned UbuntuMono R and B UFOs into a .glyphs file with this code and Glyphs.app crashes on trying to open it. Seems to be related to the hex data representing the data/ folder through. 

Yes. Glyphs didn't handle the 1MB of hex data. I fixed it.

schriftgestalt:
I compared the file UbuntuMono.glyphs with a the file written by Glyphs. There was one difference. glyphs writes multiple unicodes as a string (`unicode = "0000,0008,001D";`) to pretend that it is valid plist. I think had it without quotes and someone complained.

anthrotype:
> glyphs writes multiple unicodes as a string (unicode = "0000,0008,001D";) to pretend that it is valid plist. I think had it without quotes and someone complained.
 
I for one was not super happy about that, I’m glad to hear that you reverted the change. We can change GlyphsLib back to outputting quotes around strings that contain extra characters (the comma in this case), however we will still need to be able to parse .glyphs filed generated by earlier Glyphs.app versions.

schriftgestalt:
the code that didn’t write the quote was in the repo for a few days. But the ‘unicode’ value has to be interpreted as a string anyhow, so parsing shouldn’t be a problem.

madig:
> the glyphOrder parameter does not necessarily reflect the actually glyphs in the font. 

I guess I can see where you are coming from with this, but I must disagree with the conclusion. The UFO spec explicitly allows missing or extra glyphs in there and we use non-existing placeholder glyphs in our workflow. I think Glyphs should not change the meaning of that key. If your goal is to prevent a different ordering when ex- and reimporting UFOs, maybe reinterpret `com.schriftgestaltung.glyphOrder == False` to mean "Ignore public.glyphOrder for ordering purposes." but always faithfully round-trip `public.glyphOrder`. You can decide whether to store an explicit glyphOrder parameter based on whether `public.glyphOrder` matches the implicit order of the .glyphs file.

schriftgestalt:
> You can decide whether to store an explicit glyphOrder parameter based on whether public.glyphOrder matches the implicit order of the .glyphs file. Yes: delete com.schriftgestaltung.glyphOrder internally and use default order. No: write glyphOrder parameter and use it for ordering.

That is what the key means. At least it has the same result.

madig:
Please test Glyphs with com.schriftgestaltung.glyphOrder == False and non-existing glyph names in public.glyphOrder then, Glyphs hacks up the order and may crash on save.

schriftgestalt:
do you have a test file that causes the hiccup?

madig:
Take any Ubuntu UFO and insert com.schriftgestaltung.glyphOrder == False  in its lib.plist.

madig:
Please also test the case where a public.glyphOrder does not exist, which may also crash Glyphs. The key is optional according to the spec.

PR #421, opened 2018-09-11
Cache pip packages and pre-commit installation

madig:
This hopefully cuts down on the awfully long setup time of pre-commit on each commit.


madig:
Yep, seems to have halved the lint time from ~3 minutes to ~1,5 minutes.

PR #425, opened 2018-09-12
UFO roundtripping: keep groups at top

madig:
See https://github.com/googlei18n/glyphsLib/issues/422.


anthrotype:
Yeah, let’s aim to drop python 2 support for the 1st of January 2019.

madig:
A quick check in Glyphs with Ubuntu-R and all features set to "automatic" shows: the final feature file has table OS/2, hhea, something and name, then glyph classes, then language declarations etc. So this should be good. @anthrotype?

PR #426, opened 2018-09-13
Add test for preserving a GDEF table

madig:
Don't want to throw the code away.

Any suggestions for more tests?


belluzj:
Could you add a comment in the test saying that you expect the GDEF to be preserved regardless of minimize_ufo_diffs?

belluzj:
and a link to your GDEF issue

PR #428, opened 2018-09-14
Fix parsing backslash escaped glyph names in features

anthrotype:
Fixes https://github.com/googlei18n/fontmake/issues/445

### DIFF
--- a/tests/writer_test.py
+++ b/tests/writer_test.py
@@ -1025,7 +1025,7 @@ rememberToDownloadARealRemindersApp = 1;}"',
             '"130 431 LINE {\\"\\012key\\\\"\';\\012\\012\\012\\" '
             '= \\"\\\\"\'value\\012several lines\\012;\\012\\"'
             ';\\n\\";\\" = \\";\\012\\";\\n'
-            'escapeception = \\"\\\\\\\\"\\\\\'\\\\n\\\\\\\\n\\";}"'
+            'escapeception = \\"\\\\\\\\"\\\\\'\\\\\\n\\\\\\\\\\n\\";}"'
         )
         self.assertWritesValue(node, expected_output)
         # Check that we can read the userData back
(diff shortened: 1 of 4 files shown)


tests/writer_test.py
```diff
@@ -1025,7 +1025,7 @@ def test_write_node(self):
             '"130 431 LINE {\\"\\012key\\\\"\';\\012\\012\\012\\" '
             '= \\"\\\\"\'value\\012several lines\\012;\\012\\"'
             ';\\n\\";\\" = \\";\\012\\";\\n'
-            'escapeception = \\"\\\\\\\\"\\\\\'\\\\n\\\\\\\\n\\";}"'
+            'escapeception = \\"\\\\\\\\"\\\\\'\\\\\\n\\\\\\\\\\n\\";}"'
```
belluzj: Can you check that your new value is indeed the output of the latest Glyphs version? I remember getting surprising results... So can you run the code above to create a point and assign userData in Glyphs and save the file, to check which output you get.

anthrotype: I don’t have a Mac at hand, I was hoping one of you guys would try and check this.
I spent the whole day scratching my head around this and this is the only way that i could make it round trip with arbitrary level of nesting and escaping.

madig: I get this string written to disk from glyphs:
```
nodes = (
"404 387 LINE {\"\012key\\"';\012\012\012\" = \"\\"'value\012several lines\012;\012\";\n\";\" = \";\012\";\nescapeception = \"\\\\"\\'\\n\\\\n\";}",
"209 390 LINE",
"299 619 LINE"
);
```
Now to test for equality...

madig: Glyph writes the string that was before your change :(

madig: Opening the file above again in Glyphs makes a slash disappear:
```
{
    "\nkey\"';\n\n\n" = "\"'value\nseveral lines\n;\n";
    ";" = ";\n";
    escapeception = "\\\"\\'\n\\\n";
}
```
@schriftgestalt :)

schriftgestalt:
I checked this. When I run your python script, the .glyphs file contains some `\n` to the end of the node string. That is "fixed", when saving and reopening the file. But all slashes stay the same.

PR #430, opened 2018-09-14
Make sure fontMasterID lib keys in UFOs are unique

madig:
Fixes https://github.com/googlei18n/glyphsLib/issues/400.

### DIFF
--- a/Lib/glyphsLib/builder/masters.py
+++ b/Lib/glyphsLib/builder/masters.py
@@ -90,6 +90,15 @@ def to_glyphs_master_attributes(self, source, master):
         # GSFontMaster has a random id by default
         pass
 
+    if master.id.lower() in (m.id.lower() for m in self.font.masters):
+        raise ValueError(
+            "{} contains a '{}' lib key with the duplicate value '{}'. All given "
+            "masters must have a unique ID or data will get corrupted. Please "
+            "check for this lib key and either remove it (will be regenerated) or "
+            "change the value. If there is no key, you just witnessed something "
+            "unlikely.".format(ufo.path, MASTER_ID_LIB_KEY, master.id)
+        )
+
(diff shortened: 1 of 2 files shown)


Lib/glyphsLib/builder/masters.py
```diff
@@ -90,6 +90,13 @@ def to_glyphs_master_attributes(self, source, master):
         # GSFontMaster has a random id by default
         pass
 
+    if master.id.lower() in (m.id.lower() for m in self.font.masters):
+        raise ValueError(
+            "{} contains a '{}' lib key with the duplicate value '{}'. All given masters must have a unique ID or data will get corrupted. Please check for this lib key and either remove it (will be regenerated) or change the value. If there is no key, you just witnessed something unlikely.".format(
```
anthrotype: you should split these long lines manually, black doesn't help there

madig: This is something that black should have one day, I need it constantly 😁

tests/builder/builder_test.py
```diff
@@ -949,6 +949,28 @@ def test_italic_angle(self):
...
+        master2.ascender = 0
+        master2.capHeight = 0
+        master2.descender = 0
+        master2.xHeight = 0
+        font.masters.append(master2)
+        ufos = to_ufos(font, minimize_glyphs_diffs=True)
+
+        try:
+            to_glyphs(ufos)
+        except Exception as e:
```
anthrotype: why do that? you want to hide the traceback? I'd say let it raise

madig: This is a should-not-raise test :) I copy-pasted the concept from https://stackoverflow.com/questions/4319825/python-unittest-opposite-of-assertraises.

anthrotype: Ok

PR #432, opened 2018-09-18
add --generate-GDEF option

anthrotype:
as discussed in https://github.com/googlei18n/glyphsLib/issues/427, this makes the generation of the GDEF table in the exported UFO features independent of the other option `--no-preserve-glyphsapp-metadata` (which corresponds to `minimize_glyphs_diffs` keyword argument of UFOBuilder).
A new argument `generate_GDEF` is added to the functions that build UFOs (we have a lot of them!), and a corresponding command-line option `--generate-GDEF` is also added to the glyphs2ufo script.

The default value of this new option is different between the console script (which defaults to False) and the programmatic interface (which defaults to True). This is in the same vein as the other boolean options like propagate_anchors and minimize_glyphs_diffs that similarly have opposite defaults between the API and the CLI.

This is for historical reason and to keep backward compatibility with tools (e.g. fontmake) that expect the ufo-building functions in glyphsLib to produce ready-to-export UFOs (i.e. without extra editor's metadata, with anchors propagated to composites and a GDEF table generated in the features).
On the other hand, the two companion console scripts `glyphs2ufo` and `ufo2glyphs` are designed to be used primarily for going back and forth between UFO and glyphs sources during the font development process, hence their defaults reflect that use case. 

So doing `fontmake -g MyFont.glyphs -o ufo` is not the same as `glyphs2ufo MyFont.glyphs`. In the latter, the anchors are not propagated to composites, and there is extra data that is needed to reduce the diffs when roundtripping back to Glyphs format, and the features do not contain a GDEF table.
 
To obtain the same output as fontmake when using glyphs2ufo script one would need to additionally pass `--propagate-anchors --generate-GDEF` (and optionally `--no-propagate-glyphsapp-metadata` although the latter should be innocuous for the sake of building an OTF from the generated UFOs).

related fontmake issue https://github.com/googlei18n/fontmake/issues/457

### DIFF
--- a/tests/builder/features_test.py
+++ b/tests/builder/features_test.py
@@ -23,6 +23,8 @@ import defcon
 
 from glyphsLib import to_glyphs, to_ufos, classes
 
+import pytest
+
 
 def make_font(features):
     ufo = defcon.Font()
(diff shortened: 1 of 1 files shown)


madig:
lgtm. If we add more options, we should probably replace all the single options with an `options` dict or something.

PR #436, opened 2018-09-30
remove unicodes list hack; write unicode with quotes if looks like decimal numbers

anthrotype:
Glyphs.app now always writes multiple unicodes within quotes (as the string contains a comma which is not a valid unquoted character in openstep plist, being reserved as the array item separator).
Therefore GlyphsLib doesn't need to keep supporting this invalid syntax. Multiple, comma-separated unicodes list are only supported when they are wrapped in quotes.

I also propose to change the writer so that it writes _single_ unicode strings within quotes when it's ambiguous whether they are an hexadecimal string or a decimal number (e.g. 0041).
If the single unicode string does not start with a digit, then the quotes can be omitted, as it's clear it cannot be a decimal number.

This would allow to write a parser that can look at the first character of an unquoted string, and quickly decide whether it may be a (integer or float) number, or a simple string.

### DIFF
--- a/Lib/glyphsLib/parser.py
+++ b/Lib/glyphsLib/parser.py
@@ -30,7 +30,6 @@ class Parser(object):
     """Parses Python dictionaries from Glyphs source files."""
 
     value_re = r'(".*?(?<!\\)"|[-_./$A-Za-z0-9]+)'
-    unicode_list_re = re.compile(r"\s*([0-9a-fA-F]+(,[0-9a-fA-F]+)*)")
     start_dict_re = re.compile(r"\s*{")
     end_dict_re = re.compile(r"\s*}")
     dict_delim_re = re.compile(r"\s*;")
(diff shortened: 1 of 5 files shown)


schriftgestalt:
> I also propose to change the writer so that it writes single unicode strings within quotes when it's ambiguous whether they are an hexadecimal string or a decimal number (e.g. 0041).

That would break compatibility with Glyphs??

anthrotype:
no, it won't. I tried, and Glyphs.app seems to not care about the extra quotes around single unicode strings.

schriftgestalt:
Yes. Glyphs can read it fine but it will break round tripping.

anthrotype:
> Glyphs can read it fine but it will break round tripping.

ok I take that back. We'll keep writing single unicode strings without the quotes until Glyphs.app does.

PR #441, opened 2018-10-12
set custom 'append' mode for ufo2ft KernFeatureWriter 

anthrotype:
ufo2ft users did not like that I changed the default mode for the KernFeatureWriter to match Glyphs.app's behavior. It used to be "skip" (only add feature if not already present), but in ufo2ft v2 I changed it to "append" (add new kern lookups to existing kern feature). The idea was to support this use case:
https://glyphsapp.com/tutorials/contextual-kerning

I decided to revert the kernFeautureWriter default back to "skip". But we need to a way from glyphsLib to signal that, when building fonts from .glyphs sources with fontmake, we still wish to have the ufo2ft-generated kern feature appended to manually written kern lookups.

The solution is to store this custom glyphsLib-specific settings in the designspace lib, and have fontmake apply them when calling ufo2ft.

So, in sum, no changes for fontmake users building from .glyphs sources. But a small change to the glyphsLib-exported designspace, only when GSFont has a manually written "kern" feature.

### DIFF
--- a/Lib/glyphsLib/builder/constants.py
+++ b/Lib/glyphsLib/builder/constants.py
@@ -81,3 +81,21 @@ CODEPAGE_RANGES = {
 }
 
 REVERSE_CODEPAGE_RANGES = {value: key for key, value in CODEPAGE_RANGES.items()}
+
+UFO2FT_FEATURE_WRITERS_KEY = "com.github.googlei18n.ufo2ft.featureWriters"
+
+# ufo2ft KernFeatureWriter default to "skip" mode (i.e. do not write features
+# if they are already present), while Glyphs.app always adds its automatic
+# kerning to any user written kern lookups. So we need to pass custom "append"
+# mode for the ufo2ft KernFeatureWriter whenever the GSFont contain a non-automatic
+# 'kern' feature.
+# See https://glyphsapp.com/tutorials/contextual-kerning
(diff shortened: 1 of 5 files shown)


schriftgestalt:
If you have a custom kern or mark feature, Glyphs will use it as is unless the code contains a `# Automatic Code End` or `# Automatic Code Start` comment. 

so for example:
```
# Automatic Code End

pos a b c' <100 0 100 0>;
```

will be appended to the generated kern feature. It is not just appended, more inserted in the right place depending of the script. The above code will put it below the default latin kerning. if script tags are present it is sorted behind the automatic code for each script. That is because makeOTF allows each script only once per file.

anthrotype:
@schriftgestalt thanks Georg, I'll see if I can add that to ufo2ft as well, as it seems like a useful feature to have regardless of font source format.

I have a question though: when was this behaviour introduced in Glyphs.app? If I grep for those special comments in the noto-source repository I cannot find them among them, and I'm certain that there are at least a few noto fonts (like Sans Arabic or Sans Tibetan) that I remember define a some manual kern lookups that need be extended with the autogenerated kern feature, and they do not seem to have those special comments.

Does current Glyphs.app _not_ generate automatic kern for these?

schriftgestalt:
I just checked again and for `kern` and `dist` it doesn’t use the comments. Only for the various `mark` features.

anthrotype:
thanks for checking!

PR #444, opened 2018-10-24
make sure the master.id is unique

schriftgestalt:
Glyphs does the same

### DIFF
--- a/Lib/glyphsLib/classes.py
+++ b/Lib/glyphsLib/classes.py
@@ -534,6 +534,8 @@ class FontFontMasterProxy(Proxy):
 
     def append(self, FontMaster):
         FontMaster.font = self._owner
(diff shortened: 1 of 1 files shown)


belluzj:
Thanks @schriftgestalt!

Lib/glyphsLib/classes.py
```diff
@@ -534,6 +534,8 @@ def values(self):
 
     def append(self, FontMaster):
         FontMaster.font = self._owner
+        if FontMaster.id and len(FontMaster.id) > 0 and self[FontMaster.id]:
```
belluzj: Why do you do an extra check of the `len`? It looks like `if FontMaster.id` will already be False for the empty string?

```suggestion
        if FontMaster.id and self[FontMaster.id]:
```

PR #450, opened 2018-10-24
some performances improvements

schriftgestalt:


madig:
Quick timing of `Point()` instantiation before:
```
In [6]: %timeit Point(100, 200)
1.2 µs ± 8.43 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)

In [7]: %timeit Point(100, 200)
1.18 µs ± 3.43 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)

In [8]: %timeit Point(100, 200)
1.15 µs ± 6.37 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
```
and after:
```
In [11]: %timeit Point(x=100, y=200)
443 ns ± 4.07 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)

In [12]: %timeit Point(x=100, y=200)
442 ns ± 2.01 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)

In [13]: %timeit Point(x=100, y=200)
442 ns ± 2.18 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
```
I suppose this is because it skips the `assert` and `super()`. That class hierarchy looks pretty bonkers...

schriftgestalt:
the value assignment is used when the Point is initialised from a string `Point("{100,200}")`.

schriftgestalt:
Your code looks better.

Lib/glyphsLib/classes.py
```diff
@@ -2117,14 +2119,13 @@ def shouldWriteValueForKey(self, key):
 class GSAnchor(GSBase):
     _classesForName = {"name": unicode, "position": Point}
     _parent = None
-    _defaultsForName = {"position": Point(0, 0)}
+    _defaultsForName = {"position": Point(x=0, y=0)}
 
     def __init__(self, name=None, position=None):
-        super(GSAnchor, self).__init__()
```
madig: Why are you removing the super() calls?

schriftgestalt: The super() calls are only needed to pre-populate the instance variables. But maybe `__slots__` is a better mechanism.

madig:
Hm. I'm not comfortable removing those calls as that can lead to
brittleness.

madig:
I imported the bulk of the changes here, I'll leave the rest for another time and close this PR. I'd need some conflict resolving anyway.

PR #451, opened 2018-10-26
Use the OS/2 width class to map the wdth axis

belluzj:
Possible improvement on the situation in https://github.com/googlei18n/fontmake/issues/470

@anthrotype

### DIFF
--- a/Lib/glyphsLib/builder/axes.py
+++ b/Lib/glyphsLib/builder/axes.py
@@ -311,11 +311,14 @@ class AxisDefinition(object):
         """Get the user location of a Glyphs master or instance.
         Masters in Glyphs can have a user location in the "Axis Location"
         custom parameter.
+
         The user location is what the user sees on the slider in his
         variable-font-enabled UI. For weight it is a value between 0 and 1000,
         400 being Regular and 700 Bold.
-        For width it's the same as the design location, a percentage of
-        extension with respect to the normal width.
+
+        For width it's a percentage of extension with respect to the normal
+        width, 100 being normal, 200 Ultra-expanded = twice as wide.
+        It may or may not match the design location.
         """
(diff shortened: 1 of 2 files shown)


anthrotype:
This change shold have been discussed a bit more, or at least better documented. I only realized the effect of this now.
Basically all the noto-source variable fonts that have wdth axis have their fvar table modified by this (plus they get an additional avar SegmentMap for wdth axi, where previously avar only contained maps for wght).
Previously the wdth axis minimum in fvar was 70 (default 100 and maximum 100); now the minimum is 62.5, which is the percentage associated to "Extra Condensed" OS/2.usWidthClass (2).

I'm still not sure if I should change (and how?) the noto-source files to have them produce the same fvar/avar tables as before, or if the change is innocuous and for the good.

anthrotype:
but i looks to me like a breaking change, since anybody who was relying on the old behavior, namely that glyphsLib would _not_ generate axis maps for `wdth` axis (i.e. assuming that master/instance locations for wdth axis were already in the same coordinate-space as in the fvar *user*-space coordinates), would now get possibly unwanted changes in the generated fvar (and avar) when building their font with glyphsLib >= 3.3.0.

belluzj:
I implemented this from your comments in https://github.com/googlefonts/fontmake/issues/470#issuecomment-433393382

I think that your original comment still makes sense, and that the design locations and user locations should be able to be chosen independently.

anthrotype:
> the design locations and user locations should be able to be chosen independently

yes, but in a way that doesn't break existing fonts

anthrotype:
reverted with https://github.com/googlefonts/glyphsLib/commit/746ad757c7791a891e9daa2f843c5756d0956c1b

let's continue discussing on this in #483

PR #453, opened 2018-10-31
Set output directory to "." if script launched in current dir and nothing is specified

madig:
Prevents crash in https://github.com/googlei18n/glyphsLib/blob/master/Lib/glyphsLib/__init__.py#L112 when glyphs2ufo is launched from same directory as .glyphs file.


anthrotype:
ah yes, `os.path.dirname("Font.glyphs")` is an empty string. lgtm

PR #460, opened 2018-11-13
this value is assumed to be float or int later

schriftgestalt:


madig:
What do you mean, how is the user location to be handled?

schriftgestalt:
It might be not needed any more, may be related to my too old fontmake.

PR #462, opened 2018-11-13
Uniquify master IDs automatically

madig:
Instead of crashing when finding duplicate master IDs, simply generate new ones.

Closes https://github.com/googlei18n/glyphsLib/pull/444.


schriftgestalt:
The MasterIDs are case sensitive. At least in Glyphs.

madig:
OK then 😬 Then I'll keep just the uniquifying part.

PR #465, opened 2018-11-15
Minor performance improvement for Point

madig:
* Simplify Point `__init__` to speed up instantiation
* Use `__slots__` to shave off a few ns.

Supersedes https://github.com/googlei18n/glyphsLib/pull/450.

### DIFF
--- a/Lib/glyphsLib/classes.py
+++ b/Lib/glyphsLib/classes.py
@@ -2381,7 +2381,7 @@ class GSAnnotation(GSBase):
     }
     _defaultsForName = {
         "angle": 0.0,
-        "position": Point(),
+        "position": Point(0, 0),
         "text": None,
         "type": 0,
         "width": 100.0,
--- a/Lib/glyphsLib/types.py
+++ b/Lib/glyphsLib/types.py
@@ -119,17 +119,26 @@ def Vector(dim):
 class Point(Vector(2)):
     """Read/write a vector in curly braces."""
 
-    def __init__(self, value=None, value2=None, rect=None):
-        if value is not None and value2 is not None:
-            value = [value, value2]
-        assert (
-            value is None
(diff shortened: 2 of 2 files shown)


Lib/glyphsLib/types.py
```diff
@@ -120,17 +120,18 @@ class Point(Vector(2)):
...
 
+        # Invoked like Point(100, 200) or Point(value=100, value2=200).
+        if value is not None and value2 is not None:
+            self.value = [value, value2]
+        # Invoked like Point("{800, 10}")
+        elif isinstance(value, (str, unicode)):
+            self.value = self.fromString(value)
+        # 🤷
+        else:
+            self.value = value
```
anthrotype: else TypeError please

madig:
New benchmark:

Without `__slots__`:
```
In [2]: %timeit Point(100, 200)
364 ns ± 2.03 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)

In [3]: %timeit Point(100, 200)
367 ns ± 6.27 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
```

With `__slots__`:
```
In [3]: %timeit Point(100, 200)
326 ns ± 0.59 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)

In [4]: %timeit Point(100, 200)
329 ns ± 2.7 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
```

Lib/glyphsLib/classes.py
```diff
@@ -2381,7 +2381,7 @@ class GSAnnotation(GSBase):
     }
     _defaultsForName = {
         "angle": 0.0,
-        "position": Point(),
+        "position": Point(0, 0),
```
anthrotype: much better, thanks

PR #478, opened 2018-12-21
Implement generation of brace layers in .designspace files

madig:
This is supposed to work one-way for now (i.e. store brace layers as seperate sources in a .designspace file). Conversion of layer sources not following the brace layer convention when going to .glyphs is something for when the use case arises.

### DIFF
--- a/Lib/glyphsLib/builder/builders.py
+++ b/Lib/glyphsLib/builder/builders.py
@@ -342,9 +342,12 @@ class GlyphsBuilder(_LoggerMixin):
         # Sort UFOS in the original order from the Glyphs file
         sorted_sources = self.to_glyphs_ordered_masters()
 
+        # Convert all full source UFOs to Glyphs masters. Sources with layer names
+        # are assumed to be sparse or "brace" layers and are ignored because Glyphs
+        # considers them to be special layers and will handle them itself.
         self._font = self.glyphs_module.GSFont()
         self._sources = OrderedDict()  # Same as in UFOBuilder
-        for index, source in enumerate(sorted_sources):
+        for index, source in enumerate(s for s in sorted_sources if not s.layerName):
             master = self.glyphs_module.GSFontMaster()
             self.to_glyphs_font_attributes(source, master, is_initial=(index == 0))
             self.to_glyphs_master_attributes(source, master)
@@ -389,7 +392,9 @@ class GlyphsBuilder(_LoggerMixin):
         """
         # TODO: (jany) really make a copy to avoid modifying the original object
         copy = designspace
-        for source in copy.sources:
+        # Load only full UFO masters, sparse or "brace" layer sources are assumed to point
+        # to existing layers within one of the full masters.
+        for source in (s for s in copy.sources if not s.layerName):
             if not hasattr(source, "font") or source.font is None:
                 if source.path:
                     # FIXME: (jany) consider not changing the caller's objects
--- a/Lib/glyphsLib/builder/builders.py
+++ b/Lib/glyphsLib/builder/builders.py
@@ -392,7 +392,7 @@ class GlyphsBuilder(_LoggerMixin):
         """
         # TODO: (jany) really make a copy to avoid modifying the original object
         copy = designspace
-        # Load only full UFO masters, sparse or "brace" layer sources are assumed 
+        # Load only full UFO masters, sparse or "brace" layer sources are assumed
         # to point to existing layers within one of the full masters.
         for source in (s for s in copy.sources if not s.layerName):
             if not hasattr(source, "font") or source.font is None:
--- a/Lib/glyphsLib/builder/builders.py
+++ b/Lib/glyphsLib/builder/builders.py
@@ -392,8 +392,8 @@ class GlyphsBuilder(_LoggerMixin):
         """
         # TODO: (jany) really make a copy to avoid modifying the original object
         copy = designspace
-        # Load only full UFO masters, sparse or "brace" layer sources are assumed to point
-        # to existing layers within one of the full masters.
+        # Load only full UFO masters, sparse or "brace" layer sources are assumed 
+        # to point to existing layers within one of the full masters.
         for source in (s for s in copy.sources if not s.layerName):
             if not hasattr(source, "font") or source.font is None:
                 if source.path:
--- a/Lib/glyphsLib/builder/sources.py
+++ b/Lib/glyphsLib/builder/sources.py
@@ -14,9 +14,11 @@
 
 from __future__ import print_function, division, absolute_import, unicode_literals
 
+import collections
 import logging
 import os
 
+import fontTools.designspaceLib
 from glyphsLib.util import build_ufo_path
 
 from .masters import UFO_FILENAME_KEY
@@ -30,6 +32,7 @@ def to_designspace_sources(self):
     regular_master = get_regular_master(self.font)
     for master in self.font.masters:
         _to_designspace_source(self, master, (master is regular_master))
+    _to_designspace_source_layer(self)
 
 
 def _to_designspace_source(self, master, is_regular):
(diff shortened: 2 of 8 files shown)


davelab6:
Cc @mjlagattuta @eliheuer

schriftgestalt:
What do you mean by the last todo?

madig:
The code currently skips sources in a designspace file that have a layer attribute when converting to .glyphs and writes out all brace layers when going to .designspace.

If the user starts with a .designspace with source layers that aren't named like Glyphs expects brace layers to be named, they won't be written back when converting to .designspace again.

madig:
True, and maybe I'll do that, but I'd have to remember to remove it again
and I was lazy in this first iteration :)

tests/test_helpers.py
```diff
@@ -166,7 +166,8 @@ def write_designspace_and_UFOs(designspace, path):
         ufo_path = os.path.join(os.path.dirname(path), basename)
         source.filename = basename
         source.path = ufo_path
-        source.font.save(ufo_path, formatVersion=3)
+        if not (hasattr(source, "layerName") and source.layerName is not None):
```
anthrotype: I suggest we bump minimum required fonttools and assume layerName attribute will be there (None by default).
I think it was added with 3.26

Lib/glyphsLib/__init__.py
```diff
@@ -132,14 +132,17 @@ def build_masters(
         if create_background_layers:
             ufo_create_background_layer_for_all_glyphs(source.font)
 
-        ufo_path = os.path.join(master_dir, source.filename)
-        clean_ufo(ufo_path)
-        source.font.save(ufo_path)
+        # Save UFO to disk in a clean way, unless we are looking at an intermediate
+        # layer, which we will hit on some other iteration.
+        if not source.layerName:
```
anthrotype: elsewhere you first check if `hasattr(source, "layerName")`; let's make it consistent (cf https://github.com/googlei18n/glyphsLib/pull/478/files#r245102549)

Lib/glyphsLib/builder/builders.py
```diff
@@ -387,8 +387,14 @@ def _valid_designspace(self, designspace):
         """Make sure that the user-provided designspace has loaded fonts and
         that names are the same as those from the UFOs.
         """
-        # TODO: (jany) really make a copy to avoid modifying the original object
-        copy = designspace
+        import copy
+
+        copy = copy.deepcopy(designspace)
```
anthrotype: if the sources have a 'font' attached with a defcon.Font, this deepcopy will probably fail
https://github.com/typesupply/defcon/issues/210

anthrotype: also not a pretty choice for a local variable name that masks the just imported `copy` module

Lib/glyphsLib/builder/sources.py
```diff
@@ -80,6 +83,59 @@ def _to_designspace_source(self, master, is_regular):
...
 
+def _to_designspace_source_layer(self):
+    brace_layers = collections.defaultdict(list)
+    layer_to_glyph = collections.defaultdict(list)
+    for glyph in self.font.glyphs:
+        for layer in glyph.layers:
+            if (
+                "{" in layer.name
+                and "}" in layer.name
+                and ".background" not in layer.name
```
anthrotype: what's this ".background" magic string?

madig: As you found, brace layers may get background layers that I have to not write out to a Designspace file.

https://github.com/googlei18n/glyphsLib/blob/fba6344b4830ee527c6a52ef2e29330e5cc2abdb/Lib/glyphsLib/builder/layers.py#L44
https://github.com/googlei18n/glyphsLib/blob/fba6344b4830ee527c6a52ef2e29330e5cc2abdb/Lib/glyphsLib/builder/layers.py#L69-L72

Lib/glyphsLib/builder/builders.py
```diff
@@ -420,7 +440,18 @@ def _valid_designspace(self, designspace):
                         )
                     )
                     setattr(source, name, getattr(source.font.info, name))
-        return copy
+
+        # Rename sparse master layers to Glyphs.app brace layer name conventions.
+        for source in sources_to_rename:
+            if not hasattr(source, "font") or source.font is None:
```
anthrotype: you can assume source.font attribute is always present

Lib/glyphsLib/builder/builders.py
```diff
@@ -387,17 +387,37 @@ def _valid_designspace(self, designspace):
...
+            else:
+                sources.append(source)
+        designspace_copy.sources = sources
+
+        filename_to_ufo = {}
+        for source in designspace_copy.sources:
             if not hasattr(source, "font") or source.font is None:
                 if source.path:
-                    # FIXME: (jany) consider not changing the caller's objects
                     source.font = defcon.Font(source.path)
```
anthrotype: @madig you removed the FIXME comment, but note that you're still modifying the original SourceDescriptor objects, because you're only doing a shallow `copy.copy`. But I think it's ok here.

Lib/glyphsLib/builder/builders.py
```diff
@@ -387,17 +387,37 @@ def _valid_designspace(self, designspace):
...
-        # TODO: (jany) really make a copy to avoid modifying the original object
-        copy = designspace
-        for source in copy.sources:
+        import copy
+
+        designspace_copy = copy.copy(designspace)
+
+        sources = []
+        sources_to_rename = []
+        for source in designspace_copy.sources:
```
anthrotype: if you want to avoid modifying the input designspace object (while still avoiding calling a fallible deepcopy over defcon.Fonts), you could make a shallow copy of the SourceDescriptors here

```suggestion
        for source in designspace_copy.sources:
            source_copy = copy.copy(source)
```

Lib/glyphsLib/builder/sources.py
anthrotype: I just wonder if there's a better way to detect whether a layer is a background layer instead of checking its name has a ".background" suffix. Isn't the latter just the way glyphsLib translates Glyphs background layers to UFO layers? or is this a convention coming from Glyphs.app itself? /cc @belluzj

Lib/glyphsLib/builder/sources.py
```diff
@@ -80,6 +83,59 @@ def _to_designspace_source(self, master, is_regular):
...
+        layer_coordinates = [
+            int(c)
+            for c in brace_layer_name[
+                brace_layer_name.index("{") + 1 : brace_layer_name.index("}")
+            ].split(",")
+        ]
+        if len(layer_coordinates) != len(designspace.axes):
+            logger.warning(
+                "Glyph(s) %s, brace layer '%s' defines more or less locations than "
+                "there are design axes.",
```
anthrotype: @madig I wonder if we should skip thus invalid brace layers instead?

Lib/glyphsLib/builder/sources.py
```diff
@@ -30,6 +32,7 @@ def to_designspace_sources(self):
     regular_master = get_regular_master(self.font)
     for master in self.font.masters:
         _to_designspace_source(self, master, (master is regular_master))
+    _to_designspace_source_layer(self)
```
anthrotype: So this mean we are always appending the layerName-sources at the end of the sources list in the designspace, correct?
I would expect that the original ordering of the master and non-master layers (associated with the latter) is kept instead.
E.g., a font has two masters Light and Black, and two intermediate "brace" layers for Regular and Medium; then the generated designspace would have the sources written out as: Light, Regular, Medium and Bold. However, according to the current code the order is Light, Bold, and then the brace layers Regular and Medium.

Lib/glyphsLib/builder/sources.py
```diff
@@ -80,6 +83,59 @@ def _to_designspace_source(self, master, is_regular):
...
+                layer_to_glyph[brace_layer_name],
+                brace_layer_name,
+            )
+
+        layer_coordinates_mapping = {}
+        for axis, location in zip(designspace.axes, layer_coordinates):
+            if not (axis.minimum <= location <= axis.maximum):
+                raise ValueError(
+                    "Glyph(s) {}, brace layer '{}', location {} is out of bounds for "
+                    "axis {}".format(
```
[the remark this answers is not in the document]
anthrotype: also, I think the values in the brace layer name notation (within the braces) are meant to be specified using the internal design-space coordinates, whereas the values in the designspace axes min/default/max normally use the user-space coordinates, so you may be comparing apples with pears

Lib/glyphsLib/builder/sources.py
anthrotype: actually, I checked WorkSans and it has several brace layers specifying more values than there are axes... Anyway, better keep it like that: i.e. read in as many layer coords as there are axes, warn if their number they exceed the latter's; if less than that, maybe don't warn but simply imply the default value for the axis that was omitted?

Lib/glyphsLib/builder/sources.py
anthrotype: another reason to keep the order is when going back from DS to glyphs, to be able to determine to which master a layer source is associated with (i.e. the one immediately preceding it)

Lib/glyphsLib/builder/sources.py
```diff
@@ -80,6 +83,59 @@ def _to_designspace_source(self, master, is_regular):
...
+                    )
+                )
+            layer_coordinates_mapping[axis.name] = location
+
+        for master in masters:
+            s = fontTools.designspaceLib.SourceDescriptor()
+            s.filename = master.filename
+            s.layerName = brace_layer_name
+            s.location = layer_coordinates_mapping
+            designspace.sources.append(s)
```
anthrotype: here we need to make sure that we don't add duplicated sources (with same filename, same layer name, and same location values)

Lib/glyphsLib/builder/sources.py
madig: Ok. What happens if there are more values than axes?

Lib/glyphsLib/builder/sources.py
```diff
@@ -102,38 +102,45 @@ def _to_designspace_source_layer(self):
...
-            axis.name: location
-            for axis, location in zip(designspace.axes, layer_coordinates)
-        }
+        for master_id in master_ids:
+            master = self._sources[master_id]
+            master_coordinates = brace_coordinates
+            if len(master_coordinates) < len(designspace.axes):
+                master_locations = [master.location[a.name] for a in designspace.axes]
+                master_coordinates = (
+                    brace_coordinates + master_locations[len(brace_coordinates) :]
```
anthrotype: maybe inline that list comprehension in here and get rid of the local variable `master_locations`

Lib/glyphsLib/builder/sources.py
```diff
@@ -102,38 +102,45 @@ def _to_designspace_source_layer(self):
...
+                )
+            elif len(master_coordinates) > len(designspace.axes):
+                logger.warning(
+                    "Glyph(s) %s, brace layer '%s' defines more locations than "
+                    "there are design axes.",
+                    layer_name_to_glyph_names[layer_name],
+                    layer_name,
+                )
+
+            # If we have more locations than axes, ignore the extra locations.
```
anthrotype: remove this comment

anthrotype:
looking good so far! :+1: 
let's add more tests tomorrow

Lib/glyphsLib/builder/sources.py
madig: Brace layers are now inserted after their associated master and sorted by location.

madig:
Not sure how healthy the change to `def _original_master_order` is. The latest commit makes tests pass, except for test_BraceTestFont as mentioned above.

madig:
Python 2 seems to grapple with the order somewhere. Sigh.

madig:
So, what happens during the round-trip is that glyphsLib 

1. Writes weight and widthClass to the global custom parameters, which makes no sense, and orders the "Disable last change" parameter above the "Axis Location" parameter.
2. Writes the weight and widthClass as custom parameters to the masters. Not sure `widthClass = 1` is appropriate for "Condensed", it is shown as "3" in the instances tab?
3. Messes with the instance information by adding interpolated weightClasses custom parameters, changing the widthClass assignments and deleting `linkStyle = Regular` from the bold instance...

Not entirely sure what to do about this...

PS: Glyphs assigns the axis values to the old "weight, width, custom" order, so in this case, Width is assigned to weightValue and Weight is assigned to widthValue, also in the instances.

anthrotype:
@madig all those issues seem to me unrelated to adding support for brace layers, which is the scope of this PR. For now disable that test and file separate issues/PR for those.

PR #486, opened 2019-01-17
don't append duplicated 'Italic' to master.name with italicAngle

anthrotype:
Fixes https://github.com/googlei18n/fontmake/issues/479 ?

### DIFF
--- a/tests/classes_test.py
+++ b/tests/classes_test.py
@@ -647,6 +647,15 @@ class GSFontMasterFromFileTest(GSObjectsTestCase):
         master.italicAngle = 10.0
         self.assertEqual("Italic", master.name)
 
+        # Test that we don't get an extra "Italic" in the name of masters
+        # whose customName already contain the string "Italic"
+        master = GSFontMaster()
+        master.weight = "Regular"
+        master.width = "Regular"
+        master.customName = "Italic"
+        master.italicAngle = 10.0
+        self.assertEqual("Italic", master.name)
+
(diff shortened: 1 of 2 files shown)


anthrotype:
Before:
```
$ ls master_ufo/
WorkSans-BlackItalicItalic.ufo 
WorkSans-ItalicItalic.designspace
WorkSans-ItalicItalic.ufo
WorkSans-ThinItalicItalic.ufo
```

After:
```
$ ls master_ufo/
WorkSans-BlackItalic.ufo
WorkSans-Italic.designspace
WorkSans-Italic.ufo
WorkSans-ThinItalic.ufo
```

anthrotype:
nah.. the string we are appending is "Italic". We simply don't want to add it twice if it's already there. "ITALIC" or "italic" are typos at best.

tests/classes_test.py
```diff
@@ -647,6 +647,15 @@ def test_name(self):
...
         self.assertEqual("Italic", master.name)
 
+        # Test that we don't get an extra "Italic" in the name of masters
+        # whose customName already contain the string "Italic"
+        master = GSFontMaster()
+        master.weight = "Regular"
+        master.width = "Regular"
+        master.customName = "Italic"
+        master.italicAngle = 10.0
+        self.assertEqual("Italic", master.name)
```
anthrotype: this would have returned "Italic Italic" without this patch

PR #490, opened 2019-01-22
Add CLI option to store (or not) editor state

madig:
When working collaboratively, storing editor state is counter-productive and can lead to merge conflicts.

### DIFF
--- a/Lib/glyphsLib/__init__.py
+++ b/Lib/glyphsLib/__init__.py
@@ -90,6 +90,7 @@ def build_masters(
     normalize_ufos=False,
     create_background_layers=False,
     generate_GDEF=True,
+    store_editor_state=True,
 ):
     """Write and return UFOs from the masters and the designspace defined in a
     .glyphs file.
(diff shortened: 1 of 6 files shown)


Lib/glyphsLib/builder/font.py
```diff
@@ -51,6 +51,9 @@ def to_ufo_font_attributes(self, family_name):
     # note = font.note
     glyph_order = list(glyph.name for glyph in font.glyphs)
 
+    if not self.store_editor_state:
+        font.DisplayStrings = None
```
belluzj: Instead of changing the font in place, could this be done at the stage where the DisplayStrings are put in the lib? 

I think it's weird if you do `font = GSFont(); doc = to_designspace(font)` and after that, `font` is modified. The whole glyphs->ufo process should only read from it.

I think this should be handled in custom_params.py, like that somehow:
```py
class DisplayStringsParamHandler(MiscParamHandler):
    def __init__(self):
        super(...).__init__(glyphs_name="DisplayStrings")

    def to_ufo(self, builder, glyphs, ufo):
        if builder.store_editor_state:
            super(...).to_ufo(...)

register(DisplayStringsParamHandler())
```

but that would mean changing the signature of `ParamHandler#to_ufo` and `#to_glyphs` to take the builder as an argument and be able to read options from it

madig: Gah. I looked at that first and thought "nah!". Maybe I could just remove it from the UFO lib afterwards?

PR #492, opened 2019-01-24
WIP: Support bracket layers

madig:
This PR does the following:

- When going from .glyphs to UFO, extract bracket layers into specially named, free-standing glyphs that can be referenced from a .designspace.
- When going back to Glyphs, put the layers back and remove the free-standing glyphs.

Todo:

- [x] Preserve original layer names
- [x] Support Python 2, sigh...
- [ ] ???

### DIFF
--- a/Lib/glyphsLib/builder/builders.py
+++ b/Lib/glyphsLib/builder/builders.py
@@ -16,7 +16,6 @@ from __future__ import print_function, division, absolute_import, unicode_litera
 
 from collections import OrderedDict, defaultdict
 import logging
-import itertools
 import os
 from textwrap import dedent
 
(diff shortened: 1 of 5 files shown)


Lib/glyphsLib/builder/builders.py
```diff
@@ -247,6 +264,105 @@ def instance_data(self):
...
+            for glyph_name in glyph_name_set:
+                crossovers[glyph_name].append(location)
+
+        # Copy bracket layers to their own glyphs.
+        for location, layers in bracket_layer_map.items():
+            for layer in layers:
+                ufo_font = self._sources[
+                    layer.associatedMasterId or layer.layerId
+                ].font.layers.defaultLayer
+                ufo_glyph_name = "{layer_parent_name}.BRACKET.{location}".format(
```
belluzj: Rename layer_parent_name to glyph_name

Lib/glyphsLib/builder/builders.py
```diff
@@ -247,6 +264,105 @@ def instance_data(self):
...
+                rule.conditionSets.append(
+                    [
+                        {
+                            "name": bracket_axis.name,
+                            "minimum": crossover_min,
+                            "maximum": crossover_max,
+                        }
+                    ]
+                )
+                rule.subs.append((glyph_name, rule.name))
```
belluzj: instead of rule.name here write ufo_glyph_name so that if someone wants to change the name of the rule then it won't break the substitution

anthrotype:
@madig @belluzj thanks for working on this!

PR #495, opened 2019-01-30
build_masters: do not write source layers to disk

madig:
This would lead to empty UFOs when writing source layers.


anthrotype:
In fontmake I solved a similar problem by keying sources by their filename
https://github.com/googlei18n/fontmake/commit/5696415ff47a064d75da99d0d209688d48f95127

madig:
Let's create designspacePlusUFOAsFontLib with just one function, `save_full_UFOs()`.

PR #510, opened 2019-02-17
Roundtrip ufo2ft filters correctly

madig:
Currently, `ufo2glyphs` with a `com.github.googlei18n.ufo2ft.filters` lib key crashes. Fixing the immediate crash shows that in the Glyphs file, you end up getting multiple `PreFilter`s with one letter of the filter name each. The `include` part isn't included. Fixing that shows that the filters are doubled on roundtripping back to UFOs.

### DIFF
--- a/Lib/glyphsLib/builder/custom_params.py
+++ b/Lib/glyphsLib/builder/custom_params.py
@@ -632,18 +632,19 @@ class FilterParamHandler(AbstractParamHandler):
             glyphs.set_custom_value(glyphs_filter_key, glyphs_filter)
 
     def to_ufo(self, builder, glyphs, ufo):
-        ufo_filters = []
-        for pre_filter in glyphs.get_custom_values("PreFilter"):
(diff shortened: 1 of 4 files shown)


madig:
This looks messy. Should ufo2ft filters even be handled by `FilterParamHandler`?

madig:
Throwing ufo2ft filters and Glyphs filters in the same handler makes little sense to me. https://glyphsapp.com/tutorials/custom-parameters lists `RemoveOverlap` and `Transformations` as the only filters available in both.

madig:
Meh. Need to actually test with a filter not available in ufo2ft.

madig:
Oh wait. the ufo2ft filters are written to the masters, which is meaningless to Glyphs. Filters are apparently only supported on instances, which land in the Designspace lib key of the instance, so all should be good. @anthrotype, @schriftgestalt?

schriftgestalt:
I have not followed the filter development but you are right, filters are only supported in instances.

PR #513, opened 2019-02-28
Integers stay integers

madig:
When round-tripping, integer coordinates in .glyphs files end up being floats in the UFO.

It seems to happen when parsing .glyphs files.


madig:
Jany says that this might not be that good of an idea, because floats in a UFO might be converted to ints, so this might result in gratuitous Git diffs. Maybe it's a normalization issue instead? Thoughts?

madig:
I'll just go ahead with this.

PR #519, opened 2019-04-02
Support public.skipExportGlyphs lib key for storing export flags

madig:
https://github.com/googlei18n/glyphsLib/issues/515.

Handling this well in a oh-please-not-more-technical-debt kind of way is tricky.

There are three ways to signal that a glyph should not be exported:
1. The old `com.schriftgestaltung.Glyphs.Export` glyph lib key
2. The new UFO-level `public.skipExportGlyphs` lib key
3. The new Designspace-level `public.skipExportGlyphs` lib key

ufo2ft will look into the Designspace lib key when using the `compile*FromDS` functions and into the UFO-level lib keys with the other `compile*` functions. Writing only to the Designspace-level lib key may complicate e.g. fontmake, which uses `glyphsLib.to_designspace` to convert a .glyphs file into DS + UFOs, but uses `compile*` or `compile*FromDS` depending on what the user specifies on the CLI.

Proposed behavior:
- ufo2glyphs (`to_glyphs`):
  - When reading a Designspace file, honor only the old glyph lib key and the Designspace-level lib key
  - When reading loose UFOs, construct a Designspace and read the UFO-level `public.skipExportGlyphs` lib key, but error out when the lists don't match or some have one and others don't
- glyphs2ufo (`to_ufos`, `to_designspace`)
  - Write UFO and Designspace-level `public.skipExportGlyphs` lib keys.

### DIFF
--- a/Lib/glyphsLib/builder/builders.py
+++ b/Lib/glyphsLib/builder/builders.py
@@ -212,6 +212,10 @@ class UFOBuilder(_LoggerMixin):
         self.to_ufo_groups()
         self.to_ufo_kerning()
 
+        # Sanitize skip list and write it to both Designspace- and UFO-level lib keys.
+        # The latter is unnecessary when using e.g. the `ufo2ft.compile*FromDS`
+        # functions, but the data may take a different path. Writing it everywhere can
+        # save on surprises/logic in other software.
         skip_export_glyphs = self._designspace.lib.get("public.skipExportGlyphs")
         if skip_export_glyphs is not None:
             skip_export_glyphs = sorted(set(skip_export_glyphs))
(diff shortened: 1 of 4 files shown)


anthrotype:
sgtm

madig:
Must check interaction with the "Keep Glyphs" and "Remove Glyphs" (and others?) parameters...

madig:
We only support "Keep Glyphs" currently and it turns out it causes feature subsetting on export from Glyphs.app. `public.skipExportGlyphs` does not, currently, and as such can't be used to adapt "Keep Glyphs".

anthrotype:
I also had to revert this, because fontmake --subset option relies on the individual glyph Export flags and works slightly differently than then new `skipExportGlyphs` feature as implemented in ufo2ft. Specifically, the latter decomposes component glyphs that are marked as non-export, whereas fontmake (which uses fonttools subsetter to subset after compilation, not on the source files) keeps the components that are marked as non-export without decomposing the composite glyphs that reference them.
We need to keep supporting the old behavior and make the new one opt-in behind some flag.

madig:
> new one opt-in behind some flag

In fontmake or glyphsLib?

aaronbell:
Hi. Did this ever get implemented in another way? Or did it sort of fall by the wayside?

aaronbell:
(I should mention that I am able to work around it by manually updating the designspace file as part of the export process, but it'd be nice not to have to do that :) )

madig:
`glyphs2ufo --write-public-skip-export-glyphs ...`

aaronbell:
Aha! Wish I'd spotted that sooner 😅

Issue #518, opened 2019-04-02
Support com.schriftgestaltung.Glyphs.ComponentInfo glif lib key to round-trip anchor alignment information


madig:
Needed for e.g. assigning acutecomb to the circumflex' top_viet anchor for Vietnamese. The information is lost otherwise.

@schriftgestalt Out of curiosity, how are anchors ordered to serve as indices?

schriftgestalt:
As far as I can see the index is for the component itself and has nothing to do with the anchors. It is a hint that the components were changed outside of Glyphs and thus the component info is invalid.

I just added support for storing the anchors.
```
      <key>com.schriftgestaltung.Glyphs.ComponentInfo</key>
      <array>
        <dict>
          <key>anchor</key>
          <string>top_1</string>
          <key>index</key>
          <integer>1</integer>
          <key>name</key>
          <string>acutecomb</string>
        </dict>
      </array>
```

madig:
Huh. So how does Glyphs know which anchor a component attaches to when automatic alignment is enabled?

schriftgestalt:
Mostly only by matching a "_XX" anchor with the "XX" counterpart. The UI presents an option to pick alternate anchors when there are multiple "XX_YY" anchors (for ligatures or possibly "top_viet"). Can you send me a sample that contains the "top_viet" anchor?

madig:
https://gitlab.gnome.org/GNOME/cantarell-fonts/blob/c7d72e2cd75522c3d522c946e4380cdc18a68ef8/src/Cantarell.glyphs

schriftgestalt:
I see. By default, it will pick the `_XX` `XX` pair. But it shows all the extras anchos in as alternatives in the anchor menu. When the used selects one of it, it will use this one instead. 


<img width="324" alt="Screenshot" src="https://user-images.githubusercontent.com/174660/55459552-243cff00-55f0-11e9-90f2-9efacbd2f620.png">

madig:
Yes. What data is stored in `com.schriftgestaltung.Glyphs.ComponentInfo`, aside from the anchor attachment?

schriftgestalt:
It is only there to record the alignment state of the components.

PR #521, opened 2019-04-09
Support ComponentInfo

madig:
#518


madig:
Blarg. I'll just remove the `assertLogs`.

PR #525, opened 2019-05-11
Support UFO's `public.skipExportGlyphs`, take two

madig:
The lib key is now supported alongside the old glyph level lib key, to avoid disruption in fontmake etc. until feature subsetting from `public.skipExportGlyphs` is implemented.

Todo:
- [ ] Check that this doesn't mess with the `public.skipExportGlyphs` logic used before.
- [ ] Add switch for `glyphs2ufo` and `UFOBuilder` to output either old glyph lib key or the new `public.skipExportGlyphs`, defaulting to old behavior until we get the rest of the toolchain up to scratch (automatic feature subsetting, leaving non-exported glyphs in as components to save on binary size)

### DIFF
--- a/Lib/glyphsLib/__init__.py
+++ b/Lib/glyphsLib/__init__.py
@@ -90,7 +90,7 @@ def build_masters(
     create_background_layers=False,
     generate_GDEF=True,
     store_editor_state=True,
-    write_public_skipexportglyphs=False,
+    write_skipexportglyphs=False,
 ):
     """Write and return UFOs from the masters and the designspace defined in a
     .glyphs file.
@@ -125,7 +125,7 @@ def build_masters(
         minimize_glyphs_diffs=minimize_glyphs_diffs,
         generate_GDEF=generate_GDEF,
         store_editor_state=store_editor_state,
-        write_public_skipexportglyphs=write_public_skipexportglyphs,
+        write_skipexportglyphs=write_skipexportglyphs,
     )
 
     # Only write full masters to disk. This assumes that layer sources are always part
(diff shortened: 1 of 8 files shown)


Lib/glyphsLib/builder/glyph.py
```diff
@@ -147,18 +150,23 @@ def to_glyphs_glyph(self, ufo_glyph, ufo_layer, master):
...
         glyph.lastChange = from_loose_ufo_time(last_change)
     if ufo_glyph.markColor:
         glyph.color = _to_glyphs_color(ufo_glyph.markColor)
+
+    # The export flag can be stored in the glyph's lib key (for upgrading legacy
+    # sources) or the Designspace-level public.skipExportGlyphs lib key (canonical
+    # place to store the information). The UFO level lib key is ignored.
     if GLYPHLIB_PREFIX + "Export" in ufo_glyph.lib:
         glyph.export = ufo_glyph.lib[GLYPHLIB_PREFIX + "Export"]
+    if ufo_glyph.name in self.designspace.lib.get("public.skipExportGlyphs", []):
```
anthrotype: maybe the builder should maintain a set of the public.skipExportGlyphs to make this a constant-time lookup instead of linear search in a list.

madig: Done. Could be better coordinated with `_fake_designspace` and the other one, we're redoing work, but ugh.

Lib/glyphsLib/__init__.py
```diff
@@ -90,6 +90,7 @@ def build_masters(
     create_background_layers=False,
     generate_GDEF=True,
     store_editor_state=True,
+    write_public_skipexportglyphs=False,
```
anthrotype: It’s ok if you leave out the “public”

PR #528, opened 2019-05-22
Only write filters to masters

madig:
Before, going from UFO with ufo2ft filters to Glyphs and back to UFOs would leave you with `com.schriftgestaltung.customParameter.GSFont.*Filter` _and_ `com.github.googlei18n.ufo2ft.filters` in the UFO lib key. It still feels wrong to mix Glyphs filters and ufo2ft filters but ugh.

### DIFF
--- a/Lib/glyphsLib/builder/custom_params.py
+++ b/Lib/glyphsLib/builder/custom_params.py
@@ -620,6 +620,9 @@ class FilterParamHandler(AbstractParamHandler):
         return (UFO2FT_FILTERS_KEY,)
 
     def to_glyphs(self, glyphs, ufo):
+        if glyphs.is_font():
+            return  # Only write filters to the masters
+
         ufo_filters = ufo.get_lib_value(UFO2FT_FILTERS_KEY)
         if ufo_filters is None:
             return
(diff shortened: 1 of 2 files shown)


schriftgestalt:
I don’t really understand what this commit mean. But the title sounds wrong. Filters are only stored and applied in instances.

madig:
I know. This is specifically about `ufo2ft` filters, not Glyphs filters. For reasons lost on me, glyphsLib was/is set up to take ufo2ft filters and store them as `Filter` or `PreFilter` in .glyphs files. Maybe the idea was once to implement some Glyphs filters in ufo2ft and store them in their own lib key. Since ufo2ft filters work specifically in the masters and are inherited from the default master from there, we want to store these filters in the master.

madig:
Ha, maybe I misunderstood my own issue. The problem is that ufo2ft filters in a UFO lib get turned into `(Pre)Filter` custom parameters, which is okay, but when going back to UFOs, are turned back into a ufo2ft UFO lib key _and_ a `com.schriftgestaltung.customParameter.GSFont.PreFilter` UFO lib key. I have to suppress the latter.

schriftgestalt:
Can’t we standardize the position and format of the filters that utf2ft and Glyphs write the same?

madig:
ufo2ft filters came into life to support Glyphs filters afaik, so maybe? @anthrotype

PR #531, opened 2019-06-25
Factor out the application of Glyphs instance data

madig:
I want to use it in fontmake to do in-memory manipulation.


madig:
Blargh. For types to work correctly, you need to import the symbols. Maybe I'll just remove them.

Issue #529, opened 2019-06-19
Support Replace Prefix custom parameter


punchcutter:
This is mentioned somewhat in https://github.com/googlefonts/glyphsLib/issues/506 and https://github.com/googlefonts/glyphsLib/issues/253 and in some fontmake issues, but to fully support including external feature files the way Glyphs does we need to read this parameter. We have external files defined with `Replace Prefix` for each instance.

madig:
Out of curiosity, which/what kind of instance-specific data are you trying to insert?

anthrotype:
Thanks @punchcutter for the reminder.
So, this "Replace Prefix" custom parameter only works for instances, right? Not for masters.
If that's the case, I believe this approach of swapping features "prefixes" won't work when generating a variable font from the same source file. But perhaps that's ok if the source is only meant to be exported as static instances.

schriftgestalt:
Yes. This parameter only works for static instances. 

How to control details for variable fonts is to be seen. If someone has a good suggestion...

punchcutter:
What we have right now has chained contextual GSUB with different rules based on each master weight. Then all 36 instances point to one of those. Each one is hundreds of lines so although It's possible to keep it all inside the Glyphs file it's a little easier to handle as separate files. Since it's Noto it will go through the fontmake pipeline so we need to make sure this will work or we will have to rework the file.

anthrotype:
is this Noto source file going to be used to build a variable font as well as static instances, or only the latter?

punchcutter:
Uh oh. I forgot the Noto pipeline now makes variable fonts out of everything. Nothing in Noto was designed to be a variable font originally. They just have multiple weights so for the most part it's possible to make them variable. In that case these include statements aren't going to work.

anthrotype:
I see. Can I also ask what Noto font specifically this is about?

punchcutter:
It's Noto Serif Devanagari.

punchcutter:
Looking at this more there are a couple ways a Glyphs Prefix can be used
- Point to an external feature file with `include()`
- Hold lookups that can be referred to in contextual lookups written inside each feature

Originally I was only thinking of the first one, but to consolidate the external code I moved it into the Prefix. That way using Replace Prefix will replace the entire code block instead of only replacing an `include` statement.

anthrotype:
> chained contextual GSUB with different rules based on each master weight.

[...]

PR #534, opened 2019-07-15
Support Replace Prefix custom parameter

punchcutter:
Not sure if this is the most desirable way, but it's just a copy of the `Replace Feature` custom parameter and works if the `Replace Prefix` value is only code that needs to be inserted. If it's an `include()` statement pointing to external files then that's another issue to be dealt with.

### DIFF
--- a/Lib/glyphsLib/builder/custom_params.py
+++ b/Lib/glyphsLib/builder/custom_params.py
@@ -27,7 +27,7 @@ from .constants import (
     REVERSE_CODEPAGE_RANGES,
     PUBLIC_PREFIX,
 )
-from .features import replace_feature
+from .features import replace_feature, replace_prefix
 
 """Set Glyphs custom parameters in UFO info or lib, where appropriate.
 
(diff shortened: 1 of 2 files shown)


punchcutter:
Wasn't meant to be a total fix, but tests fail with this so drop it for now.

madig:
Me and Cosimo are currently on holidays, so maybe wait with this until someone is back and can take a look.

punchcutter:
This quick fix works so we can generate Noto Serif Devanagari as mentioned in https://github.com/googlefonts/glyphsLib/issues/529 and test before delivery, but it will need to be fixed properly. One problem here is that with `Replace Feature` a feature is actually enclosed in brackets with the name on both ends so it's easy to know where it begins and ends. Prefixes don't have anything like that so I added an `End Prefix` tag to encapsulate which then makes it easy to replace the entire Prefix properly, but that messes up round-tripping.

schriftgestalt:
Those parameters should be applied before the .fea file is written. But that would mean that the prefix, classes and features lists are stored inside the .ufo?

anthrotype:
fontmake applies the instance custom parameters on the instance UFOs. The master UFOs, from which the instances' features are copied, have the features.fea already stringified; so the problem is to parse and identify the prefix so it can be replaced with the custom one.

If it makes it simpler to cut it out, I'd be OK with adding an `# End Prefix` magic comment at the end.
However, there should already be some code written by @belluzj that accomplishes this task of splitting a features.fea into its component prefix, classes and features -- the one used in glyphsLib when converting UFOs features into .glyphs.

I just came back from my break, I'll take a look some time this week.

marekjez86:
@anthrotype : could you check it ASAP?

anthrotype:
@marekjez86 I'll look at it tomorrow

PR #537, opened 2019-07-24
Add support for 'Replace Prefix' custom parameter

### LINKED ISSUE #529: Support Replace Prefix custom parameter

anthrotype:
Fixes #529
Supersedes #534

/cc @punchcutter @marekjez86


marekjez86:
@anthrotype  @punchcutter :  thanks

anthrotype:
@belluzj thanks Jany

Issue #535, opened 2019-07-23
Bracket layer generation: update components to reference bracket layer base glyphs


punchcutter:
If a bracket layer is made up of components the components also need to be updated to point to the newly created layer. For example we have lowercase "a" with bracket at [150]. This creates "a.BRACKET.150", but "aogonek" also has [150] so "aogonek.BRACKET.150" should be updated so that the "a" component points to "a.BRACKET.150". I did this real quick and it seems fine, but I don't have time to mess with it more right now...

```
        # Finally, copy bracket layers to their own glyphs.
        bracket_layer_names = []
        for location, layers in bracket_layer_map.items():
            for layer in layers:
                ufo_font = self._sources[
                    layer.associatedMasterId or layer.layerId
                ].font.layers.defaultLayer
                ufo_glyph_name = "{glyph_name}.BRACKET.{location}".format(
                    glyph_name=layer.parent.name, location=location
                )
                bracket_layer_names.append(ufo_glyph_name)
                ufo_glyph = ufo_font.newGlyph(ufo_glyph_name)
                self.to_ufo_glyph(ufo_glyph, layer, layer.parent)
                ufo_glyph.unicodes = []  # Avoid cmap interference
                ufo_glyph.lib[GLYPHLIB_PREFIX + "_originalLayerName"] = layer.name

        for location, layers in bracket_layer_map.items():
            for layer in layers:
                ufo_font = self._sources[
                    layer.associatedMasterId or layer.layerId
                ].font.layers.defaultLayer
                for bracket_layer_name in bracket_layer_names:
                    for comp in ufo_font[bracket_layer_name].components:
                        comp_glyph_name = "{glyph_name}.BRACKET.{location}".format(
                        glyph_name=comp._baseGlyph, location=location)
                        if comp_glyph_name in bracket_layer_names:
                            comp.baseGlyph = comp_glyph_name

```

anthrotype:
I guess this is the same issue that @madig reported in https://github.com/googlefonts/glyphsLib/issues/468#issuecomment-499468096 ?

anthrotype:
@punchcutter by the way, does Glyphs.app itself support composite glyphs with bracket layers? I think it doesn't yet, see https://forum.glyphsapp.com/t/the-bracket-layer-does-not-work-with-the-component-in-vb-font/11577/6

punchcutter:
[...]

Issue #468, opened 2018-11-20
Support brace and bracket layers


madig:
Thanks to Behdad's recent efforts on sparse masters, we could now support brace layers. This is probably an issue for multiple layers of the stack. I'll start here because glyphsLib needs to output something. I think this ties into https://github.com/fonttools/fonttools/issues/1308.

designspaceLib supports substitution rules (https://github.com/fonttools/fonttools/blob/master/Lib/fontTools/designspaceLib/__init__.py#L140) to support bracket layers. How should brace layers/sparse masters be supported? For every brace layer, make a new source entry with the appropriate location and set `layerName` to the name of the brace layer? If e.g. there was a brace layer for `e` named `Medium {465 100}` in `MyFont_Th.ufo`:

```XML
    <sources>
        <source familyname="MyFont VF" filename="MyFont_Th.ufo" name="MyFont VF Thin" stylename="Thin">
            <lib copy="1" />
            <groups copy="1" />
            <features copy="1" />
            <info copy="1" />
            <location>
                <dimension name="Weight" xvalue="100" />
            </location>
        </source>
        <source familyname="MyFont VF" filename="MyFont_Th.ufo" layer="Medium {465 100}">
            <location>
                <dimension name="Weight" xvalue="465" />
            </location>
        </source>
        <source familyname="MyFont VF" filename="MyFont_Blk.ufo" name="MyFont VF Black" stylename="Black">
            <location>
                <dimension name="Weight" xvalue="1000" />
            </location>
        </source>
    </sources>
```

What already works is using a complete UFO (without feature file) as a sparse master, as I wrote here: https://github.com/googlei18n/fontmake/issues/477#issuecomment-440303283.

There's an example declaration for bracket layers at https://github.com/fonttools/fonttools/blob/master/Lib/fontTools/designspaceLib/__init__.py#L141. Will have to understand those better before including them in the example here.

madig:
I'm looking into bracket layers as Designspace has that covered by `rules`. Glyphs.app supports three cases of bracket layers as described in https://glyphsapp.com/tutorials/alternating-glyph-shapes, not counting the "Rename Glyphs" thing:

1. Bracket layer for all layers
2. Bracket layer for one layer
3. Reverse bracket layer for one layer

[...]

PR #538, opened 2019-07-26
[bracket layers] support reverse, unbalanced and composite glyphs

anthrotype:
Prompted by https://github.com/googlefonts/glyphsLib/issues/535, I decided to work on completing the bracket layers support started in https://github.com/googlefonts/glyphsLib/issues/468.

This adds support for
- "reverse" bracket layers (e.g. `]500]`) where the location represent the range's maximum, as opposed to regular bracket layers (`[500]`) where value indicates a minimum. 
- what Nikolaus called "unbalanced" bracket layers, i.e. when only some of master layers have explicilty defined bracket layers. The missing master layers are implicitly duplicated as bracket layers.
- composite glyphs with bracket layers that match the definition of their components' bracket layers (#535): the components' base glyphs need to be adjusted accordingly.

The PR is marked as WIP because next week I want to refine and add more specific tests for the new features.

### DIFF
--- a/Lib/glyphsLib/builder/builders.py
+++ b/Lib/glyphsLib/builder/builders.py
@@ -31,9 +31,7 @@ from .axes import WEIGHT_AXIS_DEF, WIDTH_AXIS_DEF, find_base_style, class_to_val
 
 GLYPH_ORDER_KEY = PUBLIC_PREFIX + "glyphOrder"
 
-BRACKET_LAYER_RE = re.compile(
-    r".*(?P<first_bracket>[\[\]])\s*(?P<value>\d+)\].*"
-)
+BRACKET_LAYER_RE = re.compile(r".*(?P<first_bracket>[\[\]])\s*(?P<value>\d+)\s*\].*")
 BRACKET_GLYPH_TEMPLATE = "{glyph_name}.{rev}BRACKET.{location}"
 REVERSE_BRACKET_LABEL = "REV_"
 BRACKET_GLYPH_RE = re.compile(
--- a/Lib/glyphsLib/builder/builders.py
+++ b/Lib/glyphsLib/builder/builders.py
@@ -206,7 +206,9 @@ class UFOBuilder(_LoggerMixin):
                 continue
 
             # Save processing bracket layers for when designspace() is called, as we
-            # have to extract them to free-standing glyphs.
+            # have to extract them to free-standing glyphs -- unless the parent glyph is
+            # set to non-export (in which case makes no sense to have Designspace rules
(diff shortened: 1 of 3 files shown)


madig:
@schriftgestalt I think we asked this already, but does Glyphs currently support more than one number for bracket layers?

schriftgestalt:
It does not.

anthrotype:
Cool, another mini-language to parse and support.

schriftgestalt:
You are free to propose a better solution. 
One other solution would be to have an actually data structure in the background and use the layer name only as an overview for the UI.

madig:
> One other solution would be to have an actually data structure in the background and use the layer name only as an overview for the UI.

I would actually want to have this. I think the Glyphs UI for bracket _and_ brace layers could be improved by seperating layer name and location data. See for example the mini-design-space view in FL6, which I find much easier to parse quickly than juggling brace layer numbers in my head. A break-out view UI to precisely set transition points for bracket layers (including preview) would be much more user-friendly. But let's discuss this in a seperate channel :)

Lib/glyphsLib/builder/builders.py
```diff
@@ -197,8 +208,8 @@ def masters(self):
             # Save processing bracket layers for when designspace() is called, as we
             # have to extract them to free-standing glyphs.
             if (
-                "[" in layer.name
-                and "]" in layer.name
+                BRACKET_LAYER_RE.match(layer.name)
+                and glyph.export
```
anthrotype: I added a check so that glyphs marked as non-export that contain bracket layers do not get processed as such; the layers are exported as regular layers to the UFO, not as separate bracket glyphs

belluzj: Why? Could you add a comment to explain why?

anthrotype: oh sure I can. I thought it made sense, but I can take it out if you think it's not a good idea. I was thinking, a non-export glyph may be one which contains unfinished work that is not interpolatable; and unbalanced bracket layers must be so, because they implicitly reference master layers. So marking the glyph as non-export should not prevent to build a varfont.
I guess with the new skipExportGlyphs feature this is not an issue because sources are subsetted before interpolation/compilation; but the old subsetting feature in fontmake happens after compilation.

anthrotype: oh now I remember another reason. You don't want to add desingspace rules referencing glyphs that are not going to be there in the exported font.

belluzj: Cool, that makes total sense, I was just wondering why

PR #543, opened 2019-08-15
Make sure master UFO filenames are unique even when a UFO_FILENAME_KEY is present

madig:


Lib/glyphsLib/builder/sources.py
```diff
@@ -82,25 +81,28 @@ def _to_designspace_source(self, master, is_regular):
...
+    ):
+        source.filename += n
+        n += "_"
+        logger.warning(
+            "The master with the style name '{}' (ID {}) has the same style name "
+            "as another one. All masters should have distinctive "
+            "(style) names. Use the 'Master name' custom parameter"
+            " on a master to give it a unique name. Proceeding "
+            "with an unchanged name, but appending '_' to the file"
+            " name on disk.".format(source.styleName, master.id)
```
belluzj: The warning only makes sense when there's no UFO_FILENAME_KEY.

Ideally, the warning should depend on the if/else above:
* if UFO_FILENAME_KEY: warn about two masters having the same UFO file name, and advise to change the custom parameter "UFO Filename" in the Glyphs UI
* else: keep current warning

Lib/glyphsLib/builder/sources.py
```diff
@@ -82,25 +81,28 @@ def _to_designspace_source(self, master, is_regular):
...
+    # Attention: The following is done regardless of where source.filename came from.
+    # Depending on the duplicate's order in the master list, this could lead to the
+    # legitimate master's filename getting an underscore appended!
+    n = "_"
+    while any(
+        s is not source and s.filename == source.filename
+        for s in self._sources.values()
+    ):
+        source.filename += n
+        n += "_"
```
belluzj: I think it would be better to add `#1`, `#2` etc to match what other tools do

PR #546, opened 2019-08-19
Don't write (Pre)Filter of masters to duplicate UFO lib keys

madig:
While Glyphs.app only supports (Pre)Filters in instances, glyphsLib also handles the same custom parameter in masters, transforming them into `com.github.googlei18n.ufo2ft.filters` UFO lib entries. I think the [ufo2ft filters](https://github.com/googlefonts/ufo2ft/tree/master/Lib/ufo2ft/filters) take inspiration from Glyphs.app filters and were made for specific use cases.

This PR fixes these master (Pre)Filters being transformed into the `com.github.googlei18n.ufo2ft.filters` key _and_ also being written verbatim to the `com.schriftgestaltung.customParameter.GSFont.(Pre)Filter` UFO lib key.

Supersedes #528.

### DIFF
--- a/Lib/glyphsLib/builder/custom_params.py
+++ b/Lib/glyphsLib/builder/custom_params.py
@@ -611,6 +611,9 @@ class GlyphOrderParamHandler(AbstractParamHandler):
 register(GlyphOrderParamHandler())
 
 
+# While Glyphs.app only supports (Pre)Filters in instances, we handle the same custom
+# parameter for masters as well, transforming them into
+# `com.github.googlei18n.ufo2ft.filters` UFO lib entries.
 # See https://github.com/googlefonts/glyphsLib/issues/214
 class FilterParamHandler(AbstractParamHandler):
     def glyphs_names(self):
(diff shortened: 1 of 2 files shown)


madig:
FWIW, I still find the filters situation confusing. Ideally, filters would be applied after instance UFOs have been generated, which the new fontmake can do.

schriftgestalt:
Why are the filters applied to the masters?

anthrotype:
in UFO world, masters are whole, separate UFO sources, and one may want to compile these masters into binary fonts (either standalone fonts or interpolatable masters for producing VFs). Whereas in Glyphs.app only instances proper can be exported to binary fonts, and masters do not exist as separate font sources.
fontmake, without -i/--interpolate option, only compiles the master fonts; when ufo2ft gets a UFO (whch can be either a master or an instance, ufo2ft doesn't care), it applies any filters to it before compiling to binary font.
It's true that ufo2ft filters were conceived to replicate Glyphs', and the intended use case was to apply them for instances only; however Nikolaus found some use for setting some ufo2ft filters (e.g. PropagateAnchors) on UFO masters, and apparently ufo2ft is happy to apply those.

For some reasons, when roundtripping these UFO masters back to .glyphs, such master-level ufo2ft filters are being stored twice (once as custom parameters [greyed out, of course, but we know that, it's just to save them somewhere], and again as unknown parameters in the UFO lib), which is what this PR is trying to fix.

madig:
One of my original use-cases was making use of ufo2ft filters to automate some things in some projects and glyphsLib tripped over them on round-tripping. I would be equally happy to remove the transform of UFO lib ufo2ft filters into Glyphs Filters and have them stored as verbatim custom parameters.

madig:
Note about the comments in `FilterParamHandler`: The data is read and written from and to masters. Instances don't turn up in my roundtrip tests. Filters for instances are put into the DS instance libs verbatim.

madig:
We decided to not round-trip ufo2ft filters as custom parameters because the only place where they make sense is instances.

belluzj:
I think this is what should happen ideally:
```          
  +------+
  |GSFont+-------------------+
  +----+-+                   |
       |                     |
     +-+-----------+       +-+----------+
     |GSFontMaster |       |GSIntance   |
     +-------------+       +------------+
      userData                    customParameters
        com...ufo2ft.filters        Filter & PreFilter

           ^  |                      |  ^
roundtrips |  |                      |  |
           |  v                      |  |
       lib                           |  | roundtrips
         com...ufo2ft.filters        |  |
     +-----------+                   v  |
     |Master UFO |          lib
     +---+-------+            com.schriftgestaltung.customParameter...
         |
     +---+-----+        +----------+                    +-----------------+
     | Source  |        | Instance |    ------------>   |Interpolated UFO |
     +---+-----+        +-----+----+                    +-----------------+
         |                    |          goes 1 way        lib
 +-------+-----+              |     apply_instance_data()    com...ufo2ft.filters
 | Designspace +--------------+
 +-------------+
```

The UFO2FT filters should roundtrip as-is between UFO source masters and GSFontMaster, because that's how we use them in the UFO workflow with 1 master UFO = 1 final font with filters applied.

The Glyphs filters defined on GSInstance should keep doing what they were doing already:
* first be copied as-is into the designspace instance's lib, which should roundtrip back to Glyphs
* then be converted to ufo2ft equivalents and put in the final interpolated UFOs before they are compiled into final fonts. Those should not roundtrip because the interpolated UFO is discarded after compilation.

anthrotype:
sounds good to me, thanks Jany for clarifying this.

PR #548, opened 2019-08-20
Store a master's ufo2ft filter in the master's userData

madig:


Lib/glyphsLib/builder/custom_params.py
```diff
@@ -611,28 +619,58 @@ def to_ufo(self, builder, glyphs, ufo):
...
+#
+# The Glyphs filters defined on GSInstance should keep doing what they were doing
+# already:
+#
+#  - first be copied as-is into the designspace instance's lib, which should roundtrip
+#    back to Glyphs
+#  - then be converted to ufo2ft equivalents and put in the final interpolated UFOs
+#    before they are compiled into final fonts. Those should not roundtrip because the
+#    interpolated UFO is discarded after compilation.
+#
```
belluzj: Maybe add that the param handler below handles only the "goes 1 way" case from the diagram

tests/builder/custom_params_test.py
```diff
@@ -409,8 +409,9 @@ def test_ufo2ft_filter_roundtrip(self):
         self.assertEqual(self.ufo.lib[UFO2FT_FILTERS_KEY], ufo_filters)
```
belluzj: The test is good but looks misleading. Maybe you could add the comment that explains this:

The first half is testing the "goes 1 way" part of the diagram, which happens on instances (although the test is performed on a master, but the code works the same for both), and checks that the filter syntax is translated from Glyphs to ufo2ft.

The second half tests the roundtrip of ufo2ft filters from UFO to GSFontMaster and back to UFO.

Issue #544, opened 2019-08-15
Drop Python 2 support


madig:
The upcoming fontTools 4.x branch will drop Python 2 support. We might want to use new features from it as we go, so drop Python 2?

This would prevent glyphsLib from being used in Glyphs, but I know of no one who actually does that.

hugovk:
Please see PR https://github.com/googlefonts/glyphsLib/issues/544.

PR #550, opened 2019-09-11
Drop support for legacy Python 2.7

hugovk:
Fixes https://github.com/googlefonts/glyphsLib/issues/544.

### DIFF
--- a/Lib/glyphsLib/classes.py
+++ b/Lib/glyphsLib/classes.py
@@ -37,7 +37,6 @@ from glyphsLib.parser import Parser
 from glyphsLib.writer import Writer
 from collections import OrderedDict
 from io import StringIO
-from fontTools.misc.py23 import unicode
 from glyphsLib.affine import Affine
 
 
@@ -250,7 +249,7 @@ def parse_hint_target(line=None):
 
 
 def isString(string):
-    return isinstance(string, (str, unicode))
+    return isinstance(string, str)
 
 
 def transformStructToScaleAndRotation(transform):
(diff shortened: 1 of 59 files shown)


madig:
Uh-oh, I feared this when I saw you forking this repo but had no way of contacting you. I haven't actually heard back from @anthrotype if we should drop Python 2 support already. I know I want to, but this lib can theoretically be used in Glyphs, which is Py2 only. So... @anthrotype? We want to heavily refactor the lib anyway at some point in the undefined future. Hmhmhm.

Lib/glyphsLib/classes.py
```diff
@@ -38,7 +36,8 @@
 from glyphsLib.parser import Parser
 from glyphsLib.writer import Writer
 from collections import OrderedDict
-from fontTools.misc.py23 import unicode, basestring, UnicodeIO, unichr, open
+from io import StringIO
+from fontTools.misc.py23 import unicode
```
anthrotype: you probably forgot to remove this line

anthrotype:
> this lib can theoretically be used in Glyphs, which is Py2 only

But do people actually use glyphsLib from within the Glyphs.app Macros window? Why would they want to use it? If it's for exporting/importing to UFO then they can keep using it from the command-line. 
I spoke with @schriftgestalt in Tokyo and he said that python3 support may be coming soon for Glyphs.app.

madig:
Continue at https://github.com/googlefonts/glyphsLib/pull/554.

Issue #547, opened 2019-08-19
zeroing combining marks advances from right-hand side?


anthrotype:
when glyphsLib exports combining mark glyphs (category "Mark", subCategory "Nonspacing") it simply sets their advance width to 0.

However, I just noticed that Glyphs.app not only sets their advance to 0, but it also _shifts_ their outlines to the ~~right~~ left by an amount equal to their advance width, so that the original right-side bearing is maintained.

@schriftgestalt can you confirm this?

I wonder if we need to do the same.

anthrotype:
sorry, I meant Glyphs.app "also shifts their outline to the _left_", not to the _right_.

in other words, it _subtracts_ the advance width from the glyph's horizontal (x) coordinates, including the components and anchors.

schriftgestalt:
Yes. Glyphs is doing that. But you need to shift all components of those marks. Nowadays you can much more on mark positioning. The reason to do the shift was that it would leave the mark in a somewhat OK position if there is not mark positioning.

m4rc1e:
I'm tempted to fix this so I can get cleaner diffs when I'm comparing fonts genned with glyphsapp against fontmake. Any objections?

anthrotype:
see https://github.com/googlefonts/glyphsLib/pull/551#issuecomment-532183247

m4rc1e:
> I personally would prefer if designers explicitly drew their combining mark glyphs with a zero advance width to begin with, instead of relying on the compiler to modify that at export time.

That's a good enough reason not to implement this imo. Instead, I'll write a check in our glyphapp scripts to ensure that comb marks are zero width.

anthrotype:
I remember there was some reason why designer make their combining mark with non-zero width but then Glyphs exports them with 0 width.. something to do with the way they show up in the UI.

m4rc1e:
Ah, I guess it's because of the preview window. The window doesn't insert any whitespace between glyphs so for 0 width glyphs they'd all be overlapping.

schriftgestalt:
When you select several marks in font view and open them in edit view, they would be all on top of each other. Same when you just have some letters and a mark in the middle it would be overlapping with the letters. It is much easier when they all have their own space. The requirement that those marks shouldn’t have a width is purely technical and I try to not bother the users with those.

madig:
Hm. Would it make sense to draw things that are detected as marks differently, i.e. make them not overlap while leaving the technical requirement of zero advance width out of sight?

[...]

PR #551, opened 2019-09-12
custom_params: add subscript and superscript params

m4rc1e:
I've noticed my VFs don't contain the subscript and superscript OS/2 attribs. Since we're pushing hard on OpenSans VF, I feel we need to add them.

To start with, I've only added "openTypeOS2SubscriptYOffset". I've got it to work but I don't know if my implementation is correct. If someone can review and confirm my current approach is ok, I'll add the rest of the attribs and make sure they're ordered better.


m4rc1e:
Alright, I'm now happy with this.

I'll start reading up on tox as well.

tests/builder/custom_params_test.py
```diff
@@ -120,31 +123,37 @@ def test_set_fsSelection_flags(self):
...
-        self.set_custom_params()
-        self.assertEqual(self.ufo.info.postscriptUnderlineThickness, 100)
+    def test_integer_parameters(self):
+        """Test casting glyphsapp customParameters whose values are just
+        integers into ufo equivalents."""
+        integer_params = [
+            "underlinePosition",
+            "underlineThickness",
+            "underlinePosition",
+            "underlineThickness",
```
anthrotype: You're testing twice the same parameters.
Also the strikeoutSize and Position are missing here

m4rc1e: Good catch. Thank you.

m4rc1e:
ping @anthrotype

anthrotype:
To be honest, I don't like this feature too much. Setting advance width to 0 is clear and easy (we do that already), whereas also shifting the mark glyph (as well as all its composites, anchors, etc.) to the left seems a bit arbitrary to me. It would also change all the glyf and GPOS tables if one rebuilds the same source font after glyphsLib would introduce this feature (which is actually what you want, in order to reduce the diff noise in your comparison tool).
But if you or others feel strongly about the need for implementing this, I am not going to veto it.
I personally would prefer if designers explicitly drew their combining mark glyphs with a zero advance width to begin with, instead of relying on the compiler to modify that at export time.

m4rc1e:
Think you meant #547

anthrotype:
oops, yes :)

anthrotype:
@m4rc1e note that if you want these custom params to be set in the generated VF, you will have to set them on the _masters_, not just on the instances. If they vary across masters, an MVAR table will contain deltas for them.
If you only set them on the instances, they won't be applied for the VF, but only for the static instances.

m4rc1e:
I'm aware sir and thank you for cutting a new release!

PR #554, opened 2019-09-14
Drop support for legacy Python 2.7

### LINKED ISSUE #544: Drop Python 2 support

madig:
Continuation from #550 because I can't push to hugovk's branch.

Closes #544.

There are more things that could be changed, like ,making the code in `types.ValueType` use f-strings, etc., but I feel that can also be done another time.


madig:
@hugovk Can you please give the bot your ok?

hugovk:
@googlebot I consent.

madig:
Thanks!

hugovk:
You're welcome!

PR #564, opened 2019-11-06
add support for ufoLib2 (in addition to defcon)

anthrotype:
use ufoLib2 by default -- but still allow to use defcon (or a compatible UFO library) as the `ufo_module` in build_masters, to_designspace, to_ufos, etc.

glyphs2ufo also has a new `--ufo-module` option.

### DIFF
--- a/Lib/glyphsLib/cli.py
+++ b/Lib/glyphsLib/cli.py
@@ -150,6 +150,16 @@ def main(args=None):
     parser_ufo2glyphs.add_argument(
         "--output-path", help="The path to write the Glyphs file to."
     )
+    parser_ufo2glyphs.add_argument(
+        "--ufo-module",
+        metavar="UFO_MODULE",
+        choices=("ufoLib2", "defcon"),
+        default="ufoLib2",
+        help=(
+            "Select the default library for reading UFOs. Choose between: %(choices)s "
+            "(default: %(default)s)"
+        ),
+    )
     group = parser_ufo2glyphs.add_argument_group(
         "Roundtripping between UFOs and Glyphs"
     )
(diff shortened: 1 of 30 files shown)


Lib/glyphsLib/cli.py
```diff
@@ -224,7 +235,7 @@ def ufo2glyphs(options):
         designspace.read(designspace_file)
         object_to_read = designspace
     elif all(source.endswith(".ufo") and os.path.isdir(source) for source in sources):
-        ufos = [defcon.Font(source) for source in sources]
+        ufos = [ufoLib2.Font.open(source) for source in sources]
```
madig: shouldn't this be ufo_module instead of ufoLib2?

anthrotype: No because that is ufo2glyphs

anthrotype: ufo_module is only for specifying the output UFO library

tests/test_helpers.py
```diff
@@ -191,20 +212,23 @@ def deboolize(lib):
...
-def normalize_ufo_lib(path):
+def normalize_ufo_lib(path, ufo_module):
     """Go through each `lib` element recursively and transform `bools` into
     `int` because that's what's going to happen on round-trip with Glyphs.
     """
-    font = defcon.Font(path)
+    try:
+        font = ufo_module.Font.open(path)
+    except AttributeError:
+        font = ufo_module.Font(path)
```
madig: maybe it's worth it to add `ufo_module_open` and `ufo_module_save` helper functions to encapsulate the API diffs. Also makes it easier to search and replace when defcon support is dropped.

Lib/glyphsLib/cli.py
madig: Ah, right.

PR #567, opened 2019-12-08
Profile guided performance optimization

madig:
I discovered https://github.com/nvdv/vprof and used it to run this script (Cantarell from https://gitlab.gnome.org/GNOME/cantarell-fonts/tree/dd36d229a2c304d17052dd62ad959f4e210c3c75/src):

```Python
import glyphsLib.cli

glyphsLib.cli.main(["ufo2glyphs", "Cantarell.designspace"])
```

Fedora 31 x64, Python 3.7.5, Ryzen 1700, all data in /tmp, which is a RAM-based tmpfs.

init

I tackled `GSNode._indices()` and `floatToString`.

After not calling `_indices()` on every node:

second

After simplifying `floatToString`:

third

That shaved off roughly 1 second, so it's a ~25% speed up :)

### DIFF
--- a/Lib/glyphsLib/classes.py
+++ b/Lib/glyphsLib/classes.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 
+import copy
 import logging
 import math
 import os
@@ -1332,11 +1333,11 @@ class GSGuideLine(GSBase):
 
     def __init__(self):
         self.alignment = ""
-        self.angle = 0
+        self.angle = self._defaultsForName["angle"]
         self.filter = ""
         self.locked = False
         self.name = ""
-        self.position = Point(0, 0)
+        self.position = copy.deepcopy(self._defaultsForName["position"])
         self.showMeasurement = False
 
     def __repr__(self):
@@ -1463,27 +1464,27 @@ class GSFontMaster(GSBase):
         self._name = None
         self._userData = None
         self.alignmentZones = []
-        self.ascender = 800
-        self.capHeight = 700
+        self.ascender = self._defaultsForName["ascender"]
+        self.capHeight = self._defaultsForName["capHeight"]
         self.customName = ""
-        self.customValue = 0
-        self.customValue1 = 0
-        self.customValue2 = 0
-        self.customValue3 = 0
-        self.descender = -200
+        self.customValue = self._defaultsForName["customValue"]
+        self.customValue1 = self._defaultsForName["customValue1"]
+        self.customValue2 = self._defaultsForName["customValue2"]
+        self.customValue3 = self._defaultsForName["customValue3"]
+        self.descender = self._defaultsForName["descender"]
         self.font = None
         self.guides = []
         self.horizontalStems = 0
         self.iconName = ""
         self.id = str(uuid.uuid4()).upper()
-        self.italicAngle = 0
+        self.italicAngle = self._defaultsForName["italicAngle"]
         self.verticalStems = 0
         self.visible = False
-        self.weight = "Regular"
-        self.weightValue = 100
-        self.width = "Regular"
-        self.widthValue = 100
-        self.xHeight = 500
+        self.weight = self._defaultsForName["weight"]
+        self.weightValue = self._defaultsForName["weightValue"]
+        self.width = self._defaultsForName["width"]
+        self.widthValue = self._defaultsForName["widthValue"]
+        self.xHeight = self._defaultsForName["xHeight"]
 
     def __repr__(self):
         return '<GSFontMaster "{}" width {} weight {}>'.format(
@@ -1774,7 +1775,7 @@ class GSPath(GSBase):
     _parent = None
 
     def __init__(self):
-        self.closed = True
+        self.closed = self._defaultsForName["closed"]
         self.nodes = []
 
     @property
@@ -2082,7 +2083,7 @@ class GSComponent(GSBase):
                 dx, dy = offset
                 self.transform = Transform(xx, 0, 0, yy, dx, dy)
             else:
-                self.transform = Transform(1, 0, 0, 1, 0, 0)
+                self.transform = copy.deepcopy(self._defaultsForName["transform"])
         else:
             self.transform = transform
 
@@ -2229,10 +2230,10 @@ class GSSmartComponentAxis(GSBase):
 
     def __init__(self):
         self.bottomName = ""
-        self.bottomValue = 0
+        self.bottomValue = self._defaultsForName["bottomValue"]
         self.name = ""
         self.topName = ""
-        self.topValue = 0
+        self.topValue = self._defaultsForName["topValue"]
 
     def shouldWriteValueForKey(self, key):
         if key in ("bottomValue", "topValue"):
@@ -2249,7 +2250,10 @@ class GSAnchor(GSBase):
 
     def __init__(self, name=None, position=None):
         self.name = "" if name is None else name
-        self.position = Point(0, 0) if position is None else position
+        if position is None:
+            self.position = copy.deepcopy(self._defaultsForName["position"])
+        else:
+            self.position = position
 
     def __repr__(self):
         return '<{} "{}" x={:.1f} y={:.1f}>'.format(
@@ -2325,21 +2329,16 @@ class GSHint(GSBase):
     )
 
     def __init__(self):
-        self._origin = None
-        self._originNode = None
-        self._other1 = None
-        self._other2 = None
-        self._otherNode1 = None
-        self._otherNode2 = None
-        self._target = None
-        self._targetNode = None
         self.horizontal = False
         self.name = ""
         self.options = 0
-        self.place = None
-        self.scale = None
+        self.origin = self._defaultsForName["origin"]
+        self.other1 = self._defaultsForName["other1"]
+        self.other2 = self._defaultsForName["other2"]
+        self.place = self._defaultsForName["place"]
+        self.scale = self._defaultsForName["scale"]
         self.settings = {}
-        self.stem = -2
+        self.stem = self._defaultsForName["stem"]
         self.type = ""
 
     def shouldWriteValueForKey(self, key):
@@ -2563,11 +2562,11 @@ class GSAnnotation(GSBase):
     _parent = None
 
     def __init__(self):
-        self.angle = 0
-        self.position = Point(0, 0)
-        self.text = None
-        self.type = 0
-        self.width = 100
+        self.angle = self._defaultsForName["angle"]
+        self.position = copy.deepcopy(self._defaultsForName["position"])
+        self.text = self._defaultsForName["text"]
+        self.type = self._defaultsForName["type"]
+        self.width = self._defaultsForName["width"]
 
     @property
     def parent(self):
@@ -2660,23 +2659,25 @@ class GSInstance(GSBase):
 
     def __init__(self):
         self._customParameters = []
-        self.active = True
+        self.active = self._defaultsForName["active"]
         self.custom = None
-        self.customValue = 0
-        self.customValue1 = 0
-        self.customValue2 = 0
-        self.customValue3 = 0
-        self.instanceInterpolations = {}
+        self.customValue = self._defaultsForName["interpolationCustom"]
+        self.customValue1 = self._defaultsForName["interpolationCustom1"]
+        self.customValue2 = self._defaultsForName["interpolationCustom2"]
+        self.customValue3 = self._defaultsForName["interpolationCustom3"]
+        self.instanceInterpolations = copy.deepcopy(
+            self._defaultsForName["instanceInterpolations"]
+        )
         self.isBold = False
         self.isItalic = False
         self.linkStyle = ""
         self.manualInterpolation = False
         self.name = "Regular"
         self.visible = True
-        self.weight = "Regular"
-        self.weightValue = 100
-        self.width = "Medium (normal)"
-        self.widthValue = 100
+        self.weight = self._defaultsForName["weightClass"]
+        self.weightValue = self._defaultsForName["interpolationWeight"]
+        self.width = self._defaultsForName["widthClass"]
+        self.widthValue = self._defaultsForName["interpolationWidth"]
 
     customParameters = property(
         lambda self: CustomParametersProxy(self),
@@ -2803,13 +2804,13 @@ class GSBackgroundImage(GSBase):
 
     def __init__(self, path=None):
         self._R = 0.0
-        self._alpha = 50
         self._sX = 1.0
         self._sY = 1.0
+        self.alpha = self._defaultsForName["alpha"]
         self.crop = Rect()
         self.imagePath = path
         self.locked = False
-        self.transform = Transform(1, 0, 0, 1, 0, 0)
+        self.transform = copy.deepcopy(self._defaultsForName["transform"])
 
     def __repr__(self):
         return "<GSBackgroundImage '%s'>" % self.imagePath
@@ -2985,14 +2986,14 @@ class GSLayer(GSBase):
         self.associatedMasterId = ""
         self.backgroundImage = None
         self.color = None
-        self.leftMetricsKey = None
+        self.leftMetricsKey = self._defaultsForName["leftMetricsKey"]
         self.parent = None
-        self.rightMetricsKey = None
-        self.vertOrigin = 0
-        self.vertWidth = 0
+        self.rightMetricsKey = self._defaultsForName["rightMetricsKey"]
+        self.vertOrigin = self._defaultsForName["vertOrigin"]
+        self.vertWidth = self._defaultsForName["vertWidth"]
         self.visible = False
-        self.width = 600
-        self.widthMetricsKey = None
+        self.width = self._defaultsForName["width"]
+        self.widthMetricsKey = self._defaultsForName["widthMetricsKey"]
 
     def __repr__(self):
         name = self.name
@@ -3321,31 +3322,31 @@ class GSGlyph(GSBase):
     def __init__(self, name=None):
         self._layers = OrderedDict()
         self._unicodes = []
-        self._userData = None
         self.bottomKerningGroup = ""
         self.bottomMetricsKey = ""
-        self.category = None
-        self.color = None
-        self.export = True
-        self.lastChange = None
-        self.leftKerningGroup = None
+        self.category = self._defaultsForName["category"]
+        self.color = self._defaultsForName["color"]
+        self.export = self._defaultsForName["export"]
+        self.lastChange = self._defaultsForName["lastChange"]
+        self.leftKerningGroup = self._defaultsForName["leftKerningGroup"]
         self.leftKerningKey = ""
-        self.leftMetricsKey = None
+        self.leftMetricsKey = self._defaultsForName["leftMetricsKey"]
         self.name = name
-        self.note = None
+        self.note = self._defaultsForName["note"]
         self.parent = None
         self.partsSettings = []
         self.production = ""
-        self.rightKerningGroup = None
+        self.rightKerningGroup = self._defaultsForName["rightKerningGroup"]
         self.rightKerningKey = ""
-        self.rightMetricsKey = None
-        self.script = None
+        self.rightMetricsKey = self._defaultsForName["rightMetricsKey"]
+        self.script = self._defaultsForName["script"]
         self.selected = False
-        self.subCategory = None
+        self.subCategory = self._defaultsForName["subCategory"]
         self.topKerningGroup = ""
         self.topMetricsKey = ""
+        self.userData = self._defaultsForName["userData"]
         self.vertWidthMetricsKey = ""
-        self.widthMetricsKey = None
+        self.widthMetricsKey = self._defaultsForName["widthMetricsKey"]
 
     def __repr__(self):
         return '<GSGlyph "{}" with {} layers>'.format(self.name, len(self.layers))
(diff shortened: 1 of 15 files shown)


madig:
Loading UFOs eagerly does not change performance, but cleans up the flame graph and we need to load everything anyway:

Bildschirmfoto vom 2019-12-08 13-29-11

madig:
This is what glyphs2ufo looks like on a Mac mini (late 2014):

image

That `clean_ufo` looks weird. Doesn't ufoLib/2 do that automatically?

madig:
In the ufo2glyphs case, the function `_get_glyph_by_string` takes up a large chunk of time. The function starts with the comment `# FIXME: (jany) looks inefficient`, which is correct 😁

madig:
The `Font.glyphs` list is a headache to deal with. I'm torn between making minimal changes and replacing the `Font._glyphs` container with a multi-index container, so all the lookups are O(1).

schriftgestalt:
In glyphs I have a name2Glyphs dict that is updated a) when a glyphs is added/removed b) when a glyph name is changed. I didn't add that to the python version as I wasn’t sure how to make sure that all changes trigger a cache update.

madig:
That's what I experimented with. It is a bit of an invasive change though... and I found that Glyphs will not clear out glyphs with duplicate names, so there are ugly edge-cases to take care of. The second problem is that the ability to look up glyphs not just by their name but also by their Unicode char and hex string tanks performance for the case where a glyph name is not present: the glyphsLib code will then iterate over all glyphs again, usually finding nothing. In the profiler, that code path was hotter than the actual glyph name lookup :(

schriftgestalt:
There is another unicode2glyphs cache ;)

madig:
One of the bigger chunks left in the middle is `GSBase.__init__`. Initializing GS classes is very expensive.

madig:
The glyphs2ufo case is so dominated by reading the .glyphs file and writing the UFOs that optimizing glyphsLib inbetween makes little sense.

Reading and writing UFOs can take ~50% of the time both ways. Writing .glyphs files takes 10%, reading 30%. Interestingly, writing UFOs on Py3.7 is slower than on 3.6 🤔

schriftgestalt:
The `GSBase.__init__` was intended as a temporary solution until all classes have all properties written out. And for those classes you don’t need to call `super().__init__()` any more. GSNode seems to be done, so its `super().__init__()` can go. Next would be `GSPath` `GSComponent`, GSAnchor` and `GSLayer` (those classes are used for the most objects). That should remove most of the overhead.
The `_classesForName` seems to be used for some other things, too, so it might be good to keep it around.

madig:
Ah! I think I now understand the init mechanism better. I removed init from various classes as suggested and it does indeed cut down the flame graph somewhat.

madig:
@schriftgestalt can a node be of type "n/a"?

schriftgestalt:
No, that is only there for debugging

madig:
I found that while profiling glyphs2ufo of NotoSans-MM.glyphs, 5-6 seconds of 55 seconds in total on my Mac mini late 2014 are spent on converting the lastChanged glyph attribute to a UFO timestamp. Wow.

madig:
I did a quick comparison of `glyphsLib.parser.Parser` against openstep-plist:

```Python
from pathlib import Path
import glyphsLib.parser
import openstep_plist

glyphs_text = Path("../noto-source/src/NotoSans-MM.glyphs").read_text()

p = glyphsLib.parser.Parser(dict)
g1 = p.parse(glyphs_text)

g2 = openstep_plist.loads(glyphs_text, use_numbers=True)
```

openstep_plist doesn't even show up in the flame graph. Total profiler running time 9s, of that 0.4s for openstep_plist 😊

madig:
The by far biggest chunk (~50%) of the glyphs2ufo time goes into writing UFOs to disk. There must be a way to speed that up in ufoLib. Parallel serialization and dumping to disk using multiprocessing or asyncio or a combination of those maybe.

madig:
Thanks @chrissimpkins for pointing me to aiofiles. I made a test script which I can time to compare synchronous vs. async writing. I think I'm doing this wrong however:

```Python
import asyncio
import os
import random
import tempfile

import aiofiles

file_blobs = [
    bytearray(random.getrandbits(8) for _ in range(random.randint(20, 200)))
    for _ in range(3000)
]


def write_sequentially():
    tmp_dir = tempfile.mkdtemp()
    for i, blob in enumerate(file_blobs):
        with open(os.path.join(tmp_dir, str(i)), "bw+") as f:
            f.write(blob)


async def write_async():
    tmp_dir = tempfile.mkdtemp()
    await asyncio.gather(
        *(
            asyncio.ensure_future(write_async_blob(os.path.join(tmp_dir, str(i)), blob))
            for i, blob in enumerate(file_blobs)
        ),
        return_exceptions=True,
    )


async def write_async_blob(path, binary):
    async with aiofiles.open(path, "bw+") as f:
        await f.write(binary)
```
```
In [1]: %timeit asyncio.run(write_async())                                                       
1.15 s ± 18.5 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

In [2]: %timeit write_sequentially()                                                             
90.7 ms ± 419 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

async writing is slower :thinking:

chrissimpkins:
Hmm...  Might have to do with Python timing and the coroutines.  Here is what I get when I add an `if __name__ == '__main__'` statement to execute the two functions.  Timing is with gnu `time` x 3  on MBP 2.9 GHz Quad-Core Intel Core i7 + macOS 10.15.1:

For `write_async`:
```
$ gtime -p python3 arun.py
real 0.47
user 0.40
sys 0.12
$ gtime -p python3 arun.py
real 0.47
user 0.40
sys 0.13
$ gtime -p python3 arun.py
real 0.47
user 0.40
sys 0.12
```

For `write_sequentially`:

```
$ gtime -p python3 arun.py
real 0.57
user 0.20
sys 0.35
$ gtime -p python3 arun.py
real 0.58
user 0.20
sys 0.35
$ gtime -p python3 arun.py
real 0.62
user 0.21
sys 0.36
```

madig:
Hm. Running the script on a RAM-backed /tmp (and on  a SSD-backed one) on a Ryzen 1700, Fedora 31 x64, Python 3.7.5 yields:

Async:
```
(venv) /tmp $ time python a.py
0.72user 0.23system 0:00.76elapsed 126%CPU (0avgtext+0avgdata 44484maxresident)k
0inputs+0outputs (0major+10271minor)pagefaults 0swaps
(venv) /tmp $ time python a.py
0.71user 0.25system 0:00.76elapsed 127%CPU (0avgtext+0avgdata 44068maxresident)k
0inputs+0outputs (0major+10305minor)pagefaults 0swaps
(venv) /tmp $ time python a.py
0.72user 0.26system 0:00.75elapsed 130%CPU (0avgtext+0avgdata 44544maxresident)k
0inputs+0outputs (0major+10213minor)pagefaults 0swaps
```

Sequential:
```
(venv) /tmp $ time python a.py
0.14user 0.05system 0:00.19elapsed 98%CPU (0avgtext+0avgdata 16224maxresident)k
0inputs+0outputs (0major+2404minor)pagefaults 0swaps
(venv) /tmp $ time python a.py
0.14user 0.05system 0:00.20elapsed 98%CPU (0avgtext+0avgdata 16500maxresident)k
0inputs+0outputs (0major+2441minor)pagefaults 0swaps
(venv) /tmp $ time python a.py
0.14user 0.05system 0:00.19elapsed 98%CPU (0avgtext+0avgdata 16296maxresident)k
0inputs+0outputs (0major+2421minor)pagefaults 0swaps
```

So it's much slower :(

madig:
This is even slower:

```Python
# [...]

async def write_async_chunked():
    tmp_dir = tempfile.mkdtemp()

    blobs = [
        (os.path.join(tmp_dir, str(i)), binary) for i, binary in enumerate(file_blobs)
    ]
    blob_half_len = len(blobs) // 2
    tasks = [
        asyncio.ensure_future(write_async_chunked_blobs(blobs[:blob_half_len])),
        asyncio.ensure_future(write_async_chunked_blobs(blobs[blob_half_len:])),
    ]

    await asyncio.gather(
        *tasks, return_exceptions=True,
    )


async def write_async_chunked_blobs(path_to_binary):
    for path, binary in path_to_binary:
        async with aiofiles.open(path, "bw+") as f:
            await f.write(binary)

# [...]
```

```
(venv) /tmp $ time python a.py
1.80user 0.74system 0:01.98elapsed 128%CPU (0avgtext+0avgdata 19304maxresident)k
0inputs+0outputs (0major+3311minor)pagefaults 0swaps
```

Not sure whether I'm doing something wrong or if asyncio does not help us here.

chrissimpkins:
I converted to a non-tmp directory and the async function is not writing all of the expected files.  I only get 250 of the 3000 files with the above code.  If I add a semaphore to limit the number of concurrent writes to 10 - 1000 range, I see all of the files.  But you are right, it is much slower than the sequential write approach.  Perhaps OS's know how to write files better than we do? :)  I don't have much experience with the asyncio/aiofiles libraries.  Maybe this is not the correct approach.

```python
import asyncio
import os
import random
import tempfile

import aiofiles

file_blobs = [
    bytearray(random.getrandbits(8) for _ in range(random.randint(20, 200)))
    for _ in range(3000)
]


DIRPATH = "/Users/chris/Desktop/asyncio"


def write_sequentially():
    # tmp_dir = tempfile.mkdtemp()
    for i, blob in enumerate(file_blobs):
        with open(os.path.join(DIRPATH, str(i) + ".bin"), "bw+") as f:
            f.write(blob)


async def write_async():
    # tmp_dir = tempfile.mkdtemp()
    await asyncio.gather(
        *(
            asyncio.ensure_future(write_async_blob(os.path.join(DIRPATH, str(i) + ".bin"), blob))
            for i, blob in enumerate(file_blobs)
        ),
        return_exceptions=True,
    )


async def write_async_2():
    # tmp_dir = tempfile.mkdtemp()
    semaphore = asyncio.Semaphore(value=1000)
    tasks = []
    for i, blob in enumerate(file_blobs):
        task = write_async_blob(os.path.join(DIRPATH, str(i) + ".bin"), blob, semaphore)
        tasks.append(task)

    await asyncio.gather(
        *tasks,
        return_exceptions=True,
    )


async def write_async_blob(path, binary, semaphore):
    await semaphore.acquire()
    async with aiofiles.open(path, "bw+") as f:
        await f.write(binary)
    
    await semaphore.release()


if __name__ == "__main__":
    asyncio.run(write_async_2())
    # write_sequentially()
```

chrissimpkins:
> Edit2: https://trio.readthedocs.io/en/stable/reference-io.html#asynchronous-filesystem-i-o hm :(

Workaround: SSD upgrade :)

madig:
Hm. The slotting stuff is odd. It improves the prfiler runtime by 0.1 seconds, but memory usage for some classes is reported to be slightly worse by pympler:

Script:
```Python
import pympler.classtracker

import glyphsLib.cli
import glyphsLib.classes

tracker = pympler.classtracker.ClassTracker()
tracker.track_class(glyphsLib.classes.GSFont)
tracker.track_class(glyphsLib.classes.GSFontMaster)
tracker.track_class(glyphsLib.classes.GSAlignmentZone)
tracker.track_class(glyphsLib.classes.GSInstance)
tracker.track_class(glyphsLib.classes.GSCustomParameter)
tracker.track_class(glyphsLib.classes.GSClass)
tracker.track_class(glyphsLib.classes.GSFeaturePrefix)
tracker.track_class(glyphsLib.classes.GSFeature)
tracker.track_class(glyphsLib.classes.GSGlyph)
tracker.track_class(glyphsLib.classes.GSLayer)
tracker.track_class(glyphsLib.classes.GSAnchor)
tracker.track_class(glyphsLib.classes.GSComponent)
tracker.track_class(glyphsLib.classes.GSSmartComponentAxis)
tracker.track_class(glyphsLib.classes.GSPath)
tracker.track_class(glyphsLib.classes.GSNode)
tracker.track_class(glyphsLib.classes.GSGuideLine)
tracker.track_class(glyphsLib.classes.GSAnnotation)
tracker.track_class(glyphsLib.classes.GSHint)
tracker.track_class(glyphsLib.classes.GSBackgroundImage)
# No Proxy objects because they don't seem to be profile-able by pympler?!

glyphsLib.cli.main(["glyphs2ufo", "/tmp/Cantarell.glyphs"])

tracker.create_snapshot()
tracker.stats.print_summary()
```

No slots:
```
---- SUMMARY ------------------------------------------------------------------
                                         active      0     B      average   pct
  glyphsLib.classes.GSAlignmentZone          15      3.41 KB    232     B    0%
  glyphsLib.classes.GSAnchor               1047      0     B      0     B    0%
  glyphsLib.classes.GSClass                   1      0     B      0     B    0%
  glyphsLib.classes.GSComponent            3891      0     B      0     B    0%
  glyphsLib.classes.GSCustomPara...          37     10.06 KB    278     B    0%
  glyphsLib.classes.GSFeature                19      0     B      0     B    0%
  glyphsLib.classes.GSFeaturePrefix           3      0     B      0     B    0%
  glyphsLib.classes.GSFont                    1     17.98 KB     17.98 KB    0%
  glyphsLib.classes.GSFontMaster              3      0     B      0     B    0%
  glyphsLib.classes.GSGlyph                1011      0     B      0     B    0%
  glyphsLib.classes.GSInstance                6      0     B      0     B    0%
  glyphsLib.classes.GSLayer                3117      3.55 MB      1.17 KB    0%
  glyphsLib.classes.GSNode                21989     11.97 MB    570     B    1%
  glyphsLib.classes.GSPath                 1917      0     B      0     B    0%
-------------------------------------------------------------------------------
```

Slots:
```
---- SUMMARY ------------------------------------------------------------------
                                         active      0     B      average   pct
  glyphsLib.classes.GSAlignmentZone          15      3.52 KB    240     B    0%
  glyphsLib.classes.GSAnchor               1047      0     B      0     B    0%
  glyphsLib.classes.GSClass                   1      0     B      0     B    0%
  glyphsLib.classes.GSComponent            3891      0     B      0     B    0%
  glyphsLib.classes.GSCustomPara...          37     10.44 KB    288     B    0%
  glyphsLib.classes.GSFeature                19      0     B      0     B    0%
  glyphsLib.classes.GSFeaturePrefix           3      0     B      0     B    0%
  glyphsLib.classes.GSFont                    1     16.13 KB     16.13 KB    0%
  glyphsLib.classes.GSFontMaster              3      0     B      0     B    0%
  glyphsLib.classes.GSGlyph                1011      0     B      0     B    0%
  glyphsLib.classes.GSInstance                6      0     B      0     B    0%
  glyphsLib.classes.GSLayer                3117      3.20 MB      1.05 KB    0%
  glyphsLib.classes.GSNode                21989     12.64 MB    602     B    1%
  glyphsLib.classes.GSPath                 1917      0     B      0     B    0%
-------------------------------------------------------------------------------
```

Inclined to leave slots in anyway.

schriftgestalt:
When doing async stuff there is an overhead with switching threads. So you should not do too many of them but package several operations that you end up with only a few threads.

madig:
Python's async is still single-threaded. I think aiofiles uses threads to make disk IO awaitable. That's probably a different construction site...

schriftgestalt:
I’ve read a bit about Python's async. From what I understand from the introduction, it should not help in this case. Bevause each operation is not really waiting for something. Network requests could benefit from asyncIO as there is some wait time involved. 

In case of writing a .ufo it could help if you write the individual files in parallel. Each operation is isolated and shouldn't influence the others. So there shouldn’t be too many thread safety problems.

madig:
The idea was that multiple write ops could be kicked off with async inside a single thread and the write-complete time could be used to kick off the next. But maybe writing the stuff is faster than switching to the next coroutine :/ The only way to get more than one thread involved is `multiprocessing`. Maybe it would help to have a few processes that each take on part of the load (i.e. one master each) and then serialize and write to disk each sequentially.

Note to self: https://pypi.org/project/aiomultiprocess/

madig:
# Current status of PR:

As tested on: Fedora 31 x64, Python 3.7.5, Ryzen 1700, data on RAM-backed /tmp

## glyphs2ufo of Cantarell.glyphs:

Before:
```
3.09user 0.16system 0:03.28elapsed 99%CPU (0avgtext+0avgdata 113760maxresident)k
40inputs+272outputs (0major+32852minor)pagefaults 0swaps
3.10user 0.15system 0:03.27elapsed 99%CPU (0avgtext+0avgdata 113944maxresident)k
0inputs+0outputs (0major+31432minor)pagefaults 0swaps
3.07user 0.15system 0:03.24elapsed 99%CPU (0avgtext+0avgdata 113864maxresident)k
0inputs+0outputs (0major+31443minor)pagefaults 0swaps
```

After:
```
2.65user 0.21system 0:03.02elapsed 94%CPU (0avgtext+0avgdata 113712maxresident)k
41408inputs+0outputs (38major+31384minor)pagefaults 0swaps
2.65user 0.15system 0:02.82elapsed 99%CPU (0avgtext+0avgdata 113240maxresident)k
0inputs+0outputs (0major+31405minor)pagefaults 0swaps
2.67user 0.17system 0:02.86elapsed 99%CPU (0avgtext+0avgdata 113600maxresident)k
0inputs+0outputs (0major+31397minor)pagefaults 0swaps
```

## ufo2glyphs of Cantarell.designspace:

Before:
```
4.90user 0.07system 0:05.01elapsed 99%CPU (0avgtext+0avgdata 110828maxresident)k
360inputs+0outputs (1major+30619minor)pagefaults 0swaps
4.83user 0.09system 0:04.96elapsed 99%CPU (0avgtext+0avgdata 110900maxresident)k
0inputs+0outputs (0major+30589minor)pagefaults 0swaps
4.85user 0.10system 0:04.98elapsed 99%CPU (0avgtext+0avgdata 110924maxresident)k
0inputs+0outputs (0major+30604minor)pagefaults 0swaps
```

After:
```
3.17user 0.08system 0:03.28elapsed 99%CPU (0avgtext+0avgdata 113240maxresident)k
0inputs+280outputs (0major+32501minor)pagefaults 0swaps
3.20user 0.08system 0:03.30elapsed 99%CPU (0avgtext+0avgdata 113004maxresident)k
0inputs+0outputs (0major+31048minor)pagefaults 0swaps
3.18user 0.08system 0:03.28elapsed 99%CPU (0avgtext+0avgdata 112836maxresident)k
0inputs+0outputs (0major+31159minor)pagefaults 0swaps
```

chrissimpkins:
> ... I wish I could do from Rust

I recalled this comment when I came across this talk on the FOSDEM 2020 program:

https://fosdem.org/2020/schedule/event/python2020_rust/

schriftgestalt:
I don’t know much about that but wouldn’t it be easier to use cPython for performance critical code? That can speed up things quite a bit as we saw with the plist parser.

Lib/glyphsLib/classes.py
```diff
@@ -373,6 +373,8 @@ def shouldWriteValueForKey(self, key):
 
 
 class Proxy:
+    __slots__ = "_owner"
```
anthrotype: shouldn't `__slots__` be a tuple?

madig: It seems to work? :o

madig:
> I recalled this comment when I came across this talk on the FOSDEM 2020 program:
> 
> https://fosdem.org/2020/schedule/event/python2020_rust/

Yes, there are ways to have Rust interface with Python. We are not at that stage though... it's something for a redesign/major refactor of glyphsLib.

tests/classes_test.py
```diff
@@ -1562,7 +1565,7 @@ def test_position(self):
         self.assertIsInstance(self.node.position, Point)
 
     def test_type(self):
-        self.assertTrue(self.node.type in [GSNode.LINE, GSNode.CURVE, GSNode.OFFCURVE])
+        self.assertTrue(self.node.type in [LINE, CURVE, OFFCURVE])
```
anthrotype: GSNode type could be an `enum`

Lib/glyphsLib/classes.py
anthrotype: funny. looks like it's a feature
https://github.com/python/typeshed/issues/3079#issuecomment-504157352

madig:
The optimization in https://github.com/googlefonts/glyphsLib/pull/567/commits/0bde3afb8bbada900cac2940106238dece793c92 saves roughly 0.2 seconds, but adds complexity and is not the right solution yet... Keep it in but replace later?

madig:
Some measurements for using _defaultsForName on my modest Windows workstation (output pruned):

Bare init:

```
> (Measure-Command { ufo2glyphs Cantarell.designspace }).TotalMilliseconds
3286.0608
3248.3475
3249.4795

> (Measure-Command { glyphs2ufo Cantarell.glyphs }).TotalMilliseconds
5698.2601
5544.7676
5628.9273
```

Inits using _defaultsForName:

```
> (Measure-Command { ufo2glyphs Cantarell.designspace }).TotalMilliseconds
3390.8462
3317.1327
3301.2579
3316.6794

> (Measure-Command { glyphs2ufo Cantarell.glyphs }).TotalMilliseconds
5677.1592
5630.783
5637.7122
```

So the performance hit is small.

madig:
To preserve for future endeavours: https://github.com/benfred/py-spy plus https://www.speedscope.app/

run like `py-spy record --format speedscope -o profile.speedscope -- python bench_ufo2glyphs.py`, then drop the `profile.speedscope` file into https://www.speedscope.app/.

schriftgestalt:
I forgot go mention this before. In the timings mentioned above, you measure the full .glyphs to .designspace process. But you change a detail on a part that is already quite small. If that part will be twice as long, the impact on the overall time seems small. So I would measure smaller pieces when deciding about performance relevant changes.

madig:
What matters to me is end-to-end performance, so that's what I measured.

Issue #566, opened 2019-12-04
error exporting UFO when a  GSGlyph contain multiple layers with the same name


anthrotype:
when building master UFOs, if a Glyphs's glyph object has more than one supplementary layers with identical name, the builder will attempt to call `ufoLib2.Layer.newGlyph` method twice for the same layer and glyph name, resulting in errors like this

```
[...]
  File "glyphsLib/builder/builders.py", line 222, in masters
    ufo_glyph = ufo_layer.newGlyph(glyph.name)
  File "ufoLib2/objects/layer.py", line 120, in newGlyph
    raise KeyError("glyph %r already exists" % name)
KeyError: "glyph '_part.spirale' already exists"
```

anthrotype:
i am not sure how we should fix this. If it is a common thing to have multiple layers within the same glyph with the same name, then we should support it somehow without losing data. Otherwise, I'd be inclined to simply warn and only save the last duplicate glyph layer.

By the way, when were were using defcon instead of ufoLib2, we were silently taking the last duplicate-name glyph layer.

There is in fact a difference between defcon's and ufoLib2's `newGlyph` methods: the former overwrites any existing glyph, whereas the latter raises an error...

schriftgestalt:
I wouldn’t say it is common but it is possible. e.g. with interpolating 'iColor XX' or 'Color XX' layers. And a 'Color XX' layer might show up twice inside a master. 
So possible setups:
```
Regular 
    Color 1
    Color 2
Bold
    Color 1
    Color 2
```
or 
```
Regular 
    Color 1
    Color 2
    Color 1
```

madig:
I'm inclined to warn and take the last layer.

anthrotype:
> interpolating 'iColor XX' or 'Color XX' layers. And a 'Color XX' layer might show up twice inside a master

is this some special type of layer naming scheme that we should be aware of?

anthrotype:
replying to myself, yes `iColor`layers is for sbix color fonts and are documented here:
https://glyphsapp.com/tutorials/creating-an-apple-color-font

whereas `Color XX` is for COLR/CPAL or OT-SVG color fonts:
https://glyphsapp.com/tutorials/creating-a-microsoft-color-font
https://glyphsapp.com/tutorials/creating-an-svg-color-font

Especially considering that want to support building color fonts (https://github.com/googlefonts/fontmake/issues/30), we _do_ want to make sure we export such layers correctly.

[...]

PR #570, opened 2019-12-16
Rename glyph layers when duplicate

moyogo:
Glyph sub layers are sometimes duplicated in Glyhps.app source files.
For example if a user double-clicks on "Copy layer"  or for other reasons, a glyph "a" may have two glyph layers called "Dec 16 19, 12:00". Writing the UFO will fail when trying to write the glyph to "Dec 16 19, 12:00" twice.

This PR "solves" this by renaming the second layer, adding " #1", or " #n+1" .

### DIFF
--- a/Lib/glyphsLib/builder/layers.py
+++ b/Lib/glyphsLib/builder/layers.py
@@ -12,7 +12,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-
 from .constants import GLYPHS_PREFIX
 
 LAYER_ID_KEY = GLYPHS_PREFIX + "layerId"
@@ -26,6 +25,22 @@ def to_ufo_layer(self, glyph, layer):
         ufo_layer = ufo_font.layers.defaultLayer
     elif layer.name not in ufo_font.layers:
         ufo_layer = ufo_font.newLayer(layer.name)
+    elif layer.name in ufo_font.layers and glyph.name in ufo_font.layers[layer.name]:
+        self.logger.warning(
+            "%s %s: Glyph %s, layer %s: Duplicate glyph layer name"
+            % (
+                ufo_font.info.familyName,
+                ufo_font.info.styleName,
+                glyph.name,
+                layer.name,
+            )
+        )
+        n = 1
+        new_layer_name = layer.name
+        while new_layer_name in ufo_font.layers:
+            new_layer_name = layer.name + " #" + repr(n)
+            n += 1
+        ufo_layer = ufo_font.newLayer(new_layer_name)
     else:
         ufo_layer = ufo_font.layers[layer.name]
     if self.minimize_glyphs_diffs:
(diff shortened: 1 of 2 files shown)


Lib/glyphsLib/builder/layers.py
```diff
@@ -26,6 +29,19 @@ def to_ufo_layer(self, glyph, layer):
         ufo_layer = ufo_font.layers.defaultLayer
     elif layer.name not in ufo_font.layers:
         ufo_layer = ufo_font.newLayer(layer.name)
+    elif (layer.name in ufo_font.layers and
+            glyph.name in ufo_font.layers[layer.name]):
+        logger.warning(
```
anthrotype: I think builder has a logger attribute that you can call as `self.logger`

Lib/glyphsLib/builder/layers.py
```diff
@@ -26,6 +29,19 @@ def to_ufo_layer(self, glyph, layer):
...
+            "Glyph {}, layer {}: Duplicate glyph layer name".format(
+                glyph.name, layer.name
```
anthrotype: instead of using `.format` you can pass positional arguments to the logger.warning() call (using `%s` for string interpolation which is the default for logging). They will only be string-interpolated if the message is effectively logged (e.g. there's a handler, it is in level, etc.)

anthrotype:
you can fix the linter/formatting errors by running `tox -e lint`

moyogo:
Thanks!

> So, this means that when round-tripping back to .glyphs, these duplicate layers will have their name "fixed". Are we ok with that? I am.

Yes, I think that’s OK. Most of the time this is a mistake in the sources.

Lib/glyphsLib/builder/layers.py
```diff
@@ -26,6 +25,22 @@ def to_ufo_layer(self, glyph, layer):
         ufo_layer = ufo_font.layers.defaultLayer
     elif layer.name not in ufo_font.layers:
         ufo_layer = ufo_font.newLayer(layer.name)
+    elif layer.name in ufo_font.layers and glyph.name in ufo_font.layers[layer.name]:
+        self.logger.warning(
+            "%s %s: Glyph %s, layer %s: Duplicate glyph layer name"
+            % (
```
madig: Shouldn't this be simply `warning("text", arg1, ..., argN)` instead of `%`?

anthrotype: yes

anthrotype:
i'll fix the logging call

schriftgestalt:
> Yes, I think that’s OK. Most of the time this is a mistake in the sources. 
This it not true. It is more likely on purpose as described in #566. At least the suffix should be removed when the name starts with "Color" or "iColor".

anthrotype:
I think there are still issues when a duplicate glyph layer in turn has a background layer. The latter also needs to be de-duplicated, else this error occurs => https://github.com/googlefonts/fontmake/issues/630

arrowtype:
Just wanted to chime in ... this just came in clutch for me! Thanks so much, @moyogo

PR #571, opened 2019-12-17
write features to master UFOs _before_ applying master custom parameters

moyogo:
Otherwise, the 'Replace Feature' custom parameter will not work for masters, as it finds the features.fea text empty.


Lib/glyphsLib/builder/builders.py
```diff
@@ -241,7 +244,6 @@ def masters(self):
                 for source in self._sources.values():
                     source.font.lib["public.skipExportGlyphs"] = skip_export_glyphs
 
-        self.to_ufo_features()  # This depends on the glyphOrder key
```
madig: Why remove the comment?

anthrotype:
closing as superseded by https://github.com/googlefonts/glyphsLib/pull/640

PR #572, opened 2019-12-22
Use openstep-plist for parsing

madig:
This was surprisingly easy to hook up.

Tests fail because of kerning dict ordering issues.

### DIFF
--- a/Lib/glyphsLib/builder/glyph.py
+++ b/Lib/glyphsLib/builder/glyph.py
@@ -36,9 +36,15 @@ def to_ufo_glyph(self, ufo_glyph, layer, glyph):
     if note is not None:
         ufo_glyph.note = note
 
-    last_change = glyph.lastChange
-    if last_change is not None:
-        ufo_glyph.lib[GLYPHLIB_PREFIX + "lastChange"] = to_ufo_time(last_change)
+    # Optimization: profiling glyphs2ufo of NotoSans-MM.glyphs (6000 glyphs) on a Mac
+    # mini late 2014, Python 3.6.8, revealed that a whopping 17% of the time was spent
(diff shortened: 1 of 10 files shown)


madig:
Well, performance is somewhat disappointing. The time improvement for glyphs2ufo of Cantarell is ~0.4 seconds.

```
2.20user 0.15system 0:02.37elapsed 99%CPU (0avgtext+0avgdata 115128maxresident)k
0inputs+200outputs (0major+33912minor)pagefaults 0swaps

2.15user 0.17system 0:02.34elapsed 99%CPU (0avgtext+0avgdata 115032maxresident)k
0inputs+0outputs (0major+32406minor)pagefaults 0swaps

2.18user 0.13system 0:02.33elapsed 99%CPU (0avgtext+0avgdata 115056maxresident)k
0inputs+0outputs (0major+32417minor)pagefaults 0swaps
```

Zoomed in flame graph not including import time (10% of total time):
grafik

schriftgestalt:
The parsing is twice as fast as the "optimised" code. Not bad. 

I had a look at the `GSNode.read()` method. Using `split()` and some `if`s instead of regex made my 450kB test file import 8% faster. 
Screenshot
(This is not the nicest code, I know)

Another big chunk is spend in `parse_float_or_int`. I’m not sure how that could be optimised inside python.
For me it seems that the parsing is quite fast compared to the ufo writing.

madig:
Thanks for looking into this. I experimented with `split` as well but got no real speed-up :/ maybe need to look into that again.

schriftgestalt:
I can give you the code as a starting point ...
We you optimise the parser, I would only run the that in the profiler. Something like `GSFont(path)`. Then you better see the percentage of the improvements.

Issue #560, opened 2019-10-16
Changed glyphOrder behavior in Glyphs 2.6.x


madig:
We went out of our way to support Glyphs' handling of the glyphOrder custom parameter before: https://github.com/googlefonts/glyphsLib/blob/af5ab0b380ba9fc2100a0b91223d41a9c6bb0400/tests/builder/builder_test.py#L1357-L1415.

The behavior has changed has changed in 2.6.x:

When Glyphs opens a UFO with a `public.glyphOrder` key and...

1. ... no `com.schriftgestaltung.glyphOrder` key, it will copy `public.glyphOrder` _verbatim_ to the font's custom parameter `glyphOrder`, including non-existant glyphs. It will sort the glyphs ("Predefined Sorting") as specified by the font's custom parameter `glyphOrder`.
2. ... a `com.schriftgestaltung.glyphOrder` key set to a list of glyph names, it will copy `com.schriftgestaltung.glyphOrder` _verbatim_ to the font's custom parameter `glyphOrder`, including non-existant glyphs. It will _not reorder the glyphs_ and instead display them as specified in `public.glyphOrder`. If the glyphs aren't grouped by category, it may make repeated category groups (e.g. Separator: .notdef, Punctuation: period, Separator: nbspace).
3. ... a `com.schriftgestaltung.glyphOrder` key set to `False`, it will not copy `public.glyphOrder` at all and there is no font custom parameter `glyphOrder`. It will also not sort the glyphs and instead display them as specified in `public.glyphOrder`. Round-tripping back will therefore overwrite `public.glyphOrder` with the order of the .glyphs file.

When Glyphs 2.6.1 opens a UFO _without_ a `public.glyphOrder` key and...

1. ... no `com.schriftgestaltung.glyphOrder` key, it will sort the glyphs in the typical Glyphs way and not create a font custom parameter `glyphOrder`.
2. ... a `com.schriftgestaltung.glyphOrder` key set to a list of glyph names, it will copy `com.schriftgestaltung.glyphOrder` _verbatim_ to the font's custom parameter `glyphOrder`, including non-existant glyphs and will sort the glyphs ("Predefined Sorting") as specified by the font's custom parameter `glyphOrder`.
3. ... a `com.schriftgestaltung.glyphOrder` key set to `False`, it will sort the glyphs in the typical Glyphs way and not create a font custom parameter `glyphOrder`.

Now to figure out what to do.

madig:
Maybe the following:

[...]

PR #582, opened 2020-01-29
Copy glyphOrder lib key/custom parameter verbatim

madig:
Closes https://github.com/googlefonts/glyphsLib/issues/560.

The idea is to not be smart and copy `public.GlyphOrder` to the Font's custom parameters verbatim and back if it exists.

### DIFF
--- a/Lib/glyphsLib/builder/custom_params.py
+++ b/Lib/glyphsLib/builder/custom_params.py
@@ -617,21 +617,15 @@ class GlyphOrderParamHandler(AbstractParamHandler):
 
     def to_glyphs(self, glyphs, ufo):
         if glyphs.is_font():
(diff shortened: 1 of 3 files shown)


anthrotype:
if this is not urgent, I would prefer to cut a bug-fix release only with the #581 patch, and then we can review and merge this one with ease.

madig:
A glyphs2ufo2glyphs trip does not change the glyphOrder of NotoSans-MM.glyphs.

madig:
Did a test with a glyphs file with (NotoSans-MM.glyphs) and without (NotoSansOlChiki.glyphs) glyphOrder custom parameter. This PR does not change the `GlyphOrder` output.

anthrotype:
thanks for checking

PR #583, opened 2020-01-30
Cut down on noise

madig:
Closes https://github.com/googlefonts/glyphsLib/issues/584.

### DIFF
--- a/Lib/glyphsLib/builder/paths.py
+++ b/Lib/glyphsLib/builder/paths.py
@@ -25,13 +25,13 @@ def to_ufo_paths(self, ufo_glyph, layer):
         # the list is changed below, otherwise you can't draw more than once
         # per session.
         nodes = list(path.nodes)
-        for node in nodes:
-            self.to_ufo_node_user_data(ufo_glyph, node)
 
         pen.beginPath()
+
         if not nodes:
             pen.endPath()
             continue
+
         if not path.closed:
             node = nodes.pop(0)
             assert node.type == "line", "Open path starts with off-curve points"
@@ -40,11 +40,21 @@ def to_ufo_paths(self, ufo_glyph, layer):
             # In Glyphs.app, the starting node of a closed contour is always
             # stored at the end of the nodes list.
             nodes.insert(0, nodes.pop())
+
         for node in nodes:
             node_type = _to_ufo_node_type(node.type)
             pen.addPoint(
-                tuple(node.position), segmentType=node_type, smooth=node.smooth
+                tuple(node.position),
+                segmentType=node_type,
+                smooth=node.smooth,
+                name=node.userData.get("name"),
             )
+            # NOTE: Can't do path_index, node_index through enumeration here because we
+            # might have changed node order.
+            # A node's name will be stored as a UFO point's name attribute, so filter
+            # it from the Glyph node user data to avoid storing duplicate information.
+            node_user_data = {k: v for k, v in node.userData.items() if k != "name"}
+            self.to_ufo_node_user_data(ufo_glyph, node, node_user_data)
         pen.endPath()
 
 
--- a/Lib/glyphsLib/builder/user_data.py
+++ b/Lib/glyphsLib/builder/user_data.py
@@ -90,12 +90,11 @@ def to_ufo_layer_user_data(self, ufo_glyph, layer):
             ufo_glyph.lib[key] = user_data[key]
 
 
-def to_ufo_node_user_data(self, ufo_glyph, node):
-    user_data = node.userData
+def to_ufo_node_user_data(self, ufo_glyph, node, user_data: dict):
     if user_data:
         path_index, node_index = node._indices()
         key = f"{NODE_USER_DATA_KEY}.{path_index}.{node_index}"
-        ufo_glyph.lib[key] = dict(user_data)
+        ufo_glyph.lib[key] = user_data
 
 
 def to_glyphs_family_user_data_from_designspace(self):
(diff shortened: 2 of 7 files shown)


Lib/glyphsLib/builder/user_data.py
```diff
@@ -92,6 +92,8 @@ def to_ufo_layer_user_data(self, ufo_glyph, layer):
 
 def to_ufo_node_user_data(self, ufo_glyph, node):
     user_data = node.userData
+    if "name" in user_data:  # Handle point name at a higher level.
+        del user_data["name"]
```
belluzj: Why does this function modify the Glyphs object in "to_ufo"? It should only read from the Glyphs object

Lib/glyphsLib/builder/paths.py
```diff
@@ -40,11 +40,24 @@ def to_ufo_paths(self, ufo_glyph, layer):
             # In Glyphs.app, the starting node of a closed contour is always
             # stored at the end of the nodes list.
             nodes.insert(0, nodes.pop())
+
         for node in nodes:
             node_type = _to_ufo_node_type(node.type)
+            node_name = node.userData.get("name")
+            if node_name is not None:
+                # Without a proxy, one could `node.userData.pop("name", None)` instead.
+                del node.userData["name"]
```
belluzj: I still don't get why you're changing stuff in the Glyphs userData while in a "to_ufo..." function?

Lib/glyphsLib/builder/instances.py
```diff
@@ -108,19 +108,25 @@ def _to_designspace_instance(self, instance):
...
+        if styleMapStyleName != "regular":
+            ufo_instance.styleMapStyleName = styleMapStyleName
 
     ufo_instance.name = " ".join(
         (ufo_instance.familyName or "", ufo_instance.styleName or "")
     )
 
     if self.minimize_glyphs_diffs:
-        ufo_instance.lib[EXPORT_KEY] = instance.active
+        if instance.active is not True:
```
anthrotype: `instance.active is not True` -> `not instance.active`?

Lib/glyphsLib/builder/paths.py
madig: Because we write the Glyphs node user data to a glyph lib key. Glyph stores the node name in a node's user data, but the UFO spec bolts it onto the actual point, so we have to extract the name from user data and then delete it from there, so we don't write the node name to actual point AND lib key.

Lib/glyphsLib/builder/instances.py
```diff
@@ -108,19 +108,25 @@ def _to_designspace_instance(self, instance):
...
         ufo_instance.lib[WEIGHT_KEY] = instance.weight
         ufo_instance.lib[WIDTH_KEY] = instance.width
 
-        ufo_instance.lib[INSTANCE_INTERPOLATIONS_KEY] = instance.instanceInterpolations
-        ufo_instance.lib[MANUAL_INTERPOLATION_KEY] = instance.manualInterpolation
+        if instance.instanceInterpolations:
+            ufo_instance.lib[
+                INSTANCE_INTERPOLATIONS_KEY
+            ] = instance.instanceInterpolations
+        if instance.manualInterpolation is not False:
```
anthrotype: `is not False`? what's this?

Lib/glyphsLib/builder/paths.py
madig: Hrm. this does change the input file though. Grrr.

Lib/glyphsLib/builder/instances.py
madig: I think that's a 🧠💨. Something felt wrong but I couldn't place my finger on it.

Lib/glyphsLib/builder/user_data.py
```diff
@@ -167,8 +167,13 @@ def to_glyphs_layer_user_data(self, ufo_glyph, layer):
 
 def to_glyphs_node_user_data(self, ufo_glyph, node, path_index, node_index):
     key = f"{NODE_USER_DATA_KEY}.{path_index}.{node_index}"
+    # Remember node name because we overwrite userData, which internally stores the
+    # node name. We take the name directly from the UFO point in `to_glyphs_paths`.
+    old_name = node.name
     if key in ufo_glyph.lib:
         node.userData = ufo_glyph.lib[key]
+    if node.name is None:
+        node.name = old_name
```
anthrotype: i have no idea what's going on here. what does 'overwriting node.userData' have to do with node.name attribute?

madig: We do `node.userData = ufo_glyph.lib[key]`. `node.name` is sugar for `node.userData.get("name", None)`. Assigning something new to userData may change the node name then. Since we take the node name from the UFO's Point.name, we must not hold on to the node name before userData assignment.

anthrotype: > node.name is sugar for node.userData.get("name", None)

that's what doesn't make sense to me. why isn't it a regular attribute or read-only property stored in the node object itself?

Lib/glyphsLib/builder/paths.py
madig: The code now copies stuff instead of changing the input.

Lib/glyphsLib/builder/user_data.py
madig: Glyphs stores the name of a node in userData:

```
[...]
"123 456 OFFCURVE {name = \"aaa\";}"
[...]
```
Probably was the most straightforward thing to do.

madig:
I propose we also remove INSTANCE_INTERPOLATIONS_KEY completely, currently used in an instance's lib:

```XML
          <key>com.schriftgestaltung.intanceInterpolations</key>
          <dict>
            <key>56B0CDB6-C07B-476C-A69F-18FF782CB46F</key>
            <real>0.34091</real>
            <key>63F3AA79-E531-4BFB-93F0-25DD2830D098</key>
            <real>0.65909</real>
          </dict>
```

The fontmake ecosystem does not use this information and Glyphs doesn't care, because it computes the information internally anyway (right, @schriftgestalt ?).

schriftgestalt:
Please try to roundtrip stuff like that as good as possible. In combination with the key `manualInterpolation`, the values are actually used.

madig:
Which values, all instance lib keys?

madig:
Ok, I toned it down to avoid writing some empty values and do the right thing to node names.

PR #471, opened 2018-12-04
axes: fix get_regular_master

m4rc1e:
Calling font.customParameters["Variation Font Origin"] returns the master's id, not its name.

PR #587, opened 2020-02-06
[builder.axes] Add support for "Variable Font Origin" custom parameter, refactor to use master ID's instead of master names

chrissimpkins:
This PR adds support for the "Variable Font Origin" custom parameter in glyphs source files to address the issue in https://github.com/googlefonts/fontmake/issues/636.  

There is a discrepancy in the [Glyphs documentation](https://glyphsapp.com/tutorials/creating-a-variable-font) of the custom parameter name:

Prior to the commits here, the name "Variation Font Origin" was supported in the `builder.axes.get_regular_master` function.  As of the Glyphs editor Version 2.6.4 (1286) release, it was recognized that the name has been changed to "Variable Font Origin":

Furthermore, this custom parameter dictionary key yields a UUID string value, not the master name.  This required a fix in the `builder.axes.get_regular_master` function to support the correct return value.

This PR includes:

- refactored `builder.axes.get_regular_master` function to support "Variable Font Origin" custom parameter
- refactored `builder.axes.get_regular_master` function to continue to support a fallback of "Variation Font Origin" custom parameter in case that remains in use in previous versions of the editor/source spec
- refactored `builder.axes.get_regular_master` function to match against master UUID strings instead of master names
- three new glyphs source test files in the `tests/data` dir
- unit tests

---

Associated #471


chrissimpkins:
The changes here led to a test fail in a different module https://travis-ci.org/googlefonts/glyphsLib/jobs/646712148#L304

There is also a hard coded "Variation Font Origin" string in https://github.com/googlefonts/glyphsLib/blob/09e4cb6c07005fc46ebcfc88063ee0a90d6d504c/Lib/glyphsLib/builder/builders.py#L289-L297

that should likely be changed to the new name.

I'll have a look at these tomorrow

chrissimpkins:
Re: issues that arose in https://github.com/googlefonts/glyphsLib/pull/587#issuecomment-582733666


This has been marked as deprecated x 2 years.  OK to simply remove rather than try to refactor to support the new custom parameter name?

https://github.com/googlefonts/glyphsLib/blob/09e4cb6c07005fc46ebcfc88063ee0a90d6d504c/Lib/glyphsLib/builder/builders.py#L279-L281

anthrotype:
I think for backward compat with the old "Variation Font Origin" custom parameter, you should continue to match the master.name in the latter case, and use master.id for the new "Variable Font Origin".
I'm also open to simply drop support altogether for the old parameter, if the latest Glyphs.app doesn't support it any more (has it _ever_ supported it? no idea).

anthrotype:
I'm not sure what `instance_data` has to do with this custom parameter. But if neither glyphsLib nor fontmake use that any more, I'm fine to drop that (perhaps in a different PR, if it turns out to be unrelated).

anthrotype:
according to git grep, there is only one .glyphs file in the whole noto-source repository that use the old "Variation Font Origin" custom parameter (it's NotoSerifMyanmar-MM.glyphs), and it is using the master name ("Light") not the master UUID.
No noto source file uses the new "Variable Font Origin" custom parameter.

chrissimpkins:
> I think for backward compat with the old "Variation Font Origin" custom parameter, you should continue to match the master.name in the latter case

I don't know if it ever did.  See https://github.com/googlefonts/glyphsLib/pull/471#issuecomment-444228093

anthrotype:
ouch.. I wonder why we closed that PR :-/

I suggest you keep supporting the old parameter (matching the master name) but log a warning suggesting the user to rename it to the new "Variable Font Origin"

chrissimpkins:
> according to git grep, there is only one .glyphs file in the whole noto-source repository that use the old "Variation Font Origin" custom parameter (it's NotoSerifMyanmar-MM.glyphs), and it is using the master name ("Light") not the master UUID.
> No noto source file uses the new "Variable Font Origin" custom parameter.

Got it.  So it must have at some point in the past.

anthrotype:
> it must have at some point in the past.

not necessarily, since the noto sources were designed to be built using fontmake, not Glyphs.app. The two must have diverged at some point or maybe we got it wrong since the beginning.

chrissimpkins:
> I suggest you keep supporting the old parameter (matching the master name) but log a warning suggesting the user to rename it to the new "Variable Font Origin"

Will do!  How do you approach that through fontmake?  Raise exception here that is caught there?

anthrotype:
no no, just log a warning using `logging`. I think the glyphsLib UFOBuilder class has a log attribute somewhere, that holds an instance of a logging.Logger

chrissimpkins:
> no no, just log a warning using logging. I think the glyphsLib UFOBuilder class has a log attribute somewhere, that holds an instance of a logging.Logger

Got it! Will do

anthrotype:
it's called `self.logger` (see `_LoggerMixin` class in glyphsLib.builder.builders)

anthrotype:
Thanks for working on this!

chrissimpkins:
> I'm not sure what instance_data has to do with this custom parameter. But if neither glyphsLib nor fontmake use that any more, I'm fine to drop that (perhaps in a different PR, if it turns out to be unrelated).

Tests will continue to fail on this PR until the change is put in place.  There is a unit test based around it that is now failing with the changes here.

chrissimpkins:
where change = remove the method and test

anthrotype:
which test is failing

chrissimpkins:
https://travis-ci.org/googlefonts/glyphsLib/jobs/646712148#L304

anthrotype:
that `builder.instance_data` is returned from `glyphsLib.to_ufos`, we need to keep that for a while, until we decide to drop support for returning only the master UFOs without the designspace that connects them. Please for now try to keep it, thanks

chrissimpkins:
OK, let me have a look at the unit test and see if I can figure out why it is failing.  I broke something in that path.

anthrotype:
i think it's failing because the old param returned the master name, but you're only matching the master id.

Lib/glyphsLib/builder/axes.py
```diff
@@ -520,10 +520,19 @@ def get_regular_master(font):
...
+    # The current glyphs source specification supports the custom
+    # parameter name "Variable Font Origin".  This may have been
+    # named "Variation Font Origin" in the past.
+    # We support the current name with a fallback to the previous name
+    # if not found in the GSFont.customParameters dict
+    regular_id = None
+    if "Variable Font Origin" in font.customParameters:
+        regular_id = font.customParameters["Variable Font Origin"]
+    elif "Variation Font Origin" in font.customParameters:
+        regular_id = font.customParameters["Variation Font Origin"]
```
anthrotype: you should expect a master name, not a UUID, here. then return the first master with that name

chrissimpkins: https://github.com/googlefonts/glyphsLib/pull/587/commits/060c283df7ab91c42e8f3bc75605d7f21a09b1fd adds these changes and breaks the unit test

chrissimpkins: https://github.com/googlefonts/glyphsLib/pull/587/commits/71419da74e106682aa3a053671284fae87155ec5 fixes the unit test which requires a hand edit of the test glyphs source file.  Note that the test file is a hand edited glyphs source file that changes the value of the custom parameter to the master name when  "Variation Font Origin" is used as the custom parameter name.  Previously I made the assumption that this key had an ID value and simply hand edited the source file key name to "Variation Font Origin" from "Variable Font Origin" using a source file exported from Glyphs @ build 2.6.4 (1286)

chrissimpkins: and it looks like this fixes the other issue! :)

chrissimpkins:
Outstanding todo:

> no no, just log a warning using logging. I think the glyphsLib UFOBuilder class has a log attribute somewhere, that holds an instance of a logging.Logger

from https://github.com/googlefonts/glyphsLib/pull/587#issuecomment-583037553

chrissimpkins:
The logging warning was added in 367afc423afabb31c4e45d650dbb8c48fb0b201b.  I think that this is gtg on my end

Lib/glyphsLib/builder/axes.py
```diff
@@ -520,8 +520,18 @@ def get_regular_master(font):
...
+    # named "Variation Font Origin" in the past.
+    # We support the current name with a fallback to the previous name
+    # if not found in the GSFont.customParameters dict
+    if "Variable Font Origin" in font.customParameters:
+        regular_id = font.customParameters["Variable Font Origin"]
+        for master in font.masters:
+            if master.id == regular_id:
+                return master
+    elif "Variation Font Origin" in font.customParameters:
+        regular_name = font.customParameters["Variation Font Origin"]
```
anthrotype: I would check that the result from customParameters is not None (it's None when a requested parameter is missing), just in case the the attribute you're comparing the result with is also None and you get a false positive (not sure if master.name or master.id can ever be None, but better be safe).
Also you can issue the logging warning in here, after you found the old Variation Font Origin

tests/builder/axes_test.py
```diff
@@ -212,3 +215,51 @@ def test_mapping_is_same_regardless_of_axes_custom_parameter(ufo_module):
...
+    assert matched is True
+    default_master = get_regular_master(font)
+    assert default_master.name == "Regular Text"
+
+
+def test_custom_parameter_vfo_old_name():
+    """Tests get_regular_master when 'Variation Font Origin' custom parameter name
+    is used with master set to 'Regular Text'.  This custom parameter name is not
+    used in current releases of the Glyphs editor / glyphs source file specification."""
+    source_path = os.path.join("tests", "data", "CustomParameterVFO-oldname.glyphs")
```
anthrotype: I wonder if you could reduce the number of test files added to this PR. We don't need three of them probably, you can use one and modify the others (add or remove the custom parameter you need) before to call `get_default_master`

Lib/glyphsLib/builder/axes.py
chrissimpkins: 84077755fb867de058decbf5a06965ba132a9caf

tests/builder/axes_test.py
chrissimpkins: 3591e3f25624272ae5579482b6d43e20d201a4d8

chrissimpkins:
Should be gtg here @anthrotype. Let me know if there is anything else that you need.

anthrotype:
Thanks Chris! LGTM

chrissimpkins:
Thanks for all of your help here Cosimo!

PR #593, opened 2020-02-15
glyphdata: Handle ligatures with suffix

khaledhosny:
GlyphsLib fails to detect the category of glyphs like `dal_alef-ar` that GlyphsApp correctly classifies as Letter, Ligature.

### DIFF
--- a/Lib/glyphsLib/glyphdata.py
+++ b/Lib/glyphsLib/glyphdata.py
@@ -187,6 +187,12 @@ def _construct_category(glyph_name, data):
     # Detect ligatures.
     if "_" in base_name:
         base_names = base_name.split("_")
+        # The last name has a suffix, add it to all the names.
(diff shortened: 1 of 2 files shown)


madig:
Argh. The detection code is a bit of a clusterfuck. @schriftgestalt how does Glyphs detect this?

schriftgestalt:
With a lot of clusterfuck code.

madig:
Well, let's merge this and see what breaks in production.

@schriftgestalt I wonder if there is a solid scheme that can be mechanically verified by a proof assistant or something. That'd be the end of clusterfucking.

anthrotype:
LGTM

schriftgestalt:
I have a ton of unit test so rewriting that would be possible without breaking to much. I tried a few time …

Issue #585, opened 2020-01-30
Support "Axes" custom parameter 


madig:
Not sure what exactly it does, but I suppose ufo2glyphs can dump the axes from the Designspace into the parameter. What to do for glyphs2ufo? Ignore?

schriftgestalt:
The `Axes` stores the main axes definition:
```
(
	{
		Name = Weight;
		Tag = wght;
	},
	{
		Hidden = 1;
		Name = "Some Name";
		Tag = SMNA;
	}
);
```

anthrotype:
but we _do_ support this one already, don't we?

madig:
@schriftgestalt thanks.

@anthrotype yes but we don't put it into the Glyphs file on ufo2glyphs.

madig:
@schriftgestalt How does `GSFontMaster.axes` work? How does it know how many axes there actually are?

schriftgestalt:
You mean the property in the python wrapper?

https://github.com/schriftgestalt/GlyphsSDK/blob/95ba997589ffc62cb3fbcefbe990661d5e133071/ObjectWrapper/GlyphsApp/__init__.py#L1633

madig:
Oh, interesting. So, how does `countOfAxes()` determine the number of axes? And there's a hard limit of 6 axes?

schriftgestalt:
```
- (NSUInteger)countOfAxes {
	return [[self axes] count];
}
```
and `[self axes]` is accessing the custom parameter.

I added an error when one tries to add more then six axes.

madig:
But how does `axes` know how many axes there are? :grin: Does it check the 6 different master values on every master to see if any of them diverge from the default value?

schriftgestalt:
Axes checks there is a custom parameter. If not, it checks if any master has a none default setting for any of the `widthValue`, `weightValue`,  `customValue`, …

madig:
What does it do when all masters have just `customValue3` set to something non-default?

khaledhosny:
Done a while ago alongside Glyphs3 support.

Issue #568, opened 2019-12-09
Support "Axis Mapping" custom parameter for GlyphsApp fonts


arrowtype:
GlyphsApp has a Custom Property for "Axis Mappings" (it [appears to have been added in v1261](https://forum.glyphsapp.com/t/avar-axis-mappings/12425)).

However, adding this has no effect on the Designspace output from glyphsLib, as in a FontMake build.

![image](https://user-images.githubusercontent.com/45946693/70474911-6d0f5680-1aa1-11ea-92a1-a195cef79918.png)

The only way to get around this as a user, as far as I know, is to have `Axis Location` custom props on Instances, _but not_ on Masters. I get this impression from the `builder/axes` code, and a test file only got the proper designspace mapping once I did removed `axis location` props from Masters.

https://github.com/googlefonts/glyphsLib/blob/264afc2a546bdf72165787562bbbd02c41da5e52/Lib/glyphsLib/builder/axes.py#L137-L215

_However,_ it's problematic to not be able to control the axis maps at a single, top level. In the Nunito project, it means that I also had to add a random extra Instance, so that the whole Slant range would be covered by Instances, and the Slant axis would go from the min to max values present in sources (otherwise, the build fails).

Here are the files I was using:

[nunito-build-test.zip](https://github.com/googlefonts/glyphsLib/files/3941809/nunito-build-test.zip)

`NunitoSans_Pitstop-26.glyphs` is the full GlyphsApp file, with the problem of only-partial Weight axis mapping. With this, the generated wght map is partial (resulting in instances with weight values that are a bit off, with floating-point values):

```XML
    <axis tag="wght" name="Weight" minimum="200" maximum="1000" default="200">
      <map input="200" output="42"/>
      <map input="700" output="125"/>
      <map input="1000" output="208"/>
    </axis>
```

`NunitoSans_Pitstop-26--testfile_ABC.glyphs` is a GlyphsApp file with the problem fixed as a hack, but with the superfluous Instance to make the Slant axis work. With this, the generated map is:

```XML
    <axis tag="wght" name="Weight" minimum="200" maximum="1000" default="200">
      <map input="200" output="42"/>
      <map input="300" output="61"/>
      <map input="400" output="81"/>
      <map input="600" output="101"/>
      <map input="700" output="125"/>
      <map input="800" output="151"/>
      <map input="900" output="178"/>
      <map input="1000" output="208"/>
    </axis>
```

I'm building with FontMake:

```Bash
[...]

PR #594, opened 2020-02-17
Implement GSFont.axes

madig:
This PR goes towards supporting the Axes font-level parameter (https://github.com/googlefonts/glyphsLib/issues/585). I want to use it later for determining what axes to write to the Designspace.

### DIFF
--- a/tests/classes_test.py
+++ b/tests/classes_test.py
@@ -183,6 +183,76 @@ class GSFontTest(unittest.TestCase):
         font.masters.append(master)
         self.assertEqual(master.font, font)
 
+    def test_axes_default(self):
+        font = GSFont()
+        self.assertEqual(font.axes, [{"Name": "Weight", "Tag": "wght"}])
+
+    def test_axes_more_axes(self):
+        font = GSFont()
+        master = GSFontMaster()
+        font.masters.append(master)
+        master2 = GSFontMaster()
+        font.masters.append(master2)
+        self.assertEqual(font.axes, [{"Name": "Weight", "Tag": "wght"}])
+
+        master2.widthValue = 1
+        self.assertEqual(
+            font.axes,
+            [{"Name": "Weight", "Tag": "wght"}, {"Name": "Width", "Tag": "wdth"}],
+        )
+
+        master.customValue3 = 1
+        self.assertEqual(
+            font.axes,
+            [
+                {"Name": "Weight", "Tag": "wght"},
+                {"Name": "Width", "Tag": "wdth"},
+                {"Name": "Custom", "Tag": "CUS1"},
+                {"Name": "Custom2", "Tag": "CUS2"},
+                {"Name": "Custom3", "Tag": "CUS3"},
+                {"Name": "Custom4", "Tag": "CUS4"},
+            ],
+        )
+
+        font.customParameters["Axes"] = []
+        self.assertEqual(
+            font.axes,
+            [
+                {"Name": "Weight", "Tag": "wght"},
+                {"Name": "Width", "Tag": "wdth"},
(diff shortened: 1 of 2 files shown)


tests/classes_test.py
```diff
@@ -183,6 +183,74 @@ def test_font_master_proxy(self):
...
+    def test_axes_parameter(self):
+        font = GSFont()
+        font.customParameters["Axes"] = [
+            {"Name": "Weight", "Tag": "wght"},
+            {"Name": "Empty Name", "Tag": "EMPT"},
+            {"Name": "Empty Name", "Tag": "EMPT"},
+            {"Name": "Empty Name", "Tag": "EMPT"},
+            {"Name": "Empty Name", "Tag": "EMPT"},
+            {"Name": "Empty Name", "Tag": "EMPT"},
+            {"Name": "Empty Name", "Tag": "EMPT"},
```
belluzj: What are these?

madig: The default strings if you repeatedly add axes in the "Axes" dialog but don't change them.

belluzj: Hm ok... can you write it down as a comment?

madig:
Oops, I must also implement `GSFontMaster.axes`.

khaledhosny:
Can we have [GSInstance.axes](https://docu.glyphsapp.com/#GSInstance.axes) as well?

madig:
Yes, but this PR has been called into question by my co-maintainers. I was basically trying to solve axis handling with a clean approach but am now stalled by https://github.com/googlefonts/glyphsLib/issues/568#issuecomment-588256106.

khaledhosny:
For my use case, I want whatever GlyphsApp would return for these properties, and looking into https://github.com/schriftgestalt/GlyphsSDK/blob/master/ObjectWrapper/GlyphsApp/__init__.py#L1633-L1753, I think what this PR is doing is pretty close. The complicated part is getting the font axes, but may be a solution that works reliably only when `Axes` custom parameter is set is better than nothing?

madig:
If you define the Axes parameter, `GSFont.axes` returns it unchanged. You can use its' `len` to decide which master and instance values to consult.

khaledhosny:
I do this already. It would be slightly nicer if I didn’t have to replicate this logic everytime I use glyphsLib (I’m using glyphsLib and fontTools to build fonts and bypass the UFO conversion done by fontmake).

madig:
Oooh, do you have a glyphs2ft library?

khaledhosny:
Not a library, some 400 lines of ugly project specific code, but it gives me total control instead of fighting both fontmake and Glyphs into doing the super weird things I want to build.

madig:
I guess I can add a GSInstance.axes once my co-maintainers approve this one. Since I'm not sure which direction this lib is going to take in the future, I don't want to add Glyphs APIs I may not want later...

schriftgestalt:
The Glyphs python API has GSFont.axes, GSFontMaster.axes and GSInstance.axes for some time now. Each returns a list for floats.

khaledhosny:
Is there anything blocking the PR?

madig:
Me not knowing if we want this. What was I even trying to solve?

schriftgestalt:
It can be useful if glyphsLib is used for other things then just a gateway to ufo.

khaledhosny:
Like I said, I do have a need for it so I don’t see the issue in merging it. Though the focus for GlyphLib is conversion to UFO, I often use it to read and manipulate Glyphs files without going through UFO and I don’t see why supporting convenient APIs would be something undesirable.

khaledhosny:
Both GSFont and  GSInstance got axes attribute in the Glyphs 3 support work.

PR #595, opened 2020-02-19
Keep Axis axes

madig:
Add yet another heuristic to prevent a glyphs2ufo edge case where you have a single master at the default location and end up with a Designspace with no axes. If you have an Axes parameter, try to put the axes mentioned in there into the Designspace.

### DIFF
--- a/Lib/glyphsLib/builder/axes.py
+++ b/Lib/glyphsLib/builder/axes.py
@@ -146,6 +146,19 @@ def to_designspace_axes(self):
         axis.name = axis_def.name
         # TODO add support for localised axis.labelNames when Glyphs.app does
 
+        # NOTE: The following is a hack because this entire method is.
+        # get_axis_definitions may or may not generate axes that are in the font,
+        # so the bottom `if` statement filters out axes that it guesses don't actually
+        # exist. This results in the edge case where you have a single master at the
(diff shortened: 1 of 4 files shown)


tests/builder/axes_test.py
```diff
@@ -17,6 +17,7 @@
 
 import pytest
 
+import glyphsLib
```
khaledhosny: `GSFont` is already imported below.

Lib/glyphsLib/builder/axes.py
```diff
@@ -146,6 +146,19 @@ def to_designspace_axes(self):
         axis.name = axis_def.name
         # TODO add support for localised axis.labelNames when Glyphs.app does
 
+        # NOTE: The following is a hack because this entire method is.
+        # get_axis_definitions may or may not generate axes that are in the font,
```
anthrotype: > get_axis_definitions may or may not generate axes that are in the font

I don't understand this sentence.

Maybe simplify the comment like this:

```
make sure here that if there is an Axes parameter, we actually really keep axes that are defined
in there, even if they do nothing. This prevents an edge case where you have a single master at
the default location that ends up with a designspace with no axes, as the were all filtered out by
the if statement below. 
```

PR #608, opened 2020-04-28
Add pen and draw

### LINKED ISSUE #601: Add a drawPoints/draw methods to GSLayer to support FontTools pens?
Sometime one wants to work with paths in a generic way, regardless of font format, and the FontTools Pen protocol offers a nice way to abstract away from the details of TrueType, CFF, UFO, etc.
The `ttLib._TTGlyph` or defcon or ufoLib2's `Glyph` are examples.

It would be nice if GSLayer object also had something similar. 
The `glyphsLib.buider.glyph` and `.components` modules already have code to translate from GSLayer.paths and components to equivalent calls to PointPen methods. But they also contain some glyphsLib-specific code to deal with extra pieces of data that don't translate across the pen protocol.

The idea is to extract this pen-specific handling into a GSLayer.drawPoints method (and have an accompanying "draw" method that internally uses the adapter PointToSegmentPen to handle segment-oriented pens).

I drop this gist here in case anybody want to take this up:
https://gist.github.com/anthrotype/0f70ebc2143dfb01c25c47121e7e6814

madig:
In preparation for decomposition of non-default layer components decomposition, implement a pen and draw methods.

Closes #601.

### DIFF
--- a/Lib/glyphsLib/classes.py
+++ b/Lib/glyphsLib/classes.py
@@ -3868,8 +3868,15 @@ class LayerPointPen(AbstractPointPen):
         self._path.nodes.append(node)
 
     def addComponent(
-        self, baseGlyph: str, transformation: Transform, **kwargs: Any,
+        self,
+        baseGlyph: str,
+        transformation: Union[
+            Transform, Tuple[float, float, float, float, float, float]
+        ],
+        **kwargs: Any,
     ) -> None:
+        if not isinstance(transformation, Transform):
+            transformation = Transform(*transformation)
         component = GSComponent(baseGlyph, transform=transformation)
         self._layer.components.append(component)
 
--- a/Lib/glyphsLib/classes.py
+++ b/Lib/glyphsLib/classes.py
@@ -3813,7 +3813,7 @@ Number = Union[int, float]
 
 
 class LayerPointPen(AbstractPointPen):
-    """A point pen.
+    """A point pen to draw onto GSLayer object.
 
     See :mod:`fontTools.pens.basePen` and :mod:`fontTools.pens.pointPen` for an
     introduction to pens.
--- a/Lib/glyphsLib/classes.py
+++ b/Lib/glyphsLib/classes.py
@@ -3825,6 +3825,9 @@ class LayerPointPen(AbstractPointPen):
         self._path: Optional[GSPath] = None
 
     def beginPath(self, **kwargs: Any) -> None:
+        if self._path is not None:
+            raise ValueError("Call endPath first.")
+
         self._path = GSPath()
         self._path.closed = True  # Until proven otherwise.
 
--- a/Lib/glyphsLib/classes.py
+++ b/Lib/glyphsLib/classes.py
@@ -21,6 +21,14 @@ import re
 import uuid
 from collections import OrderedDict
 from io import StringIO
+from typing import Any, Optional, Tuple, Union
+
+from fontTools.pens.basePen import AbstractPen
+from fontTools.pens.pointPen import (
+    AbstractPointPen,
+    PointToSegmentPen,
+    SegmentToPointPen,
+)
 
 import glyphsLib
 from glyphsLib.affine import Affine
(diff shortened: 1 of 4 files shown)


Lib/glyphsLib/classes.py
```diff
@@ -1929,6 +1937,47 @@ def applyTransform(self, transformationMatrix):
             node.position.x = x
             node.position.y = y
 
+    def draw(self, pen: AbstractPen) -> None:
+        """Draws contour into given pen."""
```
anthrotype, line 1941:
```suggestion
        """Draws contour with the given pen."""
```

Lib/glyphsLib/classes.py
```diff
@@ -1929,6 +1937,47 @@ def applyTransform(self, transformationMatrix):
...
+        pointPen = PointToSegmentPen(pen)
+        self.drawPoints(pointPen)
+
+    def drawPoints(self, pointPen: AbstractPointPen) -> None:
+        """Draws points of contour into given point pen."""
```
anthrotype, line 1946:
```suggestion
        """Draws points of contour with the given point pen."""
```

Lib/glyphsLib/classes.py
```diff
@@ -2213,6 +2262,15 @@ def bounds(self):
     #     lambda self: self.piece,
     #     lambda self, value: setattr(self, "piece", value))
 
+    def draw(self, pen: AbstractPen) -> None:
+        """Draws component with given pen."""
+        pointPen = PointToSegmentPen(pen)
+        self.drawPoints(pointPen)
```
anthrotype, line 2268: segment pens also have a similar `addComponent` method, you can call that directly here I guess

Lib/glyphsLib/classes.py
```diff
@@ -3188,6 +3246,28 @@ def foreground(self):
...
+        pen = SegmentToPointPen(self.getPointPen())
+        return pen
+
+    def getPointPen(self) -> AbstractPointPen:
+        """Returns a point pen for others to draw points into self."""
+        pointPen = LayerPointPen(self)
+        return pointPen
+
+    def draw(self, pen: AbstractPen) -> None:
+        """Draws glyph into given pen."""
```
anthrotype, line 3260: > 
```suggestion
        """Draws glyph with the given pen."""
```

Lib/glyphsLib/classes.py
```diff
@@ -3188,6 +3246,28 @@ def foreground(self):
...
+        pointPen = PointToSegmentPen(pen)
+        self.drawPoints(pointPen)
+
+    def drawPoints(self, pointPen: AbstractPointPen) -> None:
+        """Draws points of glyph into given point pen."""
```
anthrotype, line 3265:
```suggestion
        """Draws points of glyph with the given point pen."""
```

Lib/glyphsLib/classes.py
```diff
@@ -3727,3 +3807,65 @@ def removeKerningForPair(self, fontMasterId, leftKey, rightKey, direction=LTR):
             del self._kerning[fontMasterId][leftKey]
         if not self._kerning[fontMasterId]:
             del self._kerning[fontMasterId]
+
+
+Number = Union[int, float]
+
+
+class LayerPointPen(AbstractPointPen):
+    """A point pen.
```
anthrotype, line 3816:
```suggestion
    """A point pen to draw onto GSLayer object.
```

Lib/glyphsLib/classes.py
```diff
@@ -3727,3 +3807,65 @@ def removeKerningForPair(self, fontMasterId, leftKey, rightKey, direction=LTR):
...
+        self._path: Optional[GSPath] = None
+
+    def beginPath(self, **kwargs: Any) -> None:
+        self._path = GSPath()
+
+    def endPath(self) -> None:
+        if self._path is None:
+            raise ValueError("Call beginPath first.")
+
+        self._path.closed = True
```
anthrotype, line 3835: endPath method does not a close a path. In pointPens, the open paths are marked by the first point of a contour having the "move" type.

Lib/glyphsLib/classes.py
```diff
@@ -3727,3 +3806,77 @@ def removeKerningForPair(self, fontMasterId, leftKey, rightKey, direction=LTR):
             del self._kerning[fontMasterId][leftKey]
         if not self._kerning[fontMasterId]:
             del self._kerning[fontMasterId]
+
+
+Number = Union[int, float]
+
+
+class LayerPointPen(AbstractPointPen):
```
anthrotype, line 3814: how about we move this to a glyphsLib.pens module?

anthrotype: nevermind, we can do that if/when we have more than one pen

Lib/glyphsLib/classes.py
```diff
@@ -3727,3 +3806,77 @@ def removeKerningForPair(self, fontMasterId, leftKey, rightKey, direction=LTR):
...
+            nodetype=_to_glyphs_node_type(segmentType),
+            smooth=smooth,
+            name=name,
+        )
+        if userData:
+            node.userData = userData
+        self._path.nodes.append(node)
+
+    def addComponent(
+        self, baseGlyph: str, transformation: Transform, **kwargs: Any,
```
anthrotype, line 3871: transformation should be a generic 6-tuple in the pointPen protocol, you can turn that to glyphsLib's own Transform class inside the method before initialising GSComponent

khaledhosny:
Thank you!

Lib/glyphsLib/classes.py
```diff
@@ -3727,3 +3806,84 @@ def removeKerningForPair(self, fontMasterId, leftKey, rightKey, direction=LTR):
             del self._kerning[fontMasterId][leftKey]
         if not self._kerning[fontMasterId]:
             del self._kerning[fontMasterId]
+
+
+Number = Union[int, float]
```
madig, line 3811: Argh, need to remove this...

PR #609, opened 2020-05-01
Set a background layer's foreground and parent during parsing

madig:


madig:
Oh, hm. I think I should also set `parent` on the background layer.

Lib/glyphsLib/classes.py
```diff
@@ -3448,6 +3459,8 @@ def shouldWriteValueForKey(self, key):
     def _setupLayer(self, layer, key):
         assert isinstance(key, str)
         layer.parent = self
+        if layer.hasBackground:
+            layer._background.parent = self
```
belluzj, line 3463: Is this still needed given the other changes?

madig: This is the change that actually makes tests pass.

Issue #449, opened 2018-10-24
components in background


schriftgestalt:
I have a .glyphs file that has a component in the background. It points to the same glyph. So you see whatever is in the foreground. When exporting that font with fontmake, defcon is trapped in a recursion. 

I hope that I have the latest version of all libs

```
  File "/usr/local/bin/fontmake", line 11, in <module>
    sys.exit(main())
  File "/Library/Python/2.7/site-packages/fontmake/__main__.py", line 258, in main
    project.run_from_glyphs(glyphs_path, **args)
  File "/Library/Python/2.7/site-packages/fontmake/font_project.py", line 560, in run_from_glyphs
    mti_source=mti_source)
  File "/Library/Python/2.7/site-packages/fontTools/misc/loggingTools.py", line 372, in wrapper
    return func(*args, **kwds)
  File "/Library/Python/2.7/site-packages/fontmake/font_project.py", line 131, in build_master_ufos
    font, family_name=family_name, instance_dir=instance_dir)
  File "/Library/Python/2.7/site-packages/glyphsLib/builder/__init__.py", line 104, in to_designspace
    return builder.designspace
  File "/Library/Python/2.7/site-packages/glyphsLib/builder/builders.py", line 210, in designspace
    list(self.masters)  # Make sure that the UFOs are built
  File "/Library/Python/2.7/site-packages/glyphsLib/builder/builders.py", line 157, in masters
    self.to_ufo_glyph(ufo_glyph, layer, glyph)
  File "/Library/Python/2.7/site-packages/glyphsLib/builder/glyph.py", line 116, in to_ufo_glyph
    self.to_ufo_glyph_background(ufo_glyph, layer)
  File "/Library/Python/2.7/site-packages/glyphsLib/builder/glyph.py", line 253, in to_ufo_glyph_background
    self.to_ufo_components(new_glyph, background)
  File "/Library/Python/2.7/site-packages/glyphsLib/builder/components.py", line 28, in to_ufo_components
    pen.addComponent(component.name, component.transform)
  File "/Library/Python/2.7/site-packages/defcon/pens/glyphObjectPointPen.py", line 37, in addComponent
    self._glyph.appendComponent(component)
  File "/Library/Python/2.7/site-packages/defcon/objects/glyph.py", line 708, in appendComponent
    self.insertComponent(len(self._components), component)
  File "/Library/Python/2.7/site-packages/defcon/objects/glyph.py", line 732, in insertComponent
[...]

Issue #375, opened 2018-06-19
Recursive glyph references choke glyphsLib


madig:
(Filing this here to take it out of my inbox)

Jany found a file with the following properties:

* parenleft is a glyph with outlines, and in the background it has bracketleft as a component
* bracketleft is a glyph with outlines, and in the background it has parenleft as a component

> The situation is valid but glyphsLib chokes on it (`RecursionError: maximum recursion depth exceeded while calling a python object.`) because for some reason it creates a component loop in the same layer... to investigate further

belluzj:
It's a fringe use-case but from what I understand the problem is this: in Glyphs, when you put a component in the background, the component points to the foreground of the other glyph; while in UFOs, if you have a component in a layer (including the background layer), it points to the contents of the other glyph from the same layer. That's why it makes a reference loop in UFOs but works fine in Glyphs.

anthrotype:
Aargh.. yes, that’s how UFO3 layers are designed to work. could we detect this cases and decompose to avoid the recursion?

moyogo:
Or we push for an update of the UFO spec: [Components could have a baseLayer attribute to allow components from other layers #49](https://github.com/unified-font-object/ufo-spec/issues/49).

belluzj:
Yes we could, but it loses the point of having those in Glyphs, which is to always have a synced visual of the parenleft in the background of the bracketleft for comparison purposes. We would have to "compose back" when going back to Glyphs to achieve something useful.

schriftgestalt:
One way to avoid the problem a bit is to not export the background when called by fontmake. The ufos are only used to produce binary fonts and those do not use the background. So an option for the builder might help?

madig:
I suppose that could in some instances also save time :thinking: I imagine that this might interfere with e.g. color fonts that need the layers, but I don't know if they are handled at all.

schriftgestalt:
Color fonts have multiple layers, that has nothing to do with the background.

madig:
Just hit the case where a glyph had a component referencing the glyph. glyphsLib should detect this and error out early.

arrowtype:
This was a (temporary) blocker on Merriweather, this week.

Glyphs like /p, /d, /q, & /b were referencing one another in background layers. I had to dive into defcon code and put in some print() statements to figure out what kept crashing the build.

I see three options for how this might be handled:

[...]

PR #610, opened 2020-05-04
Decompose background layers

madig:
This PR decomposes components used in background layers when going glyphs2ufo.

The UFO v3 and Glyphs data model have incompatible component reference semantics. UFO components always point to a glyph in the same layer, Glyphs components in a ...:

- master layer: point to glyphs in the same master layer, like in the UFO format.
- non-master layer: point to glyphs in the layer with the same layer name and fall back to glyphs in the associated master layer

Limit to background layers for now because functional layers like bracket and brace layers can contain components, too, and need special but too time-consuming handling. Non-functional layers like the copies you can make in Glyphs' UI with the "+" in the layer pane are likewise ignored. Bracket layers are already flattened in UFOs.

Closes https://github.com/googlefonts/glyphsLib/issues/449.
Partially tackles https://github.com/googlefonts/glyphsLib/issues/375.

### DIFF
--- a/Lib/glyphsLib/builder/components.py
+++ b/Lib/glyphsLib/builder/components.py
@@ -96,8 +96,6 @@ def to_ufo_components_background_decompose(self, ufo_glyph, layer):
                 for l in g.layers
                 if l.name == layer.name
             }
-            # FIXME: Have to use layer ID instead of name because glyphsLib collates
-            # layers with the same name.
             layers_master = {
                 g.name: l
                 for g in layer.parent.parent.glyphs
--- a/Lib/glyphsLib/builder/components.py
+++ b/Lib/glyphsLib/builder/components.py
@@ -15,6 +15,7 @@
 
 import logging
 
+from glyphsLib.classes import GSBackgroundLayer, LayerDecomposingPen
 from glyphsLib.types import Transform
 
 from .constants import GLYPHS_PREFIX, COMPONENT_INFO_KEY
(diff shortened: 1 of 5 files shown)


Lib/glyphsLib/builder/components.py
```diff
@@ -51,6 +69,50 @@ def to_ufo_components(self, ufo_glyph, layer):
             ufo_glyph.lib[_lib_key(key)] = values
 
 
+def to_ufo_components_decompose(self, ufo_glyph, layer):
+    """Draw decomposed .glyphs components with a pen, adding them to the parent
+    glyph."""
+
+    layer_id = layer.foreground.layerId
```
anthrotype, line 76: there is no mention in the method name or docstring that this applies only to background layers, but here you are accessing a property which is only present on GSBackgroundLayer objects. Would make sense to rename `to_ufo_background_components_decompose` or similar

Lib/glyphsLib/builder/components.py
```diff
@@ -51,6 +69,50 @@ def to_ufo_components(self, ufo_glyph, layer):
...
+        else:
+            # Is a non-master layer.
+            layers_nonmaster = {
+                g.name: l
+                for g in layer.parent.parent.glyphs
+                for l in g.layers
+                if l.name == layer.name
+            }
+            # FIXME: Have to use layer ID instead of name because glyphsLib collates
+            # layers with the same name.
```
anthrotype, line 99: WDYM by "glyphsLib collates layers with the same name"?

Lib/glyphsLib/builder/components.py
```diff
@@ -51,6 +69,50 @@ def to_ufo_components(self, ufo_glyph, layer):
...
+                for l in g.layers
+                if l.layerId == layer_master_id
+            }
+            layers = self._glyph_sets[layer_id] = {
+                **layers_master,
+                **layers_nonmaster,
+            }
+
+    pen = ufo_glyph.getPen()
+    dpen = LayerDecomposingPen(pen, layers)
```
anthrotype, line 112: You could use the `DecomposingRecordingPen` from fontTools to gather the simple outlines and then call its `replay` method with the `ufo_glyph.getPen()` instead of making your own `LayerDecomposingPen`

Lib/glyphsLib/builder/components.py
madig: If you have a regular layer in Glyphs and duplicate it twice, `len(Font.glyphs["A"].layers) == 3` but it's `1` in glyphsLib. Some layer wins I guess. This is significant because layers are looked up by names and the first one wins, whcih would be different in glyphsLib.

madig: Oh, it was a source error.

madig:
Should non-functional layers like the copies you can make with "+" in Glyphs' UI also be decomposed? The same thing that applies to backgrounds also applies to them.

anthrotype:
> because functional layers like bracket and brace layers can contain components, too, and need special but too time-consuming handling.

can you elaborate?

madig:
Georg said that bracket and brace layers can be used to adjust components, so decomposing them will break stuff. Bracket layers are already flattened, so they should be fine; brace layers I don't know.

belluzj:
> Should non-functional layers like the copies you can make with "+" in Glyphs' UI also be decomposed? The same thing that applies to backgrounds also applies to them.

I think we should only care about background layers, for the reason that it has happened several times that glyphsLib crashes (in defcon times) or outputs invalid UFOs because designers in Glyphs.app are putting the square bracket in the background of the parenthesis and same thing other way round, creating a loop in the UFO. So the first goal behind the work of this PR is to prevent this background loop situation from happening again.

On the other hand, I don't think I've ever heard a problem created by one of those "non functional layers" as Nikolaus calls them, and also Nikolaus says that it's complicated to handle, so I would say: leave the situation as it is and wait for a concrete problem to solve regarding those other layers.

Issue #376, opened 2018-06-19
Hints: handle up and down targets


madig:
It's apparently easy to accidentally add hints. One glyph contained in its lib field:

```
{'com.schriftgestaltung.Glyphs.glyph.rightMetricsKey': '=|', 
'com.schriftgestaltung.hints': [{'horizontal': True, 'options': 0, 'origin': [0, 2], 'stem': -2, 'target': ['d', 'o', 'w', 'n'], 'type': 'TopGhost'}], 
'public.markColor': '0.04,0.57,0.04,1'}
```

glyphsLib output:
```
Traceback (most recent call last):
  ...
  File "/Users/.../venv/lib/python3.6/site-packages/glyphsLib/builder/__init__.py", line 115, in to_glyphs
    return builder.font
  File "/Users/.../venv/lib/python3.6/site-packages/glyphsLib/builder/builders.py", line 343, in font
    self.to_glyphs_glyph(glyph, layer, master)
  File "/Users/.../venv/lib/python3.6/site-packages/glyphsLib/builder/glyph.py", line 216, in to_glyphs_glyph
    self.to_glyphs_hints(ufo_glyph, layer)
  File "/Users/.../venv/lib/python3.6/site-packages/glyphsLib/builder/hints.py", line 57, in to_glyphs_hints
    value = Point(*hint[attr])
TypeError: __init__() takes from 1 to 4 positional arguments but 5 were given
```

PR #613, opened 2020-05-29
hints: target="up" or "down"

moyogo:
See *Hints: handle up and down targets* https://github.com/googlefonts/glyphsLib/issues/376

### DIFF
--- a/Lib/glyphsLib/builder/hints.py
+++ b/Lib/glyphsLib/builder/hints.py
@@ -51,6 +51,7 @@ def to_glyphs_hints(self, ufo_glyph, layer):
             setattr(hi, attr, hint[attr])
         for attr in ["origin", "other1", "other2", "place", "scale", "target"]:
             if attr in hint:
+                # https://github.com/googlefonts/glyphsLib/pull/613
                 # handle target = ['u', 'p'] or ['d', 'o', 'w', 'n']
                 if attr == "target" and hint[attr] in (list("down"), list("up")):
                     value = "".join(hint[attr])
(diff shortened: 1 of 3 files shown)


moyogo:
@madig do we need to check target="down" or target="up" only occurs with type="TopGhost" or "BottomGhost" respectively?

madig:
No idea. @schriftgestalt ?

schriftgestalt:
> target="down" or target="up" only occurs with type="TopGhost" or "BottomGhost"

This is true. Is this answering the question?

moyogo:
>>    target="down" or target="up" only occurs with type="TopGhost" or "BottomGhost"

> This is true. Is this answering the question?

Yes, thanks! I guess glyphsLib doesn’t need to validate.

Lib/glyphsLib/builder/hints.py
```diff
@@ -49,8 +50,14 @@ def to_glyphs_hints(self, ufo_glyph, layer):
         for attr in ["horizontal", "options", "stem", "type"]:
             setattr(hi, attr, hint[attr])
         for attr in ["origin", "other1", "other2", "place", "scale", "target"]:
-            # FIXME: (jany) what about target = up/down?
             if attr in hint:
-                value = Point(*hint[attr])
+                # handle target = ['u', 'p'] or ['d', 'o', 'w', 'n']
```
anthrotype, line 54: @moyogo why do we need to `handle target = ['u', 'p'] or ['d', 'o', 'w', 'n']` ? Is that for backward compat with how these attribute was written out to UFO before this PR? If so, maybe add a comment with a link to the issue

anthrotype (commented):
What's the status of this?

BTW, I think the flake8 errors should pass after rebasing or merging master onto this.

anthrotype (approved):

Issue #603, opened 2020-03-18
Escaped characters in features names are incorrectly unescaped


khaledhosny:
Related to #602, the font attached there contains Arabic name strings in the feature names block. AFDKO’s makeotf requires non-ASCII characters to be escaped, but glyphsLib will unescape `\062d\0631\0648\0641` into `2d314841`, breaking the string completely.

jenskutilek:
Here's some code to add AFDKO replacement sequences for characters outside ASCII:

https://gist.github.com/jenskutilek/c9adaf1630665c43a7b5e5cc3c7f8b1e

But when you specify the name entry in Glyphs as Unicode (platform 3, encoding 1), I think it's not actually required to escape any characters except " and \ (which this code doesn't handle, and which have to be escaped always in AFDKO strings).

anthrotype:
I think Khaled fixed the lack of UTF-8 support in latest version of makeotf so escape sequences should no longer be required https://github.com/adobe-type-tools/afdko/pull/1133

khaledhosny:
I don’t know if Glyphs uses the new AFDKO yet, so if one is building with Glyphs escaping might still be required.

khaledhosny:
> I don’t know if Glyphs uses the new AFDKO yet, so if one is building with Glyphs escaping might still be required.

The issue about `.glyphs` files with escaped name strings, and I checked latest Glyph release and it still chokes on unescaped ASCII strings, so escaping is still needed if one will be building inside Glyphs.

jenskutilek:
In any case, with #615 escaped characters in name strings are handled correctly as far as I can see.

khaledhosny:
> In any case, with #615 escaped characters in name strings are handled correctly as far as I can see.

It does not fix the issue, unfortunately. If you try with [Qahiri.zip](https://github.com/googlefonts/glyphsLib/files/4346480/Qahiri.zip) and convert it to UFO (with `glyphs2ufo`), the generated feature code for `ss01` and `ss02` looks like: 

```fea
feature ss01 {
# notes:
# featureNames {
#    name "Dot-less Letter Forms";
#    name 3 1 0x0C01 "2d314841 4547454429 (284427 46422737)"; # حروف مهملة (بلا نقاط)
# };
featureNames {
   name "Dot-less Letter Forms";
   name 3 1 0x0C01 "2d314841 4547454429 (284427 46422737)"; # حروف مهملة (بلا نقاط)
};
[...]

PR #615, opened 2020-06-18
Export special feature note to featureNames

### LINKED ISSUE #602: Feature names are not handled
Glyphs seems to (ab)use feature notes for setting features names (see _Names for Stylistic Sets_ section in https://glyphsapp.com/tutorials/stylistic-sets).

Currently glyphsLib will export any thing in the feature notes as a comment.

Test font: [Qahiri.zip](https://github.com/googlefonts/glyphsLib/files/4346480/Qahiri.zip) (`ss01` and `ss02` features).

jenskutilek:
The special feature note from Glyphs

```
Name: Single-storey a
```

will be exported to

```afdko
featureNames {
  name "Single-storey a";
};
```

Fixes #602

### DIFF
--- a/Lib/glyphsLib/builder/features.py
+++ b/Lib/glyphsLib/builder/features.py
@@ -98,8 +98,8 @@ def _to_ufo_features(
             lines.append("# notes:")
             lines.extend("# " + line for line in feature.notes.splitlines())
             feature_name = re.search("^Name: (.+)", feature.notes, flags=re.MULTILINE)
-            if feature_name and len(feature_name.groups()) == 1:
-                name = feature_name.groups()[0]
+            if feature_name:
+                name = feature_name.groups()[-1]
                 # Replace special chars backslash and doublequote for AFDKO syntax
                 name = name.replace("\\", r"\005c")
                 name = name.replace('"', r"\0022")
--- a/Lib/glyphsLib/builder/features.py
+++ b/Lib/glyphsLib/builder/features.py
@@ -667,6 +667,21 @@ class FeatureFileProcessor:
             #    the block is only made of comments
         else:
             feature.code = self._rstrip_newlines(self.doc.text(contents))
+        if feature.automatic:
+            # See if there is a feature name in the code that should be
+            # written to the notes
+            feature_name = re.search(
+                r"(featureNames\s*{\s*name \"(.+)\";\s*};)",
+                feature.code,
+                flags=re.MULTILINE + re.DOTALL,
+            )
+            if feature_name:
+                statement, name = feature_name.groups()
+                if not re.search("^Name: (.+)", feature.notes):
+                    # Don't add if the name is already there
+                    feature.notes = f"Name: {name}\n" + feature.notes
+                # Remove the name statement from the feature code
+                feature.code = feature.code.replace(f"{statement}\n", "")
         self._font.features.append(feature)
         return True
 
--- a/Lib/glyphsLib/builder/features.py
+++ b/Lib/glyphsLib/builder/features.py
@@ -94,28 +94,25 @@ def _to_ufo_features(
     for feature in font.features:
         code = expander.expand(feature.code)
         lines = ["feature %s {" % feature.name]
-        if feature.notes:
-            feature_name = re.search(
-                "(featureNames {.+};)", feature.notes, flags=re.DOTALL
-            )
+        notes = feature.notes
+        if notes:
+            feature_name = re.search("(featureNames {.+};)", notes, flags=re.DOTALL)
             if feature_name:
                 name = feature_name.groups()[-1]
                 lines.append("# notes:")
-                lines.extend("# " + line for line in feature.notes.splitlines())
+                lines.extend("# " + line for line in notes.splitlines())
                 lines.extend(name.splitlines())
             else:
-                feature_name = re.search(r"^(Name: (.+))", feature.notes)
+                feature_name = re.search(r"^(Name: (.+))", notes)
                 if feature_name:
                     line, name = feature_name.groups()
                     # Remove the name from the note
-                    feature.notes = feature.notes.replace(f"{line}", "")
+                    notes = notes.replace(f"{line}", "")
                     # Add notes only if they still contain data
-                    if feature.notes.strip():
+                    if notes.strip():
                         lines.append("# notes:")
                         lines.extend(
-                            "# " + line
-                            for line in feature.notes.splitlines()
-                            if line.strip()
+                            "# " + line for line in notes.splitlines() if line.strip()
                         )
                     # Replace special chars backslash and doublequote for AFDKO syntax
                     name = name.replace("\\", r"\005c")
@@ -123,7 +120,7 @@ def _to_ufo_features(
                     lines.extend(["featureNames {", f'  name "{name}";', "};"])
                 else:
                     lines.append("# notes:")
-                    lines.extend("# " + line for line in feature.notes.splitlines())
+                    lines.extend("# " + line for line in notes.splitlines())
         if feature.automatic:
             lines.append("# automatic")
         if feature.disabled:
--- a/Lib/glyphsLib/builder/features.py
+++ b/Lib/glyphsLib/builder/features.py
@@ -680,21 +680,27 @@ class FeatureFileProcessor:
             #    the block is only made of comments
         else:
             feature.code = self._rstrip_newlines(self.doc.text(contents))
-        if feature.automatic:
-            # See if there is a feature name in the code that should be
-            # written to the notes
-            feature_name = re.search(
-                r"(featureNames\s*{\s*name \"(.+)\";\s*};)",
+        # See if there is a feature name in the code that should be
+        # written to the notes
+        feature_name = re.search(
+            r"(featureNames\s*{\s*name \"(.+)\";\s*};)",
+            feature.code,
+            flags=re.MULTILINE | re.DOTALL,
+        )
+        if feature_name:
+            statement, name = feature_name.groups()
+
+            # Don't add if the name is already there
+            if not re.search("^Name: (.+)", feature.notes, flags=re.DOTALL):
+                feature.notes = "\n".join([f"Name: {name}", feature.notes.lstrip()])
+
+            # Remove the name statement from the feature code
+            feature.code = re.sub(
+                r"featureNames\s*{\s*name \".+\";\s*};\n*",
+                "",
                 feature.code,
-                flags=re.MULTILINE + re.DOTALL,
+                flags=re.MULTILINE | re.DOTALL,
             )
-            if feature_name:
-                statement, name = feature_name.groups()
-                if not re.search("^Name: (.+)", feature.notes):
-                    # Don't add if the name is already there
-                    feature.notes = f"Name: {name}\n" + feature.notes
-                # Remove the name statement from the feature code
-                feature.code = feature.code.replace(f"{statement}\n", "")
         self._font.features.append(feature)
         return True
 
--- a/Lib/glyphsLib/builder/features.py
+++ b/Lib/glyphsLib/builder/features.py
@@ -97,6 +97,11 @@ def _to_ufo_features(
         if feature.notes:
             lines.append("# notes:")
             lines.extend("# " + line for line in feature.notes.splitlines())
+            feature_name = re.search("^Name: (.+)", feature.notes, flags=re.MULTILINE)
+            if feature_name and len(feature_name.groups()) == 1:
+                lines.extend(
+                    ["featureNames {", '  name "%s";' % feature_name.groups()[0], "};"]
+                )
         if feature.automatic:
             lines.append("# automatic")
         if feature.disabled:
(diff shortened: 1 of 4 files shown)


Lib/glyphsLib/builder/features.py
```diff
@@ -89,6 +89,11 @@ def _to_ufo_features(font, ufo=None, generate_GDEF=False, skip_export_glyphs=Non
         if feature.notes:
             lines.append("# notes:")
             lines.extend("# " + line for line in feature.notes.splitlines())
+            feature_name = re.search("^Name: (.+)", feature.notes, flags=re.MULTILINE)
+            if feature_name and len(feature_name.groups()) == 1:
```
anthrotype, line 93: it's unlikely that you get more than one match, but in that case I'd simply take the last one

Lib/glyphsLib/builder/features.py
```diff
@@ -89,6 +89,13 @@ def _to_ufo_features(font, ufo=None, generate_GDEF=False, skip_export_glyphs=Non
...
+                name = feature_name.groups()[0]
+                # Replace special chars backslash and doublequote for AFDKO syntax
+                name = name.replace("\x5c", r"\005c")
+                name = name.replace('"', r"\0022")
+                lines.extend(["featureNames {", '  name "%s";' % name, "};"])
```
anthrotype, line 98:
```suggestion
                lines.extend(["featureNames {", f'  name "{name}";', "};"])
```

Lib/glyphsLib/builder/features.py
jenskutilek: Thanks! Changed in a5fb108.

Lib/glyphsLib/builder/features.py
```diff
@@ -89,6 +89,13 @@ def _to_ufo_features(font, ufo=None, generate_GDEF=False, skip_export_glyphs=Non
         if feature.notes:
             lines.append("# notes:")
             lines.extend("# " + line for line in feature.notes.splitlines())
+            feature_name = re.search("^Name: (.+)", feature.notes, flags=re.MULTILINE)
+            if feature_name and feature_name.groups():
```
anthrotype, line 93: if the returned match object is not None, I think it must contain some non empty groups so the second test `feature_name.groups()` should be redundant

Lib/glyphsLib/builder/features.py
```diff
@@ -89,6 +89,13 @@ def _to_ufo_features(font, ufo=None, generate_GDEF=False, skip_export_glyphs=Non
         if feature.notes:
             lines.append("# notes:")
             lines.extend("# " + line for line in feature.notes.splitlines())
+            feature_name = re.search("^Name: (.+)", feature.notes, flags=re.MULTILINE)
+            if feature_name:
+                name = feature_name.groups()[-1]
+                # Replace special chars backslash and doublequote for AFDKO syntax
+                name = name.replace("\x5c", r"\005c")
```
anthrotype, line 96: I think this would be more readable, I don't immediately remember that backslash codepoint is U+005C
```suggestion
                name = name.replace("\\", r"\005c")
```

anthrotype:
the flake8 check failures appear to be unrelated to this PR
https://travis-ci.org/github/googlefonts/glyphsLib/jobs/699627388

@madig any idea where these are coming from?

Lib/glyphsLib/builder/features.py
jenskutilek: Sure!

jenskutilek:
> the flake8 check failures appear to be unrelated to this PR
> https://travis-ci.org/github/googlefonts/glyphsLib/jobs/699627388

Yes, they seem to appear on all recent pull requests ...

madig:
Uh. No idea why these were missed before, except if flake8 actually never ran on those files. Uhm. Feel free to disable both errors in the config for this PR.

madig:
Just updated master directly, you can rebase.

jenskutilek:
Looking at the test font @khaledhosny posted on #602, this PR is not complete yet. It seems the long AFDKO feature name syntax can also be in the feature note. I thought it should go in the regular feature text.

anthrotype:
the CLA bot is complaining with the cherry-pick of Nikolaus commit. @jenskutilek do you mind undoing that and instead do a regular merge (`git fetch origin && git merge origin/master`, provided that `origin` is the upstream repo).

setup.cfg
```diff
@@ -61,7 +61,7 @@ formats = zip
 
 [flake8]
 select = B, C, E, F, W, T4, B9
-ignore = E203, E266, E501, W503
+ignore = E203, E266, E501, W503, E741
```
anthrotype, line 64: hm, I would have rather fixed the offending code, rather than mute the flake8 warning globally

madig: I've seen that warning every time you use a single-letter variable like in a list comprehension, so I decided to mute it for now.

tests/builder/builder_test.py
```diff
@@ -1009,6 +1010,53 @@ def test_glyph_lib_Export_GDEF(self):
...
+            self.assertIn(r'name "Single\005cstorey \0022ä\0022";', ufo.features.text)
+
+    def test_glyph_lib_Export_feature_names_long_from_notes(self):
+        font = generate_minimal_font()
+        add_glyph(font, "a")
+        add_glyph(font, "a.ss01")
+        ss01 = GSFeature(name="ss01", code="sub a by a.ss01;")
+        font.features.append(ss01)
+        for note in (
+            'featureNames {\n  name 3 1 0x401 "Alternate {ä};";\n  name 3 1 0x409 "Alternate {};";\n};\n',
```
anthrotype, line 1037: it would be good if you could wrap these very long strings here and further below into multiple lines

anthrotype:
@madig ok to merge? Can you then tag a release with this after the merge? Thank you

madig:
Idle thought: is it necessary to do anything when going UFO -> Glyphs?

schriftgestalt:
If the "automatic" flag is round tripped, then it would be good to get the feature names out of the feature code and into the notes.

justvanrossum:
Ping. Just running into this. Would be nice to get it merged.

anthrotype:
I think the reason we haven't merged this yet is that it doesn't seem to fix https://github.com/googlefonts/glyphsLib/issues/603#issuecomment-646732838

I suppose we can merge this first and add a TODO for later..

Lib/glyphsLib/builder/features.py
```diff
@@ -89,6 +89,22 @@ def _to_ufo_features(font, ufo=None, generate_GDEF=False, skip_export_glyphs=Non
...
             lines.extend("# " + line for line in feature.notes.splitlines())
+            feature_name = re.search(
+                "(featureNames {.+};)", feature.notes, flags=re.DOTALL
+            )
+            if feature_name:
+                name = feature_name.groups()[-1]
+                lines.extend(name.splitlines())
+            else:
+                feature_name = re.search(
+                    "^Name: (.+)", feature.notes, flags=re.MULTILINE
```
anthrotype, line 100: according to a comment from @schriftgestalt:

> only if the first line starts with "Name:", the rest of the line will be used

here you're matching any line in features.notes that starts with `Name:`

jenskutilek: Yep, that was before Georg clarified the behaviour :)

jenskutilek:
> I think the reason we haven't merged this yet is that it doesn't seem to fix [#603 (comment)](https://github.com/googlefonts/glyphsLib/issues/603#issuecomment-646732838)
> 
> I suppose we can merge this first and add a TODO for later..

The roundtripping behaviour https://github.com/googlefonts/glyphsLib/pull/615#issuecomment-651056231 is not implemented yet. I'll see if I can do it today.

Lib/glyphsLib/builder/features.py
```diff
@@ -630,6 +630,21 @@ def _process_feature_block(self):
...
+            feature_name = re.search(
+                r"(featureNames\s*{\s*name \"(.+)\";\s*};)",
+                feature.code,
+                flags=re.MULTILINE + re.DOTALL,
+            )
+            if feature_name:
+                statement, name = feature_name.groups()
+                if not re.search("^Name: (.+)", feature.notes):
+                    # Don't add if the name is already there
+                    feature.notes = f"Name: {name}\n" + feature.notes
```
jenskutilek, line 645: The feature names may go out of sync here, because we don't check if what's already in the feature notes contains the same name as the featureNames block. Do you think that's a problem?

Lib/glyphsLib/builder/features.py
```diff
@@ -96,9 +96,7 @@ def _to_ufo_features(font, ufo=None, generate_GDEF=False, skip_export_glyphs=Non
                 name = feature_name.groups()[-1]
                 lines.extend(name.splitlines())
             else:
-                feature_name = re.search(
-                    "^Name: (.+)", feature.notes, flags=re.MULTILINE
-                )
+                feature_name = re.search("^Name: (.+)", feature.notes)
```
jenskutilek, line 99: Do you think the name should be removed from the notes comment after it has been written to the UFO feature code?

If it is left in the notes comment, we get code like this in the UFO:

```afdko
feature ss01 {
# notes:
# Name: Alternate g
featureNames {
  name "Alternate g";
};
# automatic
sub g by g.ss01;
} ss01;
```

anthrotype: Yes I think we should avoid the duplication

jenskutilek:
> Why only for the automatic features?

Not sure ;) I changed it to process all features.

tests/builder/features_test.py
```diff
@@ -147,6 +147,46 @@ def test_feature_name_automatic(tmpdir, ufo_module):
...
+    """
+    )
+
+    font, rtufo = roundtrip(ufo, tmpdir, ufo_module)
+
+    # Check code in Glyphs font
+    gs_feature = font.features[0]
+    assert gs_feature.automatic
+    assert gs_feature.code.strip() == "sub g by g.ss01;"
+    assert gs_feature.notes.strip() == "Name: Alternate g"
```
jenskutilek, line 161: This assertion fails:

```bash
        assert gs_feature.code.strip() == "sub g by g.ss01;"
>       assert gs_feature.notes.strip() == "Name: Alternate g"
E       AssertionError: assert '' == 'Name: Alternate g'
E         - Name: Alternate g

tests/builder/features_test.py:161: AssertionError
```

I have no idea why it doesn't work. When I convert a similar file with `ufo2glyphs`, everything looks fine. If you see some stupid error that I made, let me know ;)

Lib/glyphsLib/builder/features.py
```diff
@@ -650,22 +650,22 @@ def _process_feature_block(self):
...
         )
         if feature_name:
             statement, name = feature_name.groups()
 
             # Don't add if the name is already there
-            if not re.search("^Name: (.+)", feature.notes, flags=re.DOTALL):
+            if not re.search(r"^Name: (.+)", feature.notes, flags=re.DOTALL):
                 feature.notes = "\n".join(
-                    [f"Name: {name}", feature.notes.lstrip()]
+                    [f"Name: {name}", notes_text]
```
jenskutilek, line 663: This is the code that sets the feature note, but sometimes doesn't ...?

Lib/glyphsLib/builder/features.py
jenskutilek: Ok

anthrotype:
I'm now leaning towards _not_ supporting the roundtripping of these special comments when going back from UFO to Glyphs.
If they are there in .glyphs source, we cut and paste them to UFO's features.fea code; when going back to .glyphs, we keep as is and don't do anything clever with the feature notes. I'm just trying to reduce the complexity here.

anthrotype:
.. or maybe we only try to do that for "automatic" features as you did originally, to avoid surprises? sorry for the back and forth. I know how frustrating it can be when attempting to make _any_ changes in glyphsLib.

schriftgestalt:
For automatic features, it would be good to properly round trip. 

And a small heads up how this is stored in Glyphs 3 files (indentation added for clarity):
```
{
	code = "sub a by b;";
	labels = (
		{
			language = ENG;
			value = "English Name";
		},
		{
			language = FOS;
			value = "Faroese Name";
		}
	);
	tag = ss01;
}
```

jenskutilek:
I'm kind of lost here. I haven't looked at this in a while, but as I remember I can't get the roundtripping test to work because apparently at some point original data from the UFO lib is restored, overwriting any changes made at the UFO stage. It's hard to follow what exactly happens.

khaledhosny:
With #603 fixed, can we split the round-tripping part into its own PR and keep this for the Glyphs -> UFO conversion and merge it?

khaledhosny:
I have this branch rebased locally, should I force push here or open a new PR?

anthrotype:
You can force push thanks

khaledhosny:
> And a small heads up how this is stored in Glyphs 3 files (indentation added for clarity):

@schriftgestalt I’m trying to support the Glyphs 3 format, but I have a couple of questions:
1. How to map Glyphs’s label `language` to name table `langID`?
2. What to set `platformID` and `platEncID` to? (3, 1)?

khaledhosny:
The regression tests failure is because feature names are now properly handled.

khaledhosny:
This is ready, except for the mapping between glyphs language codes and name table language IDs (needed only for Glyphs 3 files).

jenskutilek:
Thank you @khaledhosny for picking this up and finishing it!

anthrotype:
Thanks Khaled!

PR #616, opened 2020-06-26
Improve brace layer name diagnostics

chrissimpkins:
I've been tinkering with brace layers while we transition an optical size axis range and came across an exception when I attempted to define a brace layer with a float rather than an int value.  The error message was:

> ValueError: invalid literal for int() with base 10: '17.01'

which is a bit cryptic. 

This PR adds a try/except block around the int cast attempt and includes information about the requirement for int values in brace layers as part of the diagnostics.  It would be ideal to point to the affected glyph and brace layer name, but a unique glyph name does not appear to be accessible in this area of the code? 

Also adds a couple of unit tests for brace layer name validity testing (against an int definition and a float definition).


anthrotype:
Why don't we simply add support for floats?

schriftgestalt:
Glyphs does support floats in brace layers.

PR #617, opened 2020-07-16
Improve anchor propagation

yanone:
This PR is addressing two separate issues of anchor propagation:

A) marks that are in turn composed of other marks (e.g. `shadda_fatha-ar`) didn’t receive propagated anchors at all
B) ligature anchors (`top_1`, `top_2` etc) weren’t correctly propagated when composed of base glyphs and a mark (e.g. `lam_alefHamzaabove-ar`).

The solution to A) was already developed in `ufo2ft`’s `PropagateAnchorsFilter`, so I simply copied the code over.

The solution to B) is to check for marks attached to named anchors (stored in `Glyph.lib['com.schriftgestaltung.Glyphs.ComponentInfo']`) and override the named anchor’s position with the transformation result.


Lib/glyphsLib/builder/anchors.py
```diff
@@ -125,3 +160,37 @@ def to_glyphs_glyph_anchors(self, ufo_glyph, layer):
...
+def _is_ligature_mark(glyph):
+    return not glyph.name.startswith("_") and "_" in glyph.name
+
+
+def _bounds(component, glyph_set):
+    """Return the (xmin, ymin) of the bounds of `component`."""
+    if hasattr(component, "bounds"):  # e.g. defcon
+        return component.bounds[:2]
+    elif hasattr(component, "draw"):  # e.g. ufoLib2
+        pen = fontTools.pens.boundsPen.BoundsPen(glyphSet=glyph_set)
```
anthrotype, line 190: ufoLib2 has Component.getBounds(layer) method, since v0.6.1:
https://github.com/fonttools/ufoLib2/blob/e9a5c5d2d1a0ddd66e6e8f51e7c58f2e494f164b/src/ufoLib2/objects/component.py#L42-L49

Lib/glyphsLib/builder/anchors.py
```diff
@@ -95,18 +113,35 @@ def _get_anchor_data(anchor_data, ufo, components, anchor_name):
...
 
-def _adjust_anchors(anchor_data, ufo, component):
-    """Adjust anchors to which a mark component may have been attached."""
+def _anchor(_glyph, _component):
+    """Pull component’s named anchor from Glyph.lib"""
+    if COMPONENT_INFO_KEY in _glyph.lib:
+        for _anchorLib in _glyph.lib[COMPONENT_INFO_KEY]:
+            if (
+                "anchor" in _anchorLib
+                and "name" in _anchorLib
```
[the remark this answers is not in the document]
yanone: More than one component with same name is possible, but in our case of a ligature glyph we want the component that’s attached to a certain named anchor (`top_2`), so I'm checking for that and believe that the component index is irrelevant here. I have nothing to compare that index against.

Lib/glyphsLib/builder/anchors.py
```diff
@@ -95,18 +113,35 @@ def _get_anchor_data(anchor_data, ufo, components, anchor_name):
...
+            ):
+                return _anchorLib["anchor"]
 
+
+def _adjust_anchors(anchor_data, ufo, parent, component):
+    """Adjust anchors to which a mark component may have been attached."""
     glyph = ufo[component.baseGlyph]
     t = Transform(*component.transformation)
     for anchor in glyph.anchors:
+        _namedAnchor = _anchor(parent, component)
```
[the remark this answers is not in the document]
yanone: Done

Lib/glyphsLib/builder/anchors.py
```diff
@@ -95,18 +113,35 @@ def _get_anchor_data(anchor_data, ufo, components, anchor_name):
...
         # only adjust if this anchor has data and the component also contains
```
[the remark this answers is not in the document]
yanone: Done

Lib/glyphsLib/builder/anchors.py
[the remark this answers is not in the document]
anthrotype: The "name" in `_anchorLib` is the name of the component not the name of the anchor (that's the `_anchorLib["anchor"]`); there can be more than one components with the same name. The component index should be enumerated when we do `for component in parent.components` and passed down to the `_anchor` function.
The COMPONENT_INFO_KEY is the additional Glyphs-specific metadata that is associated with a particular component, identified by its index, which is unique, not the name. So you should match them by index. @madig am I correct?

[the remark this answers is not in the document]
anthrotype: I suggest you add a test with a composite glyph with more than one component with the same name, where each has different named anchors; your current code will only match and use the component info of the first component with a given name.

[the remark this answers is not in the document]
yanone: Okay, I got what you mean now. I'm on it.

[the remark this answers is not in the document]
yanone: Actually this messes up even in Glyphs using `GSLayer.anchorsTraversingComponents()`. Instead of `top_1` and `top_2` both moving upwards, both stay where they were, and a previously non-existent `top` anchor appears where I though `top_1` would end up. Nothing such happens on the `top_2` side, it just stays in its old place :/

I can still implement this if you like, but I must add that this will be ahead of Glyphs and therefore inconsistent ;)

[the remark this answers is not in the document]
anthrotype: that looks like a bug which Georg can easily fix, I prefer that we do the right thing.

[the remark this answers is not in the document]
yanone: Okay, I'll let him know. 
But I still don't get your idea with the index, sorry.
This is the lib regarding the anchors of the glyph with the double components: `[{'name': 'hamzaabove-ar', 'index': 1, 'anchor': 'top_2'}, {'name': 'hamzaabove-ar', 'index': 2, 'anchor': 'top_1'}]`
What do you want me to compare the index against? I have only the anchor names in `anchor_data` and compare them against what is the result of `_anchor()` to identify when to apply transformations.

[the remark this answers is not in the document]
anthrotype: against the index of the component in the list of components in the parent composite glyph. We need to record that alongside the component name to fully indentify the specific component info metadata (where the anchor name is) that is stored in the glyph.lib (a component doesn't have a lib of its own).
The index of the component is its position in the list of glyph.components. When you iterate over the components, you use `for i, component in enumerate(glyph.components):` and store that index and pass it down to where you need it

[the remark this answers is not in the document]
yanone: The test is a bit more tricky. The test glyph I created was fictional. A real-world `lamAlef-ar` with two components however propagates the anchors correctly in Glyphs, such as `lamVabove_alefWasla-ar`. So I need to find a test glyphs that's defined in the `GlyphData.xml` that has two identical components.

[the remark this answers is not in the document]
schriftgestalt: I just tried and the "top_2" vs "top" was correct when running `GSLayer.anchorsTraversingComponents()`. Can you point me to a sample file where it doesn't work?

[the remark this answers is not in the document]
yanone: I did run into the issue you pointed out with my fictional test glyph with two duplicate marks, and solved it.

Lib/glyphsLib/builder/anchors.py
yanone: I can change that if you like, but it means introducing `ufoLib2` as a dependency. `fontTools` is already a dependency today, so that works. I don't know which is better here. Let me know.

anthrotype: ufoLib2 is already a dependency of glyphsLib, see https://github.com/googlefonts/glyphsLib/blob/1ad88c1dedb3072f1d9e62dc58b268496a1c6bc3/setup.cfg#L37

yanone: Okay. I overlooked that, sorry

yanone: I don't understand how `getBounds()` works, especially that thing with the Drawable.
However, the method of bounds calculation isn't related to my issue here with the anchors, so I will skip this one and leave it to you. It works without changes.
I’m going into a summer break in two days and need to finish this up. Sorry for that.

anthrotype: no problem!

basically you would just do `return return component.getBounds(glyph_set)`

yanone:
@anthrotype 
With the exception of the bounds calculation I've addressed all your remarks. Please have another look.

anthrotype:
i'll take a look soon, thanks

Lib/glyphsLib/builder/anchors.py
```diff
@@ -95,18 +113,36 @@ def _get_anchor_data(anchor_data, ufo, components, anchor_name):
...
+        # adjust either if this anchor has data and the component also contains
+        # the associated mark anchor (e.g. "_top" for "top") ...
         if anchor.name in anchor_data and any(
             a.name == "_" + anchor.name for a in glyph.anchors
         ):
             anchor_data[anchor.name] = t.transformPoint((anchor.x, anchor.y))
+        # ... or component is attached to a specific named anchor
+        # (e.g. top_2 for a ligature glyph)
+        # rather than to the standard anchors (top/bottom)
+        elif _componentAnchor in anchor_data:
```
anthrotype, line 144: shouldn't the second case (ie. explicit named component.anchor whatever that is) take precedence over the first one?

yanone: You are right. Earlier I had it flipped already but something didn't work, so I had left it untouched.
I just changed it around.

Lib/glyphsLib/builder/anchors.py
```diff
@@ -132,17 +132,17 @@ def _adjust_anchors(anchor_data, ufo, parent, component):
     t = Transform(*component.transformation)
     _componentAnchor = _componentAnchorFromLib(parent, component)
     for anchor in glyph.anchors:
-        # adjust either if this anchor has data and the component also contains
+        # adjust either if component is attached to a specific named anchor
+        # (e.g. top_2 for a ligature glyph)
+        # rather than to the standard anchors (top/bottom)
+        if _componentAnchor in anchor_data:
```
anthrotype, line 138: I know it's unlikely that `anchor_data` dict contains an item with key `None` or `""`, but to avoid surprises it'd be better to check `if _componentAnchor and _componentAnchor in anchor_data`

yanone: Okay, done

madig:
> Promote the mark that is positioned closest to the origin to a base.

Random thought: I have written at least two different algorithms for determining the base glyph. It's almost as if there should be some form of semantics for this, somewhere.

Lib/glyphsLib/builder/anchors.py
```diff
@@ -60,14 +62,30 @@ def _propagate_glyph_anchors(self, ufo, parent, processed):
...
+            component = _component_closest_to_origin(mark_components, ufo)
+        except Exception as e:
+            raise Exception(
+                "Error while determining which component of composite "
+                "'{}' is the lowest: {}".format(parent.name, str(e))
+            )
+        mark_components.remove(component)
+        base_components.append(component)
+        glyph = ufo[component.baseGlyph]
+        anchor_names |= {a.name for a in glyph.anchors}
```
madig, line 79: Do we need to go recursive here?

anthrotype: the recursive-ness is in the self-call to `_propagate_glyph_anchors` above

Lib/glyphsLib/builder/anchors.py
```diff
@@ -60,14 +62,30 @@ def _propagate_glyph_anchors(self, ufo, parent, processed):
...
+    if mark_components and not base_components and _is_ligature_mark(parent):
+        # The composite is a mark that is composed of other marks (E.g.
+        # "circumflexcomb_tildecomb"). Promote the mark that is positioned closest
+        # to the origin to a base.
+        try:
+            component = _component_closest_to_origin(mark_components, ufo)
+        except Exception as e:
+            raise Exception(
+                "Error while determining which component of composite "
+                "'{}' is the lowest: {}".format(parent.name, str(e))
```
madig, line 74: Maybe add a `from e` here to chain the exception.

Lib/glyphsLib/builder/anchors.py
```diff
@@ -60,14 +62,30 @@ def _propagate_glyph_anchors(self, ufo, parent, processed):
...
+            )
```
anthrotype, line 75: yes, good idea

```suggestion
            ) from e
```

PR #618, opened 2020-08-03
Support 'Axis Mappings' customParameter

m4rc1e:
Support the 'Axis Mappings' customParameter.

If this customParameter has been defined, it takes precedence over all our other methods to assemble the designspace axes.

Todo:
- Possibly add a more informative comment
- regularDesignLoc and regularUserLoc are copy pasta. We may want to refactor this.

### DIFF
--- a/Lib/glyphsLib/builder/axes.py
+++ b/Lib/glyphsLib/builder/axes.py
@@ -296,7 +296,7 @@ def to_glyphs_axes(self):
     if axes_parameter and not _is_subset_of_default_axes(axes_parameter):
         self.font.customParameters["Axes"] = axes_parameter
 
-    if self.minimize_ufo_diffs:
+    if any(a.map for a in self.designspace.axes):
         mapping = {
             axis.tag: {str(k): v for k, v in axis.map} for axis in self.designspace.axes
         }
--- a/Lib/glyphsLib/builder/axes.py
+++ b/Lib/glyphsLib/builder/axes.py
@@ -141,7 +141,13 @@ def to_designspace_axes(self):
     assert isinstance(regular_master, classes.GSFontMaster)
 
     custom_mapping = self.font.customParameters["Axis Mappings"]
-
+    if not custom_mapping:
+        logger.warning(
+            "Recommended 'Axis Mappings' customParameter not set. This parameter "
+            "allows developers define an explicit mapping, otherwise the mapping "
+            "is calculated by using the 'Axis Locations' customParameters or by "
+            "using values from the instances and masters."
+        )
     for axis_def in get_axis_definitions(self.font):
         axis = self.designspace.newAxisDescriptor()
         axis.tag = axis_def.tag
@@ -171,7 +177,7 @@ def to_designspace_axes(self):
             else:
                 logger.debug(
                     f"Skipping {axis.tag} since it hasn't been defined "
-                    "in the Axis Mapping"
+                    "in the Axis Mapping."
                 )
                 continue
         # See https://github.com/googlefonts/glyphsLib/issues/280
@@ -191,7 +197,7 @@ def to_designspace_axes(self):
             regularDesignLoc = axis_def.get_design_loc(regular_master)
             regularUserLoc = axis_def.get_user_loc(regular_master)
         else:
-            # Build the mapping from the isntances because they have both
+            # Build the mapping from the instances because they have both
             # a user location and a design location.
             instance_mapping = {}
             for instance in self.font.instances:
--- a/Lib/glyphsLib/builder/axes.py
+++ b/Lib/glyphsLib/builder/axes.py
@@ -175,7 +175,7 @@ def to_designspace_axes(self):
                 )
                 continue
         # See https://github.com/googlefonts/glyphsLib/issues/280
-        elif font_uses_new_axes(self.font):
+        elif font_uses_axis_locations(self.font):
             # Build the mapping from the "Axis Location" of the masters
             # TODO: (jany) use Virtual Masters as well?
             mapping = {}
@@ -259,7 +259,7 @@ def to_designspace_axes(self):
         )
 
 
-def font_uses_new_axes(font):
+def font_uses_axis_locations(font):
     # It's possible for fonts to have the 'Axes' parameter but to NOT specify
     # the master locations using 'Axis Location', in which case we have to
     # resort to using instances or other old tricks to get the mapping.
--- a/Lib/glyphsLib/builder/axes.py
+++ b/Lib/glyphsLib/builder/axes.py
@@ -170,7 +170,7 @@ def to_designspace_axes(self):
                 else:
                     regularUserLoc = min(mapping)
             else:
-                logger.warning(
+                logger.debug(
                     f"Skipping {axis.tag} since it hasn't been defined "
                     "in the Axis Mapping"
                 )
--- a/Lib/glyphsLib/builder/axes.py
+++ b/Lib/glyphsLib/builder/axes.py
@@ -160,13 +160,21 @@ def to_designspace_axes(self):
             axis_wanted = True
 
         # See https://github.com/googlefonts/glyphsLib/issues/568
-        if custom_mapping and axis.tag in custom_mapping:
-            mapping = {int(k): v for k, v in custom_mapping[axis.tag].items()}
-            regularDesignLoc = axis_def.get_design_loc(regular_master)
-            # Glyphs masters don't have a user location, so we compute it by
-            # looking at the axis mapping in reverse.
-            reverse_mapping = [(dl, ul) for ul, dl in sorted(mapping.items())]
-            regularUserLoc = interp(reverse_mapping, regularDesignLoc)
+        if custom_mapping:
+            if axis.tag in custom_mapping:
+                mapping = {float(k): v for k, v in custom_mapping[axis.tag].items()}
+                regularDesignLoc = axis_def.get_design_loc(regular_master)
+                reverse_mapping = {v: k for k, v in mapping.items()}
+                if regularDesignLoc in reverse_mapping:
+                    regularUserLoc = reverse_mapping[regularDesignLoc]
+                else:
+                    regularUserLoc = min(mapping)
+            else:
+                logger.warning(
+                    f"Skipping {axis.tag} since it hasn't been defined "
+                    "in the Axis Mapping"
+                )
+                continue
         # See https://github.com/googlefonts/glyphsLib/issues/280
         elif font_uses_new_axes(self.font):
             # Build the mapping from the "Axis Location" of the masters
@@ -292,20 +300,10 @@ def to_glyphs_axes(self):
         self.font.customParameters["Axes"] = axes_parameter
 
     if self.minimize_ufo_diffs:
-        # TODO: (jany) later, when Glyphs can manage general designspace axes
-        # self.font.userData[AXES_KEY] = [
-        #     dict(
-        #         tag=axis.tag,
-        #         name=axis.name,
-        #         minimum=axis.minimum,
-        #         maximum=axis.maximum,
-        #         default=axis.default,
-        #         hidden=axis.hidden,
-        #         labelNames=axis.labelNames,
-        #     )
-        #     for axis in self.designspace.axes
-        # ]
-        pass
+        mapping = {
+            axis.tag: {str(k): v for k, v in axis.map} for axis in self.designspace.axes
+        }
+        self.font.customParameters["Axis Mappings"] = mapping
 
 
 class AxisDefinition:
--- a/Lib/glyphsLib/builder/axes.py
+++ b/Lib/glyphsLib/builder/axes.py
@@ -140,6 +140,8 @@ def to_designspace_axes(self):
     regular_master = get_regular_master(self.font)
     assert isinstance(regular_master, classes.GSFontMaster)
 
+    custom_mapping = self.font.customParameters["Axis Mappings"]
+
     for axis_def in get_axis_definitions(self.font):
         axis = self.designspace.newAxisDescriptor()
         axis.tag = axis_def.tag
@@ -157,8 +159,16 @@ def to_designspace_axes(self):
         ):
             axis_wanted = True
 
+        # See https://github.com/googlefonts/glyphsLib/issues/568
+        if custom_mapping and axis.tag in custom_mapping:
+            mapping = {int(k): v for k, v in custom_mapping[axis.tag].items()}
+            regularDesignLoc = axis_def.get_design_loc(regular_master)
+            # Glyphs masters don't have a user location, so we compute it by
+            # looking at the axis mapping in reverse.
+            reverse_mapping = [(dl, ul) for ul, dl in sorted(mapping.items())]
+            regularUserLoc = interp(reverse_mapping, regularDesignLoc)
         # See https://github.com/googlefonts/glyphsLib/issues/280
-        if font_uses_new_axes(self.font):
+        elif font_uses_new_axes(self.font):
             # Build the mapping from the "Axis Location" of the masters
             # TODO: (jany) use Virtual Masters as well?
             mapping = {}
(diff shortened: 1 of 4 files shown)


m4rc1e:
My implementation is based on the following points I raised in this [doc](https://docs.google.com/document/d/1RwFQteeZzzeUn2RUJBiCp0hsplFgws_K0nrAmOtLQ4w/edit)

> - If a user has included an axis mapping, it should take precedence over every other implementation we have.
> - Axis locations should be ignored if an axis mapping is present. We should raise a warn in fontmake if axis locations are present
> - Axis mappings should be structured as user_coordinate: internal_coordinate. Both Glyphsapp and GlyphsLib should use this structure.

m4rc1e:
I think this is good for a first review.

Lib/glyphsLib/builder/axes.py
```diff
@@ -157,8 +159,16 @@ def to_designspace_axes(self):
         ):
             axis_wanted = True
 
+        # See https://github.com/googlefonts/glyphsLib/issues/568
+        if custom_mapping and axis.tag in custom_mapping:
+            mapping = {int(k): v for k, v in custom_mapping[axis.tag].items()}
+            regularDesignLoc = axis_def.get_design_loc(regular_master)
+            # Glyphs masters don't have a user location, so we compute it by
+            # looking at the axis mapping in reverse.
```
anthrotype, line 167: yes, this comment is definitely copy-pasta. What you're trying to do here is computing the user-space location of the base master, in case the Axis Mappings does not contain an explicit mapping for it.
I am wondering, should we not instead require that all minimum, default and maximum coordinates along each axes that are mapped in the new Axis Mappings be explicitly specified by the user instead of doing this linear interpolation like we do here for the default? The latter interpolation only works if the the mapping contains a min and max, and min != max. Requiring all three to be set (min, default and max, with default possibly == to either min or max) would make it easier to think about it.

Lib/glyphsLib/builder/axes.py
```diff
@@ -157,8 +159,16 @@ def to_designspace_axes(self):
         ):
             axis_wanted = True
 
+        # See https://github.com/googlefonts/glyphsLib/issues/568
+        if custom_mapping and axis.tag in custom_mapping:
+            mapping = {int(k): v for k, v in custom_mapping[axis.tag].items()}
```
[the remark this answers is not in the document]
m4rc1e: The Axis Mappings customParameter stores the user coordinates as strings.

tests/builder/axes_test.py
```diff
@@ -305,3 +305,44 @@ def test_single_master_default_weight_400(ufo_module):
...
+        "wght": {k: v for k, v in wght_mapping},
+        "wdth": {k: v for k, v in wdth_mapping},
+    }
+
+    font.customParameters["Axis Mappings"] = axis_mappings
+    doc = to_designspace(font, ufo_module=ufo_module)
+
+    assert doc.axes[0].name == "Weight"
+    assert doc.axes[0].minimum == 100
+    assert doc.axes[0].maximum == 900
```
anthrotype, line 337: should you not also test the axis.default value?

Lib/glyphsLib/builder/axes.py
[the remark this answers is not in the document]
anthrotype: ah.. then float(k)

tests/builder/axes_test.py
```diff
@@ -305,3 +305,44 @@ def test_single_master_default_weight_400(ufo_module):
...
+
+    assert doc.axes[1].name == "Width"
+    assert doc.axes[1].minimum == 75
+    assert doc.axes[1].maximum == 100
+    # No axis map needed for Width because user and design coords are identical
+    assert doc.axes[1].map != wdth_mapping
+    assert doc.axes[1].map == []
+
+    font = to_glyphs(doc)
+    assert font.customParameters["Axis Mappings"] == axis_mappings
```
anthrotype, line 348: are we round-tripping the "Axis Mappings" custom parameter when we go from designspace to .glyphs? We should, but I don't see where this is happening...

m4rc1e: Yes, I'm round-tripping .

I'm converting the designspace to glyphs and then I'm checking if the Axis Mappings customParameter is equal to the axis_mappings var which is defined on line 327.

anthrotype: yes but where in the lib modules this happens? I mean, the setting of customParameters["Axis Mappings"] to the content of a designspace.axes maps? I don't see that in your non-test code

Lib/glyphsLib/builder/axes.py
```diff
@@ -157,8 +159,16 @@ def to_designspace_axes(self):
         ):
             axis_wanted = True
 
+        # See https://github.com/googlefonts/glyphsLib/issues/568
+        if custom_mapping and axis.tag in custom_mapping:
```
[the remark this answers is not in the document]
m4rc1e: You are correct. If a mapping isn't defined in the Axis Mapping, it will use the previous mechanisms. I agree with your omit approach. I'll change this and remove the copy pasta.

tests/builder/axes_test.py
m4rc1e: We now have three methods to create the mapping. Each method only produces a variable called `mapping`. It doesn't do the assignment. I create this variable in line 164:

`mapping = {float(k): v for k, v in custom_mapping[axis.tag].items()}`

The mapping variable is then converted to a designspace map in line 233:

`axis.map = sorted(mapping.items())`

anthrotype: that's the code that sets the DS axes maps from the Glyphs' custom parameter. I am talking about the reverse code, that sets the Glyphs custom parameter from the DesignSpace mappings, when one coverts a DS to .glyphs file.

m4rc1e: I'm pretty sure `glyphsLib.builder.custom_params` is able to convert customParams to ufo and back again without fault. I could be wrong here but in all my tests, it worked just fine. Just now I made a new glyphs file with an unknown customParam and it has been preserved in a roundtrip.

anthrotype: Ah.. I see what's happening. The designspace file `lib` or the master UFOs' `lib.plist` is storing the custom parameter and that explains why you can roundtrip it back. 

What I was looking for is the piece of code (there isn't one currently) that given a designspace file with some axes map elements (but no Glyphs-specific private key in its libs) would construct an "Axis Mappings" custom parameter out of it when creating the GSFont object out of the DS.
We need to think whether we want to do that or not. I would think this would be desirable. /cc @madig @belluzj

anthrotype: in any case, this is not clear in @m4rc1e's tests. By looking at the tests only one would think that a new Axis Mappings is created from the designspace axes maps only, whereas what's happening is it is simply being restored. Maybe Marc you could add a comment to clarify that.

m4rc1e: Sure thing.

belluzj: Yes I think it's very desirable that the "axis mapping" (general term, not implementation-related) is stored as a <map> element under each <axis> in the Designspace file, and actually only there (not as a lib key) so that the "axis mapping" information on each side (Designspace or Glyphs.app) is stored in that side's native format for that information.

Lib/glyphsLib/builder/axes.py
m4rc1e: > I am wondering, should we not instead require that all minimum, default and maximum coordinates along each axes that are mapped in the new Axis Mappings be explicitly specified by the user instead of doing this linear interpolation like we do here for the default? 

I guess we could do the following:

For the min and max, we can simply use `min(mapping)`, `max(mapping)`.

For the default, we can use the "Variable Font Origin" customParameter. If it isn't present, we could use the min value for the default.

Alternatively, if there is no "Variable Font Origin" customParameter, we skip this new Axis Mappings method completely because we don't have enough information. We should then log a warning informing the user that they need to specify a "Variable font Origin" in order for it to work.

m4rc1e: Using the Variable Font Origin still means we'll need to convert from internal coords to user coords which may be messy.

@anthrotype I agree that it may be best to get the user to define this info themselves. We could change the Axis Mappings data structure to:
```
{
    "wght": {
        "minimum": 100, # this and maximum can be computed if needed
        "maximum": 900, 
        "default": 100,
        "mapping": {100: 100, 400: 300, 900: 900},
    "wdth": {...}
}
```

Obviously we can only do this if Georg agrees.

anthrotype: > For the default, we can use the "Variable Font Origin" 

that logic is already implemented in `get_regular_master` function in this module, which returns the "regular" master (i.e. the base or default or neutral, not "regular" as the Regular style). You search that master's design location in the set of values of the axis mappings dict; if it's present, use the key (user-space location) as is; if missing, raise an error and require the user to add an explicit mapping for the default master.

anthrotype: no, I don't think we need to change the data structure of Axis Mappings custom parameter. We only need to check that there are explicit mappings for the minimum, default and maximum values along each axis. Note that if we don't check and enforce that here, it will be varLib later on that will complain as it attempts to build the avar table.

The requirement that min, default and max be explicitly mapped does not mean there has to be necessarily three key:value pairs; because one of them (the default) may well be == to min or max. So the minimum number for a meaningful axis that does vary is two, with default == min or max. If there's only one mapping then the min == max == default and the variation axis is really just a point and there ought to be only one position for all the involved masters and instances for that axis.

m4rc1e: Thanks for the tips, I now have the following:

```
mapping = {float(k): v for k, v in custom_mapping[axis.tag].items()}
regularDesignLoc = axis_def.get_design_loc(regular_master)

reverse_mapping = {v: k for k,v in mapping.items()}
if regularDesignLoc not in reverse_mapping:
    raise KeyError("Axis Mapping must include a mapping for the default master")
regularUserLoc = reverse_mapping[regularDesignLoc]
```

tests/builder/axes_test.py
m4rc1e: Instead of leaving a comment, I've decided to implement this. I've removed Jany's old commented out code and replaced it with a simple dict comprehension.

Lib/glyphsLib/builder/axes.py
```diff
@@ -157,8 +159,24 @@ def to_designspace_axes(self):
...
+                    regularUserLoc = min(mapping)
+            else:
+                logger.warning(
+                    f"Skipping {axis.tag} since it hasn't been defined "
+                    "in the Axis Mapping"
+                )
+                continue
         # See https://github.com/googlefonts/glyphsLib/issues/280
-        if font_uses_new_axes(self.font):
+        elif font_uses_new_axes(self.font):
```
anthrotype, line 179: i suggest renaming `font_uses_new_axes` to `font_uses_axis_locations` as that method is no longer "new", but de facto superseded by Axis Mappings approach

Lib/glyphsLib/builder/axes.py
```diff
@@ -282,20 +300,10 @@ def to_glyphs_axes(self):
         self.font.customParameters["Axes"] = axes_parameter
 
     if self.minimize_ufo_diffs:
-        # TODO: (jany) later, when Glyphs can manage general designspace axes
```
anthrotype, line 285: Should we not set the Axis Mappings custom parameter in the UFO->Glyphs conversion unconditionally, instead of only when `minimize_ufo_diffs` is True? What's the reasoning behind not wanting to set that all the time?

Also I'm not sure if we should keep the TODO and the commented-out code below. cc @belluzj

Lib/glyphsLib/builder/axes.py
```diff
@@ -164,11 +164,10 @@ def to_designspace_axes(self):
             if axis.tag in custom_mapping:
                 mapping = {float(k): v for k, v in custom_mapping[axis.tag].items()}
                 regularDesignLoc = axis_def.get_design_loc(regular_master)
-                reverse_mapping = {v: k for k, v in mapping.items()}
-                if regularDesignLoc in reverse_mapping:
-                    regularUserLoc = reverse_mapping[regularDesignLoc]
-                else:
-                    regularUserLoc = min(mapping)
+                # Glyphs masters don't have a user location, so we compute it by
```
anthrotype, line 173: I think this comment "Glyphs masters don't have a user location" copy pasted from the other if branch may be confusing in the Axis Mappings if branch. Please rephrase or drop

Lib/glyphsLib/builder/axes.py
```diff
@@ -141,7 +141,13 @@ def to_designspace_axes(self):
     assert isinstance(regular_master, classes.GSFontMaster)
 
     custom_mapping = self.font.customParameters["Axis Mappings"]
-
+    if not custom_mapping:
+        logger.warning(
```
anthrotype, line 145: hm warning that may be too noisy. I'd suggest an INFO level, and much shorter than this, ideally one line.

m4rc1e: How about:

''Axis Mappings' customParameter not found. Add this parameter to define an explicit mapping."


Personally, I prefer longer logs in these tools because most users won't dig into the source code so it may be beneficial for them to know what's happening. Perhaps some simple docs would be better.

anthrotype: i think documentation doesn't belong in the logging. There is nothing technically wrong with one using Axis Locations or the instance inference as they are working now, if one doesn't need the extra customization provided by Axis Mappings, so the WARNING is too much. Even INFO seems out of place for me.
This new feature can be sort of experimental at the beginning, at least until things settle and Glyphs.app has implemented it in a similar way.
I suggest we remove any such recommendations, but work on documenting how this Axis Mappings is supposed to work and put a link somewhere. Adding full automated sphinx/ReadTheDocs etc. support for glyphsLib would be quite an effort and haven't got cycles for that now.

m4rc1e: How about we just drop the logging of this feature completely for the time being? or we could simply do logger.info("Using 'Axis Mapping' customParameter")?

anthrotype: yes, that's what I'm say. Let's not log anything for now. Or if you like just log what method is being used to obtain the user:design mapping among the three supported ones.

anthrotype (approved):

Lib/glyphsLib/builder/axes.py
m4rc1e: Thanks Cosimo. I'll rip it out.

Re docs, sorry I didn't mean a full blown sphinx setup. I'm more interested in a high level overview that explains how the full fontmake stack works. Such a doc would help most font devs who are not programmers diagnose common issues or at least understand where problems may occur.

anthrotype:
i'd be ok with merging this, perhaps flag as experimental in release notes (since Glyphs.app implementation may not be 100% same). But it'd be nice if @belluzj or @madig could also do another review pass.

Lib/glyphsLib/builder/axes.py
belluzj: I agree that this should happen unconditionally, the idea behind "minimize_ufo_diffs" was to store in Glyphs some information that is only useful for minimizing diffs but that is not impacting functionality (like the order of keys in a dictionary for example) but this data about the axis mapping is not a detail; it's a core functionality of variable fonts, so it should be translated all the time.

Thank you very much for doing my TODOs :D yay for open-source!

belluzj (approved):
LGTM except for what Cosimo pointed out about "minimize_ufo_diffs". Thanks @m4rc1e for making this happen, it's going to be very useful.

Lib/glyphsLib/builder/axes.py
m4rc1e: Ok, I've removed the conditional but I've also made sure that we only include the "Axis Mapping" customParameter if the designspace has map elements. This functionality is similar to when we convert a glyphs file to a designspace. It will only include map elements if they're actually useful.

m4rc1e:
@madig friendly ping. I've made one last commit since Cosimo and Jany both approved.

PR #622, opened 2020-09-22
Fix infinite loop. Fixes #621

### LINKED ISSUE #621: Infinite loop in GSPath.segments
Glyphs paths *should not* start with a `curve` node, but *sigh* sometimes they do.

foo

When this happens, `nodeCount` is never advanced in `GSPath.segments` and the loop is never terminated.

simoncozens:


simoncozens:
Wait, something's not right.

anthrotype:
got a test?

PR #623, opened 2020-09-25
Adding jobs to support PPC64LE architecture

ezeeyahoo:
Hi,

I have made changes to add support for PPC64LE architecture, request you to please review it. If it looks good, should I add jobs for other python version too?

Thanks

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,7 @@ branches:
   only:
     - master
     - /^v\d+\.\d+.*$/
+    - ezeeyahoo-patch-1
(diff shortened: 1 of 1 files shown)


anthrotype:
Sorry but I'm not familiar with this architecture. I'm open to accept patches to make glyphsLib work with it (we are pure python module so it should just work), but I don't think I want to add this to the CI matrix just yet.

madig:
Not convinced we need architecture-specific testing for a pure Python lib.

Issue #257, opened 2017-10-27
glyphsLib.affine module vs fontTools.misc.transform?


anthrotype:
Back in June, as part of the huge classtree branch, @yanone added this "affine" module, for affine transformation matrices:
https://github.com/googlei18n/glyphsLib/blob/master/Lib/glyphsLib/affine/__init__.py

At the top it says the module is "derived from Casey Duncan's Planar package".

This looks to me similar in scope and intent to the `fontTools.misc.transform` module:
https://github.com/fonttools/fonttools/blob/master/Lib/fontTools/misc/transform.py

I wonder whether we could just use fonttools' module (which already is a dependency of glyphsLib), or extend the latter if it is lacking any functionality that's required by glyphsLib.

belluzj:
When I was writing tests, I noticed that the matrix computation did not look very precise (but I did not investigate further): https://github.com/googlei18n/glyphsLib/blob/master/tests/writer_test.py#L825-L855

anthrotype:
would be good to get rid of that piece of extra code sometime.

PR #625, opened 2020-10-09
Fix applyTransform for paths

simoncozens:


anthrotype (approved):
LGTM

There's a few autoformatting issues, please run `tox -e lint` (which will run black via pre-commit) and commit the changes, thanks!

anthrotype:
this reminds me that some day we should replace the`glyphsLib.affine` with `fontTools.misc.transform`, as I don't see any reason to have that code (with its separate copyright/licence) https://github.com/googlefonts/glyphsLib/issues/257

PR #627, opened 2020-10-13
Fix out-of-order array access in bezierminmax

simoncozens:


anthrotype:
this `bezierMinMax` method, is it part of Glyphs.app's Python API, or is it a private helper for computing the tightest bounding box? 
I would rather reuse fontTools BoundsPen so we have less code to maintain..

anthrotype (approved):
anyway, the specific code changes LGTM

simoncozens:
It's computing the bounds of a segment. Yes, used only in the GSPath bbox code, so that probably should be replaced with a pen.

Issue #597, opened 2020-03-10
GlyphsRT test failure: test_n025_noto_moyogo_v1071_NotoSansDisplay-ItalicMM.glyphs


brawer:
```
test_n025_noto_moyogo_v1071_NotoSansDisplay-ItalicMM.glyphs (tests.run_various_tests_on_various_files.GlyphsRT) ... --- <expected>
+++ <actual>
@@ -594,7 +594,7 @@
 summation,
 minus,
 radical,
-infinity,
+"infinity",
 integral,
 approxequal,
 notequal,
FAIL
test_n026_noto_moyogo_v1071_NotoSansDisplay-MM.glyphs (tests.run_various_tests_on_various_files.GlyphsRT) ... --- <expected>
+++ <actual>
@@ -594,7 +594,7 @@
 summation,
 minus,
 radical,
-infinity,
+"infinity",
 integral,
 approxequal,
 notequal,
FAIL
```

brawer:
Actually there’s more breakages; see https://github.com/googlefonts/glyphsLib/issues/600.

Issue #557, opened 2019-09-24
Glyphs GSLayer vertOrigin and vertWidth aren't being saved


punchcutter:
For vertical layout vertOrigin and vertWidth can be set in Glyphs, but these values aren't being used in glyphsLib. When going to defcon we need verticalOrigin to be set properly so that the vmtx table can be built correctly. 

Also, in Glyphs if these values aren't explicitly set on some glyphs we still get values in the vmtx table. In both cases (explicitly set or not set) we are not getting the right values when running fontmake so vertical layout is completely mangled.

punchcutter:
@anthrotype @schriftgestalt I'm not entirely sure what Glyphs is doing here, but making this "work" in glyphsLib gives completely different results with the calculations. 

Here's what I added to glyphsLib/builder/builders.py to save these values in the UFO:

```
    def masters(self):
        """Get an iterator over master UFOs that match the given family_name.
        """
        if self._sources:
            for source in self._sources.values():
                yield source.font
            return

        # Store set of actually existing master (layer) ids. This helps with
        # catching dangling layer data that Glyphs may ignore, e.g. when
        # copying glyphs from other fonts with, naturally, different master
        # ids. Note: Masters have unique ids according to the Glyphs
        # documentation and can therefore be stored in a set.
        master_layer_ids = {m.id for m in self.font.masters}

        # stores background data from "associated layers"
        supplementary_layer_data = []

        # TODO(jamesgk) maybe create one font at a time to reduce memory usage
        # TODO: (jany) in the future, return a lazy iterator that builds UFOs
        #     on demand.
        self.to_ufo_font_attributes(self.family_name)

        # Generate the main (master) layers first.
        for glyph in self.font.glyphs:
            for layer in glyph.layers.values():
                if layer.associatedMasterId != layer.layerId:
                    # The layer is not the main layer of a master
                    # Store all layers, even the invalid ones, and just skip
                    # them and print a warning below.
                    supplementary_layer_data.append((glyph, layer))
                    continue

                ufo_layer = self.to_ufo_layer(glyph, layer)
                ufo_glyph = ufo_layer.newGlyph(glyph.name)
               +ufo_glyph.verticalOrigin = layer.vertOrigin
               +ufo_glyph.height = layer.vertWidth
                self.to_ufo_glyph(ufo_glyph, layer, glyph)

```
The calculation for the vmtx `advanceHeight` and `topSideBearing` is then done in ufo2ft 

```
[...]

PR #629, opened 2020-11-03
Improve vertical typesetting

m4rc1e:
This pr will solve #597 

My implementation is based on Cosimo's https://github.com/googlefonts/glyphsLib/issues/557#issuecomment-667074856.

### DIFF
--- a/Lib/glyphsLib/builder/custom_params.py
+++ b/Lib/glyphsLib/builder/custom_params.py
@@ -322,9 +322,9 @@ GLYPHS_UFO_CUSTOM_PARAMS = (
     # ('widthClass', 'openTypeOS2WidthClass'),
     # ('winAscent', 'openTypeOS2WinAscent'),
     # ('winDescent', 'openTypeOS2WinDescent'),
-    ("vheaVertTypoAscender", "openTypeVheaVertTypoAscender"),
-    ("vheaVertTypoDescender", "openTypeVheaVertTypoDescender"),
-    ("vheaVertTypoLineGap", "openTypeVheaVertTypoLineGap"),
+    ("vheaVertAscender", "openTypeVheaVertTypoAscender"),
+    ("vheaVertDescender", "openTypeVheaVertTypoDescender"),
+    ("vheaVertLineGap", "openTypeVheaVertTypoLineGap"),
     # Postscript parameters
     ("blueScale", "postscriptBlueScale"),
     ("blueShift", "postscriptBlueShift"),
--- a/Lib/glyphsLib/builder/custom_params.py
+++ b/Lib/glyphsLib/builder/custom_params.py
@@ -325,6 +325,9 @@ GLYPHS_UFO_CUSTOM_PARAMS = (
     ("vheaVertAscender", "openTypeVheaVertTypoAscender"),
     ("vheaVertDescender", "openTypeVheaVertTypoDescender"),
     ("vheaVertLineGap", "openTypeVheaVertTypoLineGap"),
+    ("vheaVertTypoAscender", "openTypeVheaVertTypoAscender"),
+    ("vheaVertTypoDescender", "openTypeVheaVertTypoDescender"),
+    ("vheaVertTypoLineGap", "openTypeVheaVertTypoLineGap"),
     # Postscript parameters
     ("blueScale", "postscriptBlueScale"),
     ("blueShift", "postscriptBlueShift"),
--- a/Lib/glyphsLib/builder/glyph.py
+++ b/Lib/glyphsLib/builder/glyph.py
@@ -126,6 +126,26 @@ def to_ufo_glyph(self, ufo_glyph, layer, glyph):
     else:
         ufo_glyph.width = width
 
+    if self.is_vertical:
+        if "typoAscender" in layer.master.customParameters:
+            ascender = layer.master.customParameters["typoAscender"]
+        else:
+            ascender = layer.master.ascender
+        if "typoDescender" in layer.master.customParameters:
+            descender = layer.master.customParameters["typoDescender"]
+        else:
+            descender = layer.master.descender
+
+        if layer.vertWidth:
+            ufo_glyph.height = layer.vertWidth
+        else:
+            ufo_glyph.height = ascender - descender
+
+        if layer.vertOrigin:
+            ufo_glyph.verticalOrigin = ascender - layer.vertOrigin
+        else:
+            ufo_glyph.verticalOrigin = ascender
+
     self.to_ufo_background_image(ufo_glyph, layer)
     self.to_ufo_guidelines(ufo_glyph, layer)
     self.to_ufo_glyph_background(ufo_glyph, layer)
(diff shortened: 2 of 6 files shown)


Lib/glyphsLib/builder/glyph.py
```diff
@@ -126,6 +126,26 @@ def to_ufo_glyph(self, ufo_glyph, layer, glyph):
     else:
         ufo_glyph.width = width
 
+    if self.is_vertical:
+        if "typoAscender" in layer.master.customParameters:
+            ascender = layer.master.customParameters["typoAscender"]
+        else:
+            ascender = layer.master.ascender
```
m4rc1e, line 133: I checked Glyphsapp and if the typoAscender or typoDescender are missing, glyphsApp will use the master.ascender and master.descender instead. It doesn't use the vhea customParameters.

Lib/glyphsLib/builder/custom_params.py
```diff
@@ -322,9 +322,9 @@ def register(handler):
     # ('widthClass', 'openTypeOS2WidthClass'),
     # ('winAscent', 'openTypeOS2WinAscent'),
     # ('winDescent', 'openTypeOS2WinDescent'),
-    ("vheaVertTypoAscender", "openTypeVheaVertTypoAscender"),
```
m4rc1e, line 325: @belluzj Are these misspelled or did these previously work? in Glyphsapp Version 2.6.4 (1286), these names are not there.

anthrotype: maybe @schriftgestalt knows?

Lib/glyphsLib/builder/glyph.py
```diff
@@ -256,6 +257,41 @@ def to_glyphs_glyph(self, ufo_glyph, ufo_layer, master):  # noqa: C901
     self.to_glyphs_paths(ufo_glyph, layer)
     self.to_glyphs_components(ufo_glyph, layer)
     self.to_glyphs_glyph_anchors(ufo_glyph, layer)
+    self.to_glyphs_glyph_height_and_vertical_origin(ufo_glyph, master, layer)
+
+
+def to_ufo_glyph_height_and_vertical_origin(self, ufo_glyph, layer):
+    # implentation based on:
+    # https://github.com/googlefonts/glyphsLib/issues/557#issuecomment-667074856
+    if not self.is_vertical:
```
anthrotype, line 266: I would rather place this check for `is_vertical` immediately before the method is called, and assume (assert) it is within the body of the current method

Lib/glyphsLib/builder/glyph.py
```diff
@@ -256,6 +257,41 @@ def to_glyphs_glyph(self, ufo_glyph, ufo_layer, master):  # noqa: C901
...
+    else:
+        ufo_glyph.height = ascender - descender
+
+    if layer.vertOrigin:
+        ufo_glyph.verticalOrigin = ascender - layer.vertOrigin
+    else:
+        ufo_glyph.verticalOrigin = ascender
+
+
+def _vert_typesetting_metrics(master):
```
anthrotype, line 282: this method is really only getting the typo ascender and descender, maybe name it `_get_typo_ascender_descender`?

anthrotype (commented):
this is looking good. Have you tried to see if you can build Noto Mongolian after this patch? see linked issue https://github.com/googlefonts/noto-fonts/issues/1414

Lib/glyphsLib/builder/glyph.py
schriftgestalt: the short form of the vhea values is without the Typo: vheaVertAscender, vheaVertDescender, vheaVertLineGap

anthrotype: Thanks. Sono guess we should support both

m4rc1e:
> you tried to see if you can build Noto Mongolian after this patch? see linked issue [googlefonts/noto-fonts#1414](https://github.com/googlefonts/noto-fonts/issues/1414)

@anthrotype, I've just made a quick build using a custom version of fontmake that uses this pr. I built the font using the following command:

`fontmake -g NotoSansMongolian.glyphs --mti-source NotoSansMongolian.plist -o ttf`

I don't really know what every glyph's orientation should be. Some of them still look incorrect. However, this PR should only fix the heights of these glyphs when set vertically. To my eyes, this seems to be working.


Screenshot 2020-11-06 at 11 19 22

_Green: this PR, Red: current implementation (Chrome Version 86.0.4240.183)_

If I look at both font's vmtx tables, I can confirm the heights are no longer defaulting to 0.


Screenshot 2020-11-06 at 11 31 29

_left: current implementation, right: this pr_

Finally, if I generate the NotoSansMongolian.glyphs file from glyphs and compare it against the quick build. The vmtx tables are identical (apart from nonmarking return getting renamed to CR)

Screenshot 2020-11-06 at 11 40 05

Lib/glyphsLib/builder/glyph.py
m4rc1e: Thanks Georg.

Cosimo, if I move `if self.is_vertical` to the parent function, I get the following flake8 error:

`Lib/glyphsLib/builder/glyph.py:29:1: C901 'to_ufo_glyph' is too complex (20)`

I know the `to_ufo_glyph` function is a smorgasbord of conditionals but I feel this test is perhaps a bit too limiting?

Lib/glyphsLib/builder/custom_params.py
m4rc1e: Ok, I've just readded them. 

If a font doesn't contain either sets of these metrics and the font is vertical, the short names will be added when converting to a ufo.

Lib/glyphsLib/builder/glyph.py
anthrotype: you can probably mute flake8 for that function only. Try adding `# noqa: C901` to disable that check for this particular method.

anthrotype:
> Finally, if I generate the NotoSansMongolian.glyphs file from glyphs and compare it against the quick build. The vmtx tables are identical

that's great! thanks for checking

anthrotype (approved):

m4rc1e:
@belluzj fancy reviewing this?

belluzj:
I will have a look but I've never done any vertical typesetting myself so I might not spot "functional" issues

Lib/glyphsLib/builder/custom_params.py
belluzj: I'm not sure. I may have assumed that anything possible on the UFO had the matching name on the Glyphs side without the openType prefix? I can't remember

m4rc1e: Georg posted in another comment (hidden since it's resolved) that the params with "Typo" are the long names, whilst the names without are the shorter versions. I've added both names so this shouldn't be an issue.

anthrotype: https://github.com/googlefonts/glyphsLib/pull/629#discussion_r518307755

belluzj:
LGTM, maybe in addition to your glyph-level test, you could have a second test for font-level information, such as setting the default values in the UFO fontinfo?

m4rc1e:
@belluzj SGTM. I've just added the check.

anthrotype (approved):

anthrotype:
can this be merged?

m4rc1e:
If Jany is happy then yes.

PR #632, opened 2020-11-06
Use Proxy for GSFont.features/featurePrefixes

khaledhosny:
Matches Glyphs App API and for consistency with the rather identical GSFont.classes.

### DIFF
--- a/tests/classes_test.py
+++ b/tests/classes_test.py
@@ -1749,5 +1749,23 @@ class FontGlyphsProxyTest(unittest.TestCase):
             del self.font.glyphs[self.font]
 
 
+class FontClassesProxyTest(unittest.TestCase):
+    def setUp(self):
+        self.font = GSFont(TESTFILE_PATH)
+
+    def test_indxing_by_name(self):
+        assert "Languagesystems" in self.font.featurePrefixes
+        assert "c2sc_source" in self.font.classes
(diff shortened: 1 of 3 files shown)


tests/classes_test.py
```diff
@@ -1749,5 +1749,23 @@ def test_remove_glyphs(self):
             del self.font.glyphs[self.font]
 
 
+class FontClassesProxyTest(unittest.TestCase):
+    def setUp(self):
+        self.font = GSFont(TESTFILE_PATH)
+
+    def test_remove_glyphs(self):
```
schriftgestalt, line 1756: Is this a good name for the test?

madig:
Out of curiosity, is there something that didn't work before but does with this PR?

khaledhosny:
> Out of curiosity, is there something that didn't work before but does with this PR?

Without this PR they are simple lists, so you can’t access them by name, like in the examples in https://docu.glyphsapp.com/#GSFont.features. I was merging two fonts and trying to merge `Languagesystems` prefix instead of having two of it when I found this limitation. It can be worked around of course.

tests/classes_test.py
khaledhosny: Sorry, copy&paste.

anthrotype:
thanks

Issue #578, opened 2020-01-29
GDEF table does not include the bracket glyphs; no mark/mkmk generated


anthrotype:
when converting from .glyphs to UFO, glyphsLib generates a GDEF GlyphClassDef table (with Base, Mark, Ligature glyphs) based on the GlyphData.xml properties.
Now, ufo2ft uses a GDEF table (if present in the UFO features.fea) when generating mark and mkmk features. If a GDEF table definition is present in features.fea, then it only considers as bases/marks/ligatures the glyphs that are listed in the GDEF; those that aren't are simply excluded from the mark and mkmk lookups.

The problem is that, even if bracket layers contain some anchors (and the user expects these anchors to be used for generating mark/mkmk), glyphsLib is not considering these when auto-generating the GDEF table definition. This is because it generates this GDEF _before_ it actually converts the bracket _layers_ to standalone UFO glyphs. As it generates the GDEF, the bracket "glyphs" are not there yet, so they are not considered; by the time it has applied the bracket layers (copying them over to standalone glyphs), it is too late for them to be included in GDEF.

We need to reorder the logic here and make sure that GDEF contains the newly added bracket glyphs.

PR #640, opened 2020-11-19
apply master custom params _after_ writing master UFO features

anthrotype:
... otherwise, the 'Replace Feature' custom parameter will not work for masters, as it finds the features.fea text empty.

Similar to https://github.com/googlefonts/glyphsLib/pull/571 but rebased off latest master.

### DIFF
--- a/Lib/glyphsLib/builder/builders.py
+++ b/Lib/glyphsLib/builder/builders.py
@@ -256,12 +256,20 @@ class UFOBuilder(_LoggerMixin):
                 ufo_glyph = ufo_layer.newGlyph(glyph.name)
                 self.to_ufo_glyph(ufo_glyph, layer, layer.parent)
 
-        for source in self._sources.values():
+        for master_id, source in self._sources.items():
             ufo = source.font
             if self.propagate_anchors:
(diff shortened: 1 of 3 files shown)


khaledhosny:
Isn’t this the same as https://github.com/googlefonts/glyphsLib/pull/571?

moyogo:
> Isn’t this the same as #571?

It is but rebased and better :-)

@cosimo do you want tests?

anthrotype:
Yes, in fact I linked to it in my previous comment. The difference is that this PR is based on updated master branch, thus takes into account the need to update GDEF after bracket glyphs are generated (https://github.com/googlefonts/glyphsLib/issues/578)

anthrotype:
@moyogo thanks Denis, if you do have some tests, feel free to push here or pass them along

khaledhosny:
I need to learn how to read.

madig:
@moyogo do you happen to have a test lying around?

punchcutter:
Any chance to get this in one of these days? One Noto update I've been sitting on absolutely requires this for the variable font and testing with this branch it all works as expected.

PR #641, opened 2020-11-24
Implement "Link Metrics..." custom parameters

### LINKED ISSUE #256: Custom parameter "Link Metrics With Master" support is not complete
Presence of "Link Metrics With Master" or "Link Metrics With First Master" means that the masters that have set this custom parameter must also inherit the kerning from the referenced master.

At the moment, using this parameter results in the dependent UFOs/fonts having no kerning at all.

jenskutilek:
If the custom parameters "Link Metrics To Master" or "Link Metrics To First Master" are present, this will apply glyph layer widths and kerning from the referenced source master.

Fixes #256.


Lib/glyphsLib/builder/glyph.py
```diff
@@ -114,7 +114,53 @@ def to_ufo_glyph(self, ufo_glyph, layer, glyph):  # noqa: C901
...
+                if master.name == source_master_id:
+                    found = True
+                    source_layer = self.font.glyphs[glyph.name].layers[master.id]
+                    break
+        else:
+            found = True
+        if found:
+            width = source_layer.width
+        else:
+            logger.debug(
```
simoncozens, line 157: Could you switch these log levels (debug here, info later in the same file) around? If we're referencing a master and we don't have it, that's a problem and I want to know about it; if we're applying the width from a linked master, that's expected behaviour and I don't need to know about it.

Lib/glyphsLib/builder/glyph.py
```diff
@@ -114,7 +114,53 @@ def to_ufo_glyph(self, ufo_glyph, layer, glyph):  # noqa: C901
...
+            # Try by name instead of master ID
+            found = False
+            for master in self.font.masters:
+                if master.name == source_master_id:
+                    found = True
+                    source_layer = self.font.glyphs[glyph.name].layers[master.id]
+                    break
+        else:
+            found = True
+        if found:
```
simoncozens, line 154: You don't need to keep track of a separate `found` variable here. Just check whether `source_layer` is now defined.

Lib/glyphsLib/builder/glyph.py
```diff
@@ -114,7 +114,53 @@ def to_ufo_glyph(self, ufo_glyph, layer, glyph):  # noqa: C901
...
+            width = source_layer.width
+        else:
+            logger.debug(
+                f"{layer.parent.name}: Trying to apply width from master "
+                f"'{source_master_id}': master not found."
+            )
+            # Use the current layer width
+            width = layer.width
+        if layer.width != width:
+            logger.info(
```
simoncozens, line 164: Make this debug

Lib/glyphsLib/builder/kerning.py
```diff
@@ -15,12 +15,50 @@
...
+                    if source_master.name == source_master_id:
+                        found = True
+                        source_master_id = source_master.id
+                        break
+            else:
+                found = True
+            if found:
+                kerning_source_masters[master_id] = source_master_id
+            else:
+                print(f"Source master for kerning not found: '{source_master_id}'")
```
simoncozens, line 51: Make this logger.info please.

simoncozens (changes requested):
Thanks so much for this! I spoke to someone today who needed this functionality.

What you have here works fine, but it's a bit clumsy - a lot of repetition of the source-finding code. Instead, we could have a `metrics_source` method on a `GSFontMaster` object which returns the right master - either the master itself if no custom parameters are set, or the correct master/first master based on custom parameter.

Then you only need to do that search in the one place, and you can get the width information from `self.font.glyphs[glyph.name].layers[master.metrics_source].width` and the kerning information from `self.font.masters[master_id].metrics_source.kerning`.

Let me know if you want any help implementing that, or if you don't have time to make the changes, let me know and I'll do it.

simoncozens:
I've finished off Jens' PR. Would be good for another pair of eyes to OK my changes.

simoncozens:
@googlebot I consent.

Lib/glyphsLib/builder/glyph.py
```diff
@@ -123,7 +123,22 @@ def to_ufo_glyph(self, ufo_glyph, layer, glyph):  # noqa: C901
...
+    # The width may be taken from another master via the customParameters
+    # 'Link Metrics With Master' or 'Link Metrics With First Master'.
+    master = self.font.masters[layer.associatedMasterId or layer.layerId]
+    metric_source = master.metricsSource.id
+    metric_layer = self.font.glyphs[glyph.name].layers[metric_source]
+    if metric_layer:
+        width = metric_layer.width
+    else:
+        width = None
+    if layer.width != width:
```
anthrotype, line 136: you're only applying it if width is not None (or if metric_layer is true); the `!=` may catch the case when width is None and layer.width is not None and trigger the logging debug message, when in fact you haven't applied anything because width was None. I think the logger.debug() call should be moved in the `if metric_layer:` branch after you override `width = metric_layer.width`

Lib/glyphsLib/builder/kerning.py
```diff
@@ -14,15 +14,22 @@
 
 
 import re
+import logging
 
 from .glyph import BRACKET_GLYPH_RE
 
+logger = logging.getLogger(__name__)
```
anthrotype, line 21: where is this logger being used?

tests/link_metrics_test.py
```diff
@@ -0,0 +1,49 @@
...
+import unittest
+
+from glyphsLib import load_to_ufos
+
+
+def glyphs_file_path():
+    return os.path.join(os.path.dirname(__file__), "data", "LinkMetrics.glyphs")
+
+
+class UtilTest(unittest.TestCase):
```
anthrotype, line 27: you are using plain `assert .. == ...` like in pytest-style tests but then you're sublcassing from unittest.TestCase.. Either you subclass TestCase and then use unittest-style self.assertEqual(etc.) or you simply use pytest-style testing and remove the subclassing from unittest.TestCase. You can still group pytest tests as instance methods by creating the test class and naming it such that it _starts_ with the string "Test" (see https://docs.pytest.org/en/reorganize-docs/new-docs/user/naming_conventions.html). We can configure pytest so that it also recognizes class names that ends with "Test" but it's simpler if you rename it `TestUtil`.
I prefer that you use pytest-style for newly added tests.
The test runner in pytest anyway and it understand both. But mixing ways the two like you do here is incorrect.

anthrotype:
@jenskutilek the googlebot wants you to also reply "I consent" before it's happy

simoncozens:
Thanks, good catches.

simoncozens:
> @jenskutilek the googlebot wants you to also reply "I consent" before it's happy

No, it's happy enough now, because @jenskutilek was the PR originator.

anthrotype:
> No, it's happy enough now

yeah you're right

anthrotype (approved):
LGTM

jenskutilek:
Thank you @simoncozens for picking this up!

PR #644, opened 2020-12-02
Improve compatibility with Glyphs

simoncozens:
When trying to run a Glyphs script on the command line, I noticed that I couldn't set a node's position. Glyphs allows `n.position = (x,y)`, but this creates a tuple when GSNode expects a `Point` object. But of course I can't use a `Point` object because then the script wouldn't work in Glyphs...

This PR allows setting `n.position = (x,y)` and automatically upcasting that to a `Point` object.

### DIFF
--- a/Lib/glyphsLib/classes.py
+++ b/Lib/glyphsLib/classes.py
@@ -1612,11 +1612,13 @@ class GSNode(GSBase):
     )
     _parent = None
 
-    def __init__(self, position=(0, 0), nodetype=LINE, smooth=False, name=None):
+    def __init__(self, position=(0, 0), type=LINE, smooth=False, name=None, nodetype=None):
         self._userData = None
         self._position = Point(position[0], position[1])
         self.smooth = smooth
-        self.type = nodetype
+        self.type = type
+        if nodetype is not None: # for backward compatibility
+            self.type = nodetype
         # Optimization: Points can number in the 10000s, don't access the userDataProxy
         # through `name` unless needed.
         if name is not None:
--- a/Lib/glyphsLib/classes.py
+++ b/Lib/glyphsLib/classes.py
@@ -1603,7 +1603,7 @@ class GSFontMaster(GSBase):
 
 
 class GSNode(GSBase):
-    __slots__ = ("_userData", "position", "smooth", "type")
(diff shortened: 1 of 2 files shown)


simoncozens:
I've added another commit since the explanation above. In the `GSNode` constructor in Glyphs, the parameter to set a node's type is (for better or worse) called `type`; glyphsLib calls it `nodetype`. This changes the parameter to match, but also allows for `nodetype` for compatibility.

Lib/glyphsLib/classes.py
```diff
@@ -1612,11 +1612,13 @@ class GSNode(GSBase):
     )
     _parent = None
 
-    def __init__(self, position=(0, 0), nodetype=LINE, smooth=False, name=None):
+    def __init__(self, position=(0, 0), type=LINE, smooth=False, name=None, nodetype=None):
```
anthrotype, line 1615: Why was it necessary to also rename the nodetype parameter?

Lib/glyphsLib/classes.py
```diff
@@ -1612,11 +1612,13 @@ class GSNode(GSBase):
...
         self._userData = None
-        self.position = Point(position[0], position[1])
+        self._position = Point(position[0], position[1])
         self.smooth = smooth
-        self.type = nodetype
+        self.type = type
+        if nodetype is not None: # for backward compatibility
```
anthrotype, line 1620: You could add a deprecation warning here if you want to rename the parameter

anthrotype (approved):

Lib/glyphsLib/classes.py
anthrotype: I just saw your comment, thanks

simoncozens: My thought process went: the actual constructor interface should be called `type`, to match Glyphs; but we should also support `nodetype` in case anyone's relying on `glyphsLib` having that interface.

anthrotype: that's fine. I was suggesting that we also do a `warnings.warn("nodetype is deprecated, please use type instead", DeprecationWarning)` and in a few releases we remove it altogether

PR #647, opened 2020-12-02
Run tests/publish with GitHub actions, fixes #645

### LINKED ISSUE #645: move to Github Actions
Travis CI is too slow, we need to transition to Github Actions like we did for ufo2ft and fonttools.
Any volunteers? We should basically copy over ufo2ft's setup

simoncozens:
Literally the same workflow files as ufo2ft. You will need to set `PYPI_USERNAME` and `PYPI_PASSWORD` in repository secrets.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -7,6 +7,18 @@ on:
     branches: [master]
 
 jobs:
+  lint:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up Python 3.8
+      uses: actions/setup-python@v2
+      with:
+        python-version: 3.8
+    - name: Install tox
+      run: pip install tox
+    - name: Run style and typing checks
+      run: tox -e lint
   build:
     runs-on: ${{ matrix.platform }}
     strategy:
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,0 +1,36 @@
+name: Test
+
+on:
+  push:
+    branches: [master]
+  pull_request:
+    branches: [master]
+
+jobs:
+  build:
+    runs-on: ${{ matrix.platform }}
+    strategy:
+      matrix:
+        python-version: [3.6, 3.9]
+        platform: [ubuntu-latest, windows-latest]
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up Python ${{ matrix.python-version }}
+      uses: actions/setup-python@v2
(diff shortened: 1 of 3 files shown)


.github/workflows/test.yml
```diff
@@ -0,0 +1,36 @@
...
+    - name: Install packages
+      run: pip install tox coverage
+    - name: Run Tox
+      run: tox -e py-cov
+    - name: Produce coverage files
+      run: |
+        coverage combine
+        coverage xml
+    - name: Upload coverage to Codecov
+      uses: codecov/codecov-action@v1
```
anthrotype, line 31: if we use this codecov action, then we no longer need the `[testenv:codecov]` section in tox.ini (we were running `tox -e codecov` on travis CI)
you can delete that 
https://github.com/googlefonts/glyphsLib/blob/6c409dd27cb46919c7fe10c5db2b77ce45751632/tox.ini#L27-L36

anthrotype (approved):
Thanks!
I'll add the encrypted pypi token after this is merged

anthrotype:
don't wait for travis, it'll never end. You can merge to master after you removed the unused codecov section in tox.ini. I think the actions will automatically trigger after the workflow configs are in master (they did not in this PR).

anthrotype:
I'll unplug Travis after we verify everything works

anthrotype (changes requested):
wait! you haven't set up the additional step for the `tox -e lint` that runs the autoformatter and linters. Please do that before merging, thank you!

anthrotype:
basically we want a `lint` step before the one you called `build` (it really does `test` more than `build` but who cares), similar to this one:

https://github.com/googlefonts/picosvg/blob/692a09befae8538be7d3a09ea591016604af799d/.github/workflows/ci.yml#L11-L23

(ignore the comments about pytype which don't apply for glyphsLib).
The idea is we run the static linting tools on a one python (e.g. 3.8 or 3.9) as a distinct step from the multi-python test matrix. We don't want to run `tox -e lint` for each one, as the result of the linting in theory is indepedent of the python version and it would be a waste of resources to do it multiple times.

anthrotype (approved):
thanks, LGTM

anthrotype:
I just added encrypted PYPI_{USERNAME,PASSWORD} to the repository's settings, so it should just work 🤞

anthrotype:
you can delete .travis.yml as well. I just removed the webhook

anthrotype:
i'll delete travis.yml

PR #648, opened 2020-12-03
Merge github workflows into one

anthrotype:
Make the deploy step dependent on the test step, so that we don't accidentally deploy if some tests failed.

~~Use tox-gh-actions plugin to automatically configure the tox environment(s) to run given the current GH Actions python version.~~

Automatically create a Github Release from the annotated git tag message and format the latter as markdown.


madig:
Not sure we need the tox plugin? `tox -e py` does the job and linting should be a separate, parallel job.

anthrotype:
FYI, I added instructions on how to make a release to the README:
https://github.com/googlefonts/glyphsLib#make-a-release

basically, you create annotated git tag on your local repo, write some release notes as part of the tag message, and push the tag to the remote repo (which will triggers the CI deploy step)

anthrotype:
our current workflow works with pre-releases too, I just pushed one
see https://github.com/googlefonts/glyphsLib/releases/tag/v5.3.0b1

PR #649, opened 2020-12-07
ufo2glyphs: preserve DS axis order

madig:
Previously, wght and wdth would be forcibly be made first.


belluzj:
LGTM

anthrotype (approved):
LGTM

maybe add an expilicit test with only "wght" + "wdth" that asserts that we do _not_ add the Axes custom parameter since it follows the implicit order already

anthrotype (approved):

PR #650, opened 2020-12-09
Glyphs 3 support

simoncozens:
This is very much a WIP; don't bother trying to use it yet, but you can have a look at what I'm doing and where I'm up so. So far it can read a Glyphs3 format file (but will probably fail to read Glyphs2 files now) but does not store all that boring information about "components" and "paths", like anyone actually cares about those things.

### DIFF
--- a/Lib/glyphsLib/classes.py
+++ b/Lib/glyphsLib/classes.py
@@ -3555,6 +3555,7 @@ class GSGlyph(GSBase):
 class GSFont(GSBase):
     __slots__ = (
         "DisplayStrings",
+        "_axes",
         "_classes",
         "_customParameters",
         "_featurePrefixes",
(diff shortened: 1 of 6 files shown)


schriftgestalt (commented):
I renamed name > tag to not confuse it with stylistic set names.

schriftgestalt (commented):
Renamed those to keep them together with default key sorting.

simoncozens:
This is a silly approach and I’ve thought of another one which is a whole lot cleaner. I plan to decouple the marshalling (GS* object to/from dict-like structure) from the serialisation (reading/writing different styles of plist). Classes will then have access to the whole dict and get greater flexibility about what they do with members.

Also we can make decisions while marshalling about whether to emit Glyphs2-style or Glyphs3-style data structures (eg whether to put paths and components in their own slots or in a common shapes slot). And having the serialisation step separate means we can plug in the openstep-plist library for glyphs 3 and keep the current parser/writer for Glyphs 2.

Yeah. That’s clearly the way to go.

Keeping this open for now but will close when I have made a bit more progress and have something to show for it.

anthrotype:
> decouple the marshalling (GS* object to/from dict-like structure) from the serialisation (reading/writing different styles of plist)

that was the original design, and what I also was proposing to back to when I made the openstep-plist parser

simoncozens:
-> #651

PR #651, opened 2020-12-11
Glyphs 3 support (again)

simoncozens:
Again a WIP. And I've decided to test it in small pieces, rather than try to parse a whole G3 file from the outset. Hacking is fun and all, but sometimes you've got to sit down and plan.

This has got the `GSGuide` and `Point` values working in both their G2 and G3 flavours. Nothing else yet, though.


simoncozens:
The OpenStep plist parser does not understand types; it would never be able to parse a custom parameter like our `AsteriskParameter = ( "754", "30", 7, 51, "80", "50" );`
We have to conclude that this is Glyphs format, not plist format.

anthrotype:
I don't understand, what is special about that custom parameter you quoted? What do you mean by "types"?

schriftgestalt:
It might be that the default implementation always reads strings. But that can be fixed.

anthrotype:
openstep_plist.load has `use_numbers=False` option that allows to prase unquoted strings that look like numbers as numbers, is that what you were looking for?
https://github.com/fonttools/openstep-plist/blob/b8ffae85ac07d04dc57be9b27e091b5632d9b152/tests/test_parser.py#L295-L305

simoncozens:
I'm giving up on this *again* and trying a more tractable approach. -> #652.

PR #653, opened 2020-12-18
Make sure that detaching bracket glyphs get their own production name

madig:
When a Glyphs glyph has a production name set, any bracket glyph detaching from it will inherit it unchanged, when it should also get its BRACKET suffix.

### DIFF
--- a/Lib/glyphsLib/builder/glyph.py
+++ b/Lib/glyphsLib/builder/glyph.py
@@ -14,16 +14,19 @@
 
 
 import logging
+import re
 
 import glyphsLib.glyphdata
 from .common import to_ufo_time, from_loose_ufo_time
 from .constants import GLYPHLIB_PREFIX, GLYPHS_COLORS, PUBLIC_PREFIX
+from .builders import BRACKET_GLYPH_RE
 
 logger = logging.getLogger(__name__)
 
 SCRIPT_LIB_KEY = GLYPHLIB_PREFIX + "script"
 ORIGINAL_WIDTH_KEY = GLYPHLIB_PREFIX + "originalWidth"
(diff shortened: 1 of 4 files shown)


schriftgestalt:
Wouldn't it make more sense to set the production name when the bracket glyph is created?

madig:
What do you mean?

schriftgestalt:
Maybe I misunderstand but the change seems to be that if will "correct" the production name later in the process (because it parses the glyph name). Instead I would set the production name at the same time when the glyph name is set.

madig:
This change is specifically for the Postscript name, which is historically set in that point of the code. Maybe there is a better way, but I'd leave that to a refactor of glyphsLib...

Lib/glyphsLib/builder/glyph.py
```diff
@@ -14,16 +14,19 @@
...
 from .common import to_ufo_time, from_loose_ufo_time
 from .constants import GLYPHLIB_PREFIX, GLYPHS_COLORS, PUBLIC_PREFIX
+from .builders import BRACKET_GLYPH_RE
 
 logger = logging.getLogger(__name__)
 
 SCRIPT_LIB_KEY = GLYPHLIB_PREFIX + "script"
 ORIGINAL_WIDTH_KEY = GLYPHLIB_PREFIX + "originalWidth"
 BACKGROUND_WIDTH_KEY = GLYPHLIB_PREFIX + "backgroundWidth"
+BRACKET_GLYPH_SUFFIX_RE = re.compile(r".*(\..*BRACKET\.\d+)$")
```
belluzj, line 29: This regex should be next to the other one `BRACKET_GLYPH_RE`, since I suppose they're related?

belluzj:
LGTM apart from one comment

PR #656, opened 2021-01-13
glyphdata: detect subcategory of smallcaps

madig:
@schriftgestalt is there any special logic to detecting the subcategory of `.sc` glyphs?


schriftgestalt:
If the there is a suffix ".sc", ".smcp" or ".c2sc".

madig:
Ok thanks, does this trigger for any script and category? As in, if those
substring are detected, subcategory is always set to small caps?

schriftgestalt:
In Glyphs 3, the smallcap setting is not in .subCategory any more but in the new .case (upper|lower|smallcap|minor) (the later is used for super/subscript and such). There where a lot cases where it was not clear what to use for subCategory (e.g. f_l.sc: either Ligature or SmallCap).

In Glyphs 2 it seems to be a bit more complicated. subCategory for f_l.sc is Smallcaps, for dieresiscomb.sc it is Nonspacing.

PR #657, opened 2021-01-22
Roundtrip bracket glyph kerning

madig:
Bracket glyphs inherit their parents' kerning.

### DIFF
--- a/Lib/glyphsLib/builder/builders.py
+++ b/Lib/glyphsLib/builder/builders.py
@@ -486,9 +486,8 @@ class UFOBuilder(_LoggerMixin):
         for glyph_name, glyph_bracket_layers in bracket_layer_map.items():
             for (location, reverse), layers in glyph_bracket_layers.items():
                 for layer in layers:
-                    ufo_font = ufo_layer = self._sources[
-                        layer.associatedMasterId or layer.layerId
-                    ].font
+                    layer_id = layer.associatedMasterId or layer.layerId
+                    ufo_font = self._sources[layer_id].font
                     ufo_layer = ufo_font.layers.defaultLayer
                     ufo_glyph_name = _bracket_glyph_name(glyph_name, reverse, location)
                     ufo_glyph = ufo_layer.newGlyph(ufo_glyph_name)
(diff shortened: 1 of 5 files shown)


Lib/glyphsLib/builder/builders.py
```diff
@@ -486,9 +486,10 @@ def _copy_bracket_layers_to_ufo_glyphs(self, bracket_layer_map):
         for glyph_name, glyph_bracket_layers in bracket_layer_map.items():
             for (location, reverse), layers in glyph_bracket_layers.items():
                 for layer in layers:
-                    ufo_layer = self._sources[
+                    ufo_font = ufo_layer = self._sources[
```
anthrotype, line 489: you forgot in that "ufo_layer =`, which you reassign to immediately afterwards

Lib/glyphsLib/builder/builders.py
```diff
@@ -506,6 +507,25 @@ def _copy_bracket_layers_to_ufo_glyphs(self, bracket_layer_map):
...
+                    for (first, second), value in ufo_font.kerning.items():
+                        first_match = first == glyph_name
+                        second_match = second == glyph_name
+                        if first_match and second_match:
+                            bracket_kerning[(ufo_glyph_name, ufo_glyph_name)] = value
+                        elif first_match:
+                            bracket_kerning[(ufo_glyph_name, second)] = value
+                        elif second_match:
+                            bracket_kerning[(first, ufo_glyph_name)] = value
+                    ufo_font.kerning.update(bracket_kerning)
```
anthrotype, line 528: maybe make this whole block a separate method that updates the kerning for a bracket glyph

Lib/glyphsLib/builder/groups.py
```diff
@@ -92,6 +93,7 @@ def to_glyphs_groups(self):
     groups = []
     for source in self._sources.values():
         for name, glyphs in source.font.groups.items():
+            glyphs = [name for name in glyphs if not BRACKET_GLYPH_RE.match(name)]
```
belluzj, line 96: Maybe put a comment here to explain why you're excluding these from the groups?

belluzj (approved):
Looks good to me. Does that whole thing mean that with Glyphs.app you can't have different kerning between a glyph and its bracket substitution?

madig:
I think so. @schriftgestalt ?

schriftgestalt:
Yes. The bracket layers are meant for smaller adjustments. For bigger changes and more control, you can make an extra glyph and write the feature variations yourself. (see https://glyphsapp.com/news/glyphs-3-make-things-you-love (search for: `OTVar feature variations directly in your feature code`).

madig:
Thanks for clarifying!

PR #659, opened 2021-01-27
check that vert{Width,Origin} are not None, can be 0

anthrotype:
Fixes https://github.com/googlefonts/glyphsLib/issues/557#issuecomment-768020307

### DIFF
--- a/Lib/glyphsLib/builder/font.py
+++ b/Lib/glyphsLib/builder/font.py
@@ -122,7 +122,8 @@ def _set_glyphs_font_attributes(self, source):
         # FIXME: (jany) should wrap in glyphs_datetime? or maybe the GSFont
         #     should wrap in glyphs_datetime if needed?
         font.date = from_ufo_time(info.openTypeHeadCreated)
-    font.upm = info.unitsPerEm
+    if info.unitsPerEm is not None:
+        font.upm = info.unitsPerEm
     if info.versionMajor is not None:
         font.versionMajor = info.versionMajor
     if info.versionMinor is not None:
--- a/Lib/glyphsLib/builder/glyph.py
+++ b/Lib/glyphsLib/builder/glyph.py
@@ -339,9 +339,9 @@ def _to_glyphs_color(color):
 
 
 def to_glyphs_glyph_height_and_vertical_origin(self, ufo_glyph, master, layer):
-    if ufo_glyph.height:
+    ascender, descender = _get_typo_ascender_descender(master)
+    if ufo_glyph.height != (ascender - descender):
         layer.vertWidth = ufo_glyph.height
 
-    if ufo_glyph.verticalOrigin:
-        ascender, _ = _get_typo_ascender_descender(master)
+    if ufo_glyph.verticalOrigin is not None and ufo_glyph.verticalOrigin != ascender:
         layer.vertOrigin = ascender - ufo_glyph.verticalOrigin
--- a/Lib/glyphsLib/builder/glyph.py
+++ b/Lib/glyphsLib/builder/glyph.py
@@ -277,12 +277,12 @@ def to_ufo_glyph_height_and_vertical_origin(self, ufo_glyph, layer):
 
     ascender, descender = _get_typo_ascender_descender(layer.master)
 
-    if layer.vertWidth:
+    if layer.vertWidth is not None:
         ufo_glyph.height = layer.vertWidth
     else:
         ufo_glyph.height = ascender - descender
 
-    if layer.vertOrigin:
+    if layer.vertOrigin is not None:
         ufo_glyph.verticalOrigin = ascender - layer.vertOrigin
     else:
         ufo_glyph.verticalOrigin = ascender
--- a/Lib/glyphsLib/builder/masters.py
+++ b/Lib/glyphsLib/builder/masters.py
@@ -118,10 +118,14 @@ def to_glyphs_master_attributes(self, source, master):
         # relative to anything.
         master.customParameters[UFO_FILENAME_CUSTOM_PARAM] = os.path.basename(ufo.path)
 
-    master.ascender = ufo.info.ascender
-    master.capHeight = ufo.info.capHeight
-    master.descender = ufo.info.descender
-    master.xHeight = ufo.info.xHeight
+    if ufo.info.ascender is not None:
+        master.ascender = ufo.info.ascender
+    if ufo.info.capHeight is not None:
+        master.capHeight = ufo.info.capHeight
+    if ufo.info.descender is not None:
+        master.descender = ufo.info.descender
+    if ufo.info.xHeight is not None:
+        master.xHeight = ufo.info.xHeight
 
     horizontal_stems = ufo.info.postscriptStemSnapH
     vertical_stems = ufo.info.postscriptStemSnapV
--- a/tests/builder/to_glyphs_test.py
+++ b/tests/builder/to_glyphs_test.py
@@ -261,7 +261,10 @@ def test_glyph_height_and_vertical_origin(ufo_module):
     a = ufo.newGlyph("a")
     a.height = 1024
     a.verticalOrigin = 100
-    ufo.newGlyph("b")
+    b = ufo.newGlyph("b")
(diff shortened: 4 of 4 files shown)


anthrotype:
I'm honestly not sure if this is the right fix. If anybody knows more, please chime in and comment /cc @m4rc1e @punchcutter

m4rc1e:
I will take a look later today.

punchcutter:
Georg told me once "If the manual set value is the same as the presumed default, it is omitted." So if it's 0 and Glyphs thinks it should be 0 then it won't explicitly add it. 
But if it's in the feature file or set in the Glyphs file it should be set so I think this looks good.

m4rc1e:
Alright, I've just taken a look at @kokiabe's glyphs file and genned some fonts. Here's a 3 way diff.


Screenshot 2021-01-27 at 11 14 51

**Left: GlyphsApp | Center: This PR | Right: Current glyphsLib implementation**

This PR fixes the issue. Thanks Cosimo!

m4rc1e:
I'll inspect what happens when we convert from ufo to glyphs after lunch. I have a hunch it may need a tweak as well.

m4rc1e:
If I modify `builder.glyph.to_glyphs_glyph_height_and_vertical_origin` (ufo -> glyphs procedure) so zero values for height and origin are also accepted, we get several failing tests.

They're failing because defcon will initialise a new glyph so its height attribute will have a value of 0, https://github.com/robotools/defcon/blob/master/Lib/defcon/objects/glyph.py#L122 e.g

```
>>> import defcon
>>> f = defcon.Font()
>>> g = f.newGlyph("a")
>>> g.height
0
```

What's really strange is that if I have an existing glif file which doesn't specify a height attribute e.g

```
<?xml version='1.0' encoding='UTF-8'?>
<glyph name="a" format="2">
  <advance width="606"/>
  <unicode hex="0061"/>
  <anchor x="248" y="0" name="bottom"/>
  <anchor x="494" y="0" name="ogonek"/>
  <anchor x="358" y="522" name="top"/>
  <outline>
    <contour>
      <point x="91" y="-10"/>
      <point x="208" y="-10" type="curve" smooth="yes"/>
      <point x="279" y="-10"/>
      <point x="359" y="31"/>
      <point x="411" y="98" type="curve"/>
      <point x="405" y="71"/>
      <point x="402" y="53"/>
      <point x="402" y="41" type="curve" smooth="yes"/>
      <point x="402" y="15"/>
      <point x="422" y="-4"/>
      <point x="459" y="-4" type="curve" smooth="yes"/>
      <point x="468" y="-4"/>
      <point x="479" y="-2"/>
      <point x="494" y="0" type="curve"/>
      <point x="510" y="76" type="line"/>
      <point x="495" y="77"/>
      <point x="488" y="82"/>
      <point x="488" y="95" type="curve" smooth="yes"/>
      <point x="488" y="99"/>
      <point x="489" y="103"/>
      <point x="490" y="108" type="curve" smooth="yes"/>
      <point x="578" y="522" type="line"/>
      <point x="490" y="522" type="line"/>
      <point x="467" y="418" type="line"/>
      <point x="445" y="481"/>
      <point x="391" y="531"/>
      <point x="311" y="531" type="curve" smooth="yes"/>
      <point x="159" y="531"/>
      <point x="22" y="374"/>
      <point x="22" y="207" type="curve" smooth="yes"/>
      <point x="22" y="89"/>
    </contour>
  </outline>
</glyph>
```

and I load this glyph in defcon and type `glyph.width` it will return `0`. Surely attributes shouldn't be accessible if they haven't been declared? my knowledge of ufo/defcon is flakey at best so I do apologise if I'm jumping to conclusions/not understanding the implementation,

m4rc1e:
> Surely attributes shouldn't be accessible if they haven't been declared? my knowledge of ufo/defcon is flakey at best so I do apologise if I'm jumping to conclusions/not understanding the implementation,

Looking at the ufo glif [spec](https://unifiedfontobject.org/versions/ufo3/glyphs/glif/#advance), it appears that the default for both height and width are `0`. However, it also says that height is optional if a width has been defined.

Since initing a new glyph or loading an existing glyph in Defcon will set the height attribute to 0, we can't use `if ufo_glyph.height is not None`. We can only use this if initing a new glyph doesn't add the height attribute by default. I could open an issue in defcon and propose we do this but it will probably break a boatload of test suites elsewhere etc.

Feel free to call out any assumptions I'm making. I'm no ufo expert.

punchcutter:
I'm a little confused about this because vertOrigin and vertWidth are not part of the "glyph" or any part of a UFO glif file. They are separate values used to determine the `vmtx` settings. They aren't even stored in a UFO because there's nowhere to store them. I mean obviously there are places, but I just did a test saving a Glyphs file (with a bunch of these values explicitly set) as UFO and everything disappears. I suppose both Glyphs and then glyphsLib should be storing these values in the lib.plist or somewhere. Either way it's also entirely possible to have these values (the vmtx table) defined in the feature file instead. That's how our Adobe CJK fonts do it. We define the values there and makeotf compiles them.

anthrotype:
yes, UFO default advance height is 0. It's not just defcon, it's the UFO spec.
https://github.com/unified-font-object/ufo-spec/issues/95#issuecomment-565425208

> vertOrigin and vertWidth are not part of the "glyph" or any part of a UFO glif file.

verticalOrigin is a lib key, whereas vertWidth corresponds to the `height` attribute of the `advance` element in the GLIF data structure.

anthrotype:
I fear that we may not be able to fully roundtrip these back when going from Glyphs -> UFO, because of the differing default values in each.

anthrotype:
(EDIT: in the previous message, I meant "from UFO -> Glyphs")

anthrotype:
actually! I think we can make it work. Instead of checking "if ufo_glyph.height is not None: gs_glyph.vertWidth = ufo_glyph.height", we need to check `if ufo_glyph.height != ascender - descender:` which is the default value according to which Glyphs.app decides whether to store vertWidth in the .glyphs file or not.

schriftgestalt:
Would that add a lot of `vertwidth = 0` if you convert a .ufo to .glyphs that wasn't exported from Glyphs?

It would help if the default in the ufo spec would be `None` instead of `0`. Is it even possible to export vertical spacing from a ufo? So it either export all zeros or omits nonspacing glyphs. What triggers the export of a `vmtx` table from the data in height field? If that value is not used to generate `vmtx`, we can ignore it and pick a lib key with a usable default?
I can add support for `vertOrigin/Width` lib keys in the ufo im/export in Glyphs.

kokiabe:
Thanks everyone! When converting from `.glyphs` to `ttf`, the `fix-vert-zero branch` is working as I expected.
https://github.com/fontworks-fonts/Train/issues/1

anthrotype:
> Would that add a lot of vertwidth = 0 if you convert a .ufo to .glyphs that wasn't exported from Glyphs?

Indeed, yes.. But I think that's the only way that we can make sure not to lose any data. Changing the default for glyph.height in the UFO speci is tricky since it'd be a breaking change.

> What triggers the export of a vmtx table from the data in height field?

in ufo2ft we check that these `openTypeVheaVert*` info attributes are set, and build vmtx/vhea/VORG accordingly
 https://github.com/googlefonts/ufo2ft/blob/74e1ee76555f09b24920d17c9afd26639b65270f/Lib/ufo2ft/outlineCompiler.py#L121-L130

schriftgestalt:
But those zero values will make the .glyphs files totally unusable. 
The vertWidth is not always saved to .ufo. What is the criteria to omit it? Can the ufo code that reads the .ufo keep track if there was a explicit value in the file? That wouldn't break much.

anthrotype:
> The vertWidth is not always saved to .ufo. What is the criteria to omit it? Can the ufo code that reads the .ufo keep track if there was a explicit value in the file? 

there's no vertWidth in UFO, there's glyph.height. It defaults to 0, and when it's zero ufoLib omits it. If there isn't an explicit height set, the value is thus understood to be 0, we can't change this for the current UFO3 version.

schriftgestalt:
But why should `glyph.height` default to 0. It would make much more sense to default to `None` (and making round-tripping possible).

PR #660, opened 2021-01-28
Fix commented class issue

simoncozens:
This adds a newline to before the closing bracket of a class definition. Why do we want to do this? Well, there's an obscure corner case I just hit today.

Glyphs allows comments in glyph classes like so:

Screenshot 2021-01-28 at 14 40 25

If the last entry in a glyph class is commented out, glyphsLib writes a UFO feature file like so:

```
@Test = [ A
# B ];
```

The closing bracket of the class is commented out, causing a syntax error.

This PR writes the feature as

```
@Test = [ A
# B
];
```

instead.

### DIFF
--- a/Lib/glyphsLib/builder/features.py
+++ b/Lib/glyphsLib/builder/features.py
@@ -78,7 +78,7 @@ def _to_ufo_features(font, ufo=None, generate_GDEF=False, skip_export_glyphs=Non
         prefix = "@" if not class_.name.startswith("@") else ""
         name = prefix + class_.name
         class_defs.append(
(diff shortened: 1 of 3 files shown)


simoncozens:
Bump @anthrotype. This just hit me again.

anthrotype:
ouch. sorry I didn't see this PR. Next time you can also ping @madig or @belluzj

anthrotype:
merged too quicky perhaps. Shouldn't we also add a newline after the opening bracket for consistency?

simoncozens:
Autogenerated feature files as part of a build pipeline are “binary equivalent”.  I don’t care what they look like, so long as they work.

PR #663, opened 2021-02-18
Close open corners in UFO

### LINKED ISSUE #255: Do something about outside open corners
When should they be removed?
* keep them in UFO and add an option in ufo2ft to remove them
* remove in UFO (but keep the original path somewhere to round-trip back)

Related:
https://forum.glyphsapp.com/t/how-to-detect-an-outside-open-corner/4117

simoncozens:
Requires https://github.com/fonttools/fonttools/pull/2192 to be merged. Fixes #255 and https://github.com/googlefonts/fontmake/issues/288

### DIFF
--- a/Lib/glyphsLib/builder/font.py
+++ b/Lib/glyphsLib/builder/font.py
@@ -88,11 +88,9 @@ def to_ufo_font_attributes(self, family_name):
 
         ufo.lib[MASTER_ORDER_LIB_KEY] = index
 
-        ufo.lib["com.github.googlei18n.ufo2ft.filters"] = [{
-            "namespace": "glyphsLib.filters",
-            "name": "eraseOpenCorners",
-            "pre": True
-        }]
+        ufo.lib["com.github.googlei18n.ufo2ft.filters"] = [
+            {"namespace": "glyphsLib.filters", "name": "eraseOpenCorners", "pre": True}
+        ]
         # FIXME: (jany) in the future, yield this UFO (for memory, lazy iter)
         self._designspace.addSource(source)
         self._sources[master.id] = source
--- a/Lib/glyphsLib/builder/font.py
+++ b/Lib/glyphsLib/builder/font.py
@@ -83,9 +83,9 @@ def to_ufo_font_attributes(self, family_name):
         self.to_ufo_names(ufo, master, family_name)
         self.to_ufo_family_user_data(ufo)
 
-        ufo.lib[UFO2FT_FILTERS_KEY] = [
+        ufo.lib.setdefault(UFO2FT_FILTERS_KEY, []).append(
             {"namespace": "glyphsLib.filters", "name": "eraseOpenCorners", "pre": True}
-        ]
+        )
 
         self.to_ufo_custom_params(ufo, font)
 
--- a/Lib/glyphsLib/builder/font.py
+++ b/Lib/glyphsLib/builder/font.py
@@ -16,7 +16,7 @@
 import logging
 
 from .common import to_ufo_time, from_ufo_time
-from .constants import GLYPHS_PREFIX
+from .constants import GLYPHS_PREFIX, UFO2FT_FILTERS_KEY
 
 logger = logging.getLogger(__name__)
 
@@ -82,15 +82,17 @@ def to_ufo_font_attributes(self, family_name):
 
         self.to_ufo_names(ufo, master, family_name)
         self.to_ufo_family_user_data(ufo)
+
+        ufo.lib[UFO2FT_FILTERS_KEY] = [
+            {"namespace": "glyphsLib.filters", "name": "eraseOpenCorners", "pre": True}
+        ]
+
         self.to_ufo_custom_params(ufo, font)
 
         self.to_ufo_master_attributes(source, master)
 
         ufo.lib[MASTER_ORDER_LIB_KEY] = index
 
-        ufo.lib["com.github.googlei18n.ufo2ft.filters"] = [
-            {"namespace": "glyphsLib.filters", "name": "eraseOpenCorners", "pre": True}
-        ]
         # FIXME: (jany) in the future, yield this UFO (for memory, lazy iter)
         self._designspace.addSource(source)
         self._sources[master.id] = source
--- a/Lib/glyphsLib/builder/font.py
+++ b/Lib/glyphsLib/builder/font.py
@@ -87,6 +87,12 @@ def to_ufo_font_attributes(self, family_name):
         self.to_ufo_master_attributes(source, master)
 
         ufo.lib[MASTER_ORDER_LIB_KEY] = index
+
+        ufo.lib["com.github.googlei18n.ufo2ft.filters"] = [{
+            "namespace": "glyphsLib.filters",
+            "name": "eraseOpenCorners",
+            "pre": True
+        }]
         # FIXME: (jany) in the future, yield this UFO (for memory, lazy iter)
         self._designspace.addSource(source)
         self._sources[master.id] = source
--- a/Lib/glyphsLib/filters/eraseOpenCorners.py
+++ b/Lib/glyphsLib/filters/eraseOpenCorners.py
@@ -1,10 +1,17 @@
 import logging
 
 from fontTools.pens.basePen import BasePen
-from fontTools.misc.bezierTools import segmentSegmentIntersections, _split_segment_at_t
+from fontTools.misc.bezierTools import (
+    segmentSegmentIntersections,
+    _split_segment_at_t,
+)
 from ufo2ft.filters import BaseFilter
 
 logger = logging.getLogger(__name__)
+
+def _pointIsLeftOfLine(line, aPoint):
+    a,b = line
+    return ((b[0] - a[0]) * (aPoint[1] - a[1]) - (b[1] - a[1]) * (aPoint[0] - a[0])) >= 0
 
 
 class EraseOpenCornersPen(BasePen):
@@ -38,11 +45,26 @@ class EraseOpenCornersPen(BasePen):
         ix = 0
         while ix < len(segs):
             next_ix = (ix + 1) % len(segs)
+
+            # Am I a line segment?
+            if not len(segs[ix]) == 2:
+                ix = ix + 1
+                continue
+
+            # Are the first point of the previous segment and the last point
+            # of the next segment both on the right side of the line?
+            # (see discussion at https://github.com/googlefonts/glyphsLib/pull/663)
+            pt1 = segs[ix - 1][0]
+            pt2 = segs[next_ix][-1]
             intersection = [
                 i
                 for i in segmentSegmentIntersections(segs[ix - 1], segs[next_ix])
                 if 0 <= i.t1 <= 1 and 0 <= i.t2 <= 1
             ]
+            if not intersection or _pointIsLeftOfLine(segs[ix], pt1) or _pointIsLeftOfLine(segs[ix], pt2):
+                ix = ix + 1
+                continue
+
             if intersection and (intersection[0].t1 > 0.5 and intersection[0].t2 < 0.5):
                 (segs[ix - 1], _) = _split_segment_at_t(
                     segs[ix - 1], intersection[0].t1
--- a/Lib/glyphsLib/filters/eraseOpenCorners.py
+++ b/Lib/glyphsLib/filters/eraseOpenCorners.py
@@ -1,0 +1,91 @@
+import logging
+
+from fontTools.pens.basePen import BasePen
+from fontTools.pens.recordingPen import RecordingPen
+from fontTools.misc.bezierTools import segmentSegmentIntersections, _split_segment_at_t
+from ufo2ft.filters import BaseFilter
+import beziers
+
+logger = logging.getLogger(__name__)
+
+
+class EraseOpenCornersPen(BasePen):
+    def __init__(self, outpen):
+        self.segments = []
+        self.is_closed = False
+        self.affected = False
+        self.outpen = outpen
+
+    def _moveTo(self, p1):
+        if self.segments or self.is_closed:
+            raise ValueError(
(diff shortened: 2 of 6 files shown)


simoncozens:
Of *course* the tests are going to fail, because I'm living in the future and GitHub Actions isn't.

simoncozens:
This is causing some problems with fontmake and certain variable fonts. Trying to track it down.

simoncozens:
The heuristic @anthrotype suggested in googlefonts/fontmake#288 was

> a filter pen that iterates over triplets (prev, curr, next) of segments in each contour, computes intersection between previous and next segment, and if intersection falls after the mid-point of the previous segment and before the mid-point of the next segment, then the filter pen should draw with the wrapped pen only the portions of previous and next segments that we want, thus excluding the current segment and surrounding sub-segments that make up the corner shape.

Unfortunately that doesn't deal with "fat" open corners like this one:

Screenshot 2021-02-26 at 13 57 08

where the intersections are at (t1=0.46, t2=0.13) and (t1=0.86,t2=t2=0.53).

I guess one approach would be to reorder things and run RemoveOverlap first, to get rid of this kind of open corner, leaving only the more simple ones. But that doesn't work in the general case as RemoveOverlap does unhelpful things to "outside" corners, putting them into a separate contour:

Screenshot 2021-02-26 at 14 07 46

So we either need to

1. Detect the difference between "inside" and "outside" open corners (inside ones being dealt with by RemoveOverlaps), or
2. Come up with a better heuristic for removing all open corners.

schriftgestalt:
The open corner processing is only done for outside corners. That means it only does sorting if the previous and next point is on the left of the line segment (as seen from that segment).

simoncozens:
> That means it only does sorting if the previous and next point is on the left of the line segment (as seen from that segment).

I don't understand this. 

Screenshot 2021-02-26 at 15 30 04

The "previous point" is the start point, right? (Where the 2 is) And the "next point" is the other selected point? They are on different sides of the line segment.

schriftgestalt:
I might have confused the direction. But both selected nodes are on the same side of the line going through the two points on the bottom (this segment goes from right to left, so the selected points are on the right of that line (when you look RTL))

simoncozens:
In the diagram above, both nodes are on the same side of the line, so it is an outside corner?

In the diagram below, both nodes are on the same side of the line, but it is an inside corner.

Screenshot 2021-02-26 at 23 56 04

schriftgestalt:
If both are on the right (not left) side when looking in the direction of the short line segment.

simoncozens:
Ah, ok, when the short line segment is aligned clockwise. Thanks, I’ll try working with that.

schriftgestalt:
I use this function: 
```
BOOL GSPointIsLeftOfLine(NSPoint a, NSPoint b, NSPoint aPoint) {
	return ((b.x - a.x) * (aPoint.y - a.y) - (b.y - a.y) * (aPoint.x - a.x)) >= 0;
}
```
and if both points are NOT left of the line, and the both t vales are below 0.5.

simoncozens:
(Not my test failure - pip-compile is being too strict and pip is being stupid about resolving dependencies.)

simoncozens:
> (Not my test failure - pip-compile is being too strict and pip is being stupid about resolving dependencies.)

Specifically, the current *release* of ufolib2 hard-requires fontTools 4.19.0 and we need 4.20.0. We need a new release of ufolib2 to unblock. (Or just stop using pip-compile which is creating too-strict dependency chains.)

anthrotype:
> ufolib2 hard-requires fontTools 4.19.0

You sure? Where did you see that. In setup.cfg, ufoLib2 only seems to require fonttools >= 4.0.0: https://github.com/fonttools/ufoLib2/blob/1201353d31ce463b13ac12cd12253c4d5a18bded/setup.cfg#L29

you should be able to re-pip-compile the requirements with `pip-compile -U setup.py` (@madig am I right)

> stop using pip-compile which is creating too-strict dependency chains

that was to make the test runs more predictable. I'm ok if you want to ditch that and keep all deps open-ended and install latest. It means you may find yourself having to deal with sudden unrelated breakages in the middle of your PR

simoncozens:
> you should be able to re-pip-compile the requirements with pip-compile -U setup.py (@madig am I right)

I already did that in fab5d27.

>> ufolib2 hard-requires fontTools 4.19.0
> You sure? Where did you see that.

Hmm, no, I'm not sure. *Something* is pulling in 4.19.0 as you can see in the [test failure](https://github.com/googlefonts/glyphsLib/pull/663/checks?check_run_id=2056838892). But pip isn't very good at telling me what it is.

simoncozens:
Ack, it's the `requirements-dev`.

simoncozens:
OK, that failing test now *is* mine, but I don't know how to fix it. We are adding a ufo2ft filter on export to UFO, but I am not sure how that should be round-tripped.

anthrotype:
> how that should be round-tripped

i think it shouldn't be roundtripped

anthrotype:
is the default (and only) behavior in Glyphs.app to erase open corners? Then when going from ufo -> glyphs, if you find that filter, you pop it out and skip importing it. Same way as you unconditionally always add one when going from glyphs -> ufo

simoncozens:
Yes, it is. So this is sort of what we do in the latest commit; we add it first, but if filters are found in the UFO, they replace it.

madig:
Please update the requirements files with the newest pip-tools to avoid needless diffs.

madig:
Thanks, I'll try to get to reviewing this this week... out of curiosity, are there use cases for this pen outside a Glyphs context? I wonder if FontLab implements this differently.

madig (approved):
Seems fine to me.

Is it maybe worth making it work on multiple paths at once (i.e. be able to process all contours of a glyph and output trimmed ones without having to re-instantiate the pen between each)?

Lib/glyphsLib/builder/font.py
```diff
@@ -82,11 +82,17 @@ def to_ufo_font_attributes(self, family_name):
 
         self.to_ufo_names(ufo, master, family_name)
         self.to_ufo_family_user_data(ufo)
+
+        ufo.lib[UFO2FT_FILTERS_KEY] = [
```
madig, line 86: What if the UFO has preexisting filters though?

anthrotype: some Glyphs-specific filters (e.g. Transformations) are stored in the same lib key when exporting custom parameters. Depending on the order of these operations (i'm not sure exactly which one happens first) it may be that that list is non-empty. Better to avoid overwriting, try do a `ufo.lib.setdefault(UFO2FT_FILTERS_KEY, []).append(...)`

anthrotype:
> I wonder if FontLab implements this differently.

oh, does Fontlab also have similarly "open corners"?

simoncozens:
It doesn't look like fontLab's "loop corners" do the same thing on the *outside* as Glyphs does.
Screenshot 2021-03-18 at 15 08 34
Screenshot 2021-03-18 at 15 09 15

Lib/glyphsLib/filters/eraseOpenCorners.py
```diff
@@ -0,0 +1,118 @@
...
+
+
+def _pointIsLeftOfLine(line, aPoint):
+    a, b = line
+    return (
+        (b[0] - a[0]) * (aPoint[1] - a[1]) - (b[1] - a[1]) * (aPoint[0] - a[0])
+    ) >= 0
+
+
+class EraseOpenCornersPen(BasePen):
```
anthrotype, line 20: yours is not really a "drawing pen" like the ones BasePen was designed to work as base class, you're not using any of its built-in facilities. I think what you actually want to sub-class is the FilterPen, or even better the ContourFilterPen, which takes another pen as parameter and accummulates the drawing commands, then at the end of a contour it "flushes" them to be processed by its filter.

https://github.com/fonttools/fonttools/blob/2687c12c2e9dc372e3f35ee5e7318c20c1519c26/Lib/fontTools/pens/filterPen.py#L80

anthrotype: oh I see, you're using `BasePen._getCurrentPoint`. ok

Lib/glyphsLib/filters/eraseOpenCorners.py
```diff
@@ -0,0 +1,118 @@
...
+
+        for seg in segs:
+            if len(seg) == 2:
+                self.outpen.lineTo(*seg[1:])
+            elif len(seg) == 3:
+                self.outpen.qCurveTo(*seg[1:])
+            elif len(seg) == 4:
+                self.outpen.curveTo(*seg[1:])
+
+        if self.is_closed:
```
anthrotype, line 98: I think `self.is_closed` should be reset at the end, before a new contour is drawn with the same pen, otherwise it stays True

Lib/glyphsLib/filters/eraseOpenCorners.py
```diff
@@ -0,0 +1,118 @@
...
+    def filter(self, glyph):
+        if not len(glyph):
+            return False
+
+        affected = False
+        contours = list(glyph)
+        glyph.clearContours()
+        outpen = glyph.getPen()
+        for contour in contours:
+            p = EraseOpenCornersPen(outpen)
```
anthrotype, line 114: pens instances should be reusable, not disposed for each individual contour. Ideally this is instantiated once outside the loop (but should make sure the state is reset at the end of each contour)

anthrotype:
I understand this behaviour is the default in Glyphs.app, however I'm a bit concerned that this filter will then be enabled for all .glyphs sources that are built via fontmake, especially given that the segment-segment intersection code in fonttools bezierTools is quite recent and does not have a complete test suite.
How about we have this behind a (yet-another-I-know) flag? One can argue that the default fontmake behavior so far has been to _not_ erase glyphs-style open corners..

Lib/glyphsLib/filters/eraseOpenCorners.py
simoncozens: I looked at ContourFilterPen - that was what you told me to use initially - but it doesn't do what we want here. We want to work on segments, not operations, so I use the pen class to accumulate the segment information and then work on it at the end.

Please don't get me started on "drawing" pens versus non-drawing "pens". It never ends well.

simoncozens:
> One can argue that the default fontmake behavior so far has been to not erase glyphs-style open corners.

Yes, one can. But one can also argue that this is a mistake and counterintuitive. The expected behaviour is that fonts produced by glyphsLib and fonts produced by Glyphs look the same. Currently they don't, which is why we have designers opening issues saying that fontmake is making their fonts look horrible, and why we have to tell people not to use open corners in fonts they submit.

If the problem is that the fontTools code doesn't have a good enough test suite, let's fix *that* problem, but that's not *this* problem. We shouldn't need to enable an optional flag to get expected behaviour. The program should do what's expected by default.

moyogo:
> It doesn't look like fontLab's "loop corners" do the same thing on the outside as Glyphs does.

FontLab does this, it's called "looped corners", but it's hidden behind a setting: Font Info > Other Values > Contour Properties > Unfill looped corners.
Checkout https://www.youtube.com/watch?v=iQ3FkgDLVOA
The latest version also has "Suggest distance X" and "Suggest distance Y" values.
image

anthrotype:
@moyogo cool, so it works the same way but more user-configurable

moyogo:
Should the erase open corner filter be merged in ufo2ft instead?

simoncozens:
If fontlabLib existed, then yes, absolutely it makes sense to put this at the next level in the pipeline. But it doesn't. (Well, it does, it's called babelfont. ;-) So for practical purposes it is only Glyphs which uses open corners.

moyogo:
I guess it doesn't matter where it is if I want to use it in a UFO workflow, since fontmake requires glyphsLib anyway.

simoncozens:
Discussed here: https://github.com/googlefonts/fontmake/issues/288#issuecomment-781456532

moyogo:
> Discussed here: googlefonts/fontmake#288 (comment)
>> So I think glyphsLib should be responsible for adding the filter to the UFO libs. It should probably be defined inside glyphsLib itself, being a glyphs-specific feature.

So is the Propagate Anchors Filter or the Transformations Filter, yet they are in ufo2ft because it's a useful filter to have in a variety of workflows. There are open corners in Noto Sans Georgian UFO sources for example. FontLab can still export UFOs to be used in fontmake.

I don't mind having filters in glyphsLib, I just don't understand what's the logic for that exception.

anthrotype:
> There are open corners in Noto Sans Georgian UFO sources for example

yeah, but that's because it was designed inside Glyphs.app then converted to UFO. I don't think there's any native UFO editor that allows one to draw/view these so-called "open corners". It's a font-editor-specific feature, a Glyphs.app-ism, hence it lives in glyphsLib. If it were to become something that is more widely used in UFO-based workflows then we may reconsider moving this to ufo2ft proper.

madig:
@schriftgestalt how does Glyphs.app handle Sofia Sans's kgreenlandic in the DSFix layer?

image

Moved aside to show the contour:

image

simoncozens:
> and if both points are NOT left of the line, and the *both t vales are below 0.5*.

@schriftgestalt, I'm confused by this. In this case:

Screenshot 2021-09-23 at 10 52 24

the t value of the intersection on both line segments is < 0.5, so that's detected as an open corner. OK, that makes sense.

But in this case:

Screenshot 2021-09-23 at 10 53 25

the t value of the intersection on the selected line segment is way > 0.5, but it's still detected as an open corner!

schriftgestalt:
this is the actual code:
```
if (((t1 < 0.5 && t2 < 0.5) || t1 < 0.3 || t2 < 0.3) && t1 > 0.0001 && t2 > 0.0001) {
```
So if one of the edges id smaller than 0.3, it will still do it.

simoncozens:
Aha! Thank you!

PR #664, opened 2021-02-23
UFO default layer with custom name

moyogo:
The layer name [`public.default`](https://unifiedfontobject.org/versions/ufo3/layercontents.plist/#publicdefault) *indicates that the default layer has no user defined name*.
glyphsLib should store the user defined default layer name and round-trip it to UFOs.

The layer lib is now stored in the master userData instead of the font userData.
When several UFOs each have a layer with the same name, their lib may be different and cannot be stored in the font userData under the same key.

This now uses `master.userData["com.schriftgestaltung.Glyphs.layerLib.<layer.name>"]["com.schriftgestaltung.Glyphs.layerName"]`, one per layer, instead of `font.userData["com.schriftgestaltung.Glyphs.layerLib.<layer.name>"]["com.schriftgestaltung.Glyphs.layerName"]`, one for all layers with the same name across UFOs.

### DIFF
--- a/Lib/glyphsLib/builder/layers.py
+++ b/Lib/glyphsLib/builder/layers.py
@@ -71,7 +71,7 @@ def _layer_order_in_glyph(self, layer):
 
 
 def to_glyphs_layer(self, ufo_layer, glyph, master):
-    if ufo_layer.name == self._sources[master.id].font.layers.defaultLayer.name:
+    if ufo_layer is self._sources[master.id].font.layers.defaultLayer:
         layer = _get_or_make_foreground(self, glyph, master)
     elif ufo_layer.name == "public.background":
         master_layer = _get_or_make_foreground(self, glyph, master)
--- a/Lib/glyphsLib/builder/layers.py
+++ b/Lib/glyphsLib/builder/layers.py
@@ -71,7 +71,7 @@ def _layer_order_in_glyph(self, layer):
 
 
 def to_glyphs_layer(self, ufo_layer, glyph, master):
-    if ufo_layer.name == "public.default":  # TODO: (jany) constant
+    if ufo_layer.name == self._sources[master.id].font.layers.defaultLayer.name:
         layer = _get_or_make_foreground(self, glyph, master)
     elif ufo_layer.name == "public.background":
         master_layer = _get_or_make_foreground(self, glyph, master)
--- a/Lib/glyphsLib/builder/user_data.py
+++ b/Lib/glyphsLib/builder/user_data.py
@@ -77,10 +77,21 @@ def to_ufo_glyph_user_data(self, ufo, glyph):
         ufo.lib[key] = dict(glyph.userData)
 
 
-def to_ufo_layer_lib(self, ufo_layer):
+def to_ufo_layer_lib(self, master, ufo, ufo_layer):
     key = LAYER_LIB_KEY + "." + ufo_layer.name
+    # glyphsLib v5.3.2 and previous versions stored the layer lib in
+    # the GSFont useData under a key named after the layer.
+    # When different original UFOs each had a layer with the same layer name,
+    # only the layer lib of the last one was stored and was exported to UFOs
     if key in self.font.userData.keys():
-        ufo_layer.lib = self.font.userData[key]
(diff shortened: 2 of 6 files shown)


anthrotype (approved):

Lib/glyphsLib/builder/user_data.py
```diff
@@ -77,10 +79,29 @@ def to_ufo_glyph_user_data(self, ufo, glyph):
...
+    # the GSFont useData under a key named after the layer.
+    # When different original UFOs each had a layer with the same layer name,
+    # only the layer lib of the last one was stored and was exported to UFOs
     if key in self.font.userData.keys():
-        ufo_layer.lib = self.font.userData[key]
+        self.logger.warning(
+            f"Layer '{ufo_layer.name}' lib is stored once in GSFont userData "
+            "instead of in each GSFontMaster but may have been different in "
+            "previous UFOs."
+        )
```
belluzj, line 93: Does this warrant a warning?

belluzj:
I was going to comment that along with updating where the UFO layer lib is stored on the Glyphs side, we should also update my diagram that explains that: https://github.com/googlefonts/glyphsLib/blob/main/tests/builder/lib_and_user_data.png

But actually my diagram seems to already be describing the new situation, so I must have implemented my own diagram wrong when I was working on this? 🙄

Lib/glyphsLib/builder/user_data.py
```diff
@@ -146,15 +167,24 @@ def to_glyphs_glyph_user_data(self, ufo, glyph):
...
+def to_glyphs_layer_lib(self, ufo_layer, master):
     user_data = {}
     for key, value in ufo_layer.lib.items():
         if _user_data_has_no_special_meaning(key):
             user_data[key] = value
 
+    # the default layer may have a custom name
+    layer_name = ufo_layer.name
+    if (
+        layer_name == self._sources[master.id].font.layers.defaultLayer.name
```
madig, line 179:
```suggestion
        ufo_layer is self._sources[master.id].font.layers.defaultLayer
```

Maybe?

anthrotype: yeah, even better

Lib/glyphsLib/builder/user_data.py
anthrotype: the less noise the better

madig:
Does this in any way interact with Glyphs.app layer/master names? I remember newer G2 versions letting you set the master name. Does this transport into UFo or is it a separate thing?

moyogo:
> Does this in any way interact with Glyphs.app layer/master names? I remember newer G2 versions letting you set the master name. Does this transport into UFo or is it a separate thing?

I'm not able to rename the master default layer (through the UI or the API), it just uses the name of the master.

Lib/glyphsLib/builder/user_data.py
moyogo: OK. Removing the warning

Lib/glyphsLib/builder/user_data.py
moyogo: Good point!

schriftgestalt:
Master layers are not supposed to have its own name. If a files happens to have one anyway, you can ignore it. Keeping the name around is less important in Glyphs 3 files as the important information moved to `layer.attributs`.

Lib/glyphsLib/builder/layers.py
```diff
@@ -71,7 +71,7 @@ def _layer_order_in_glyph(self, layer):
 
 
 def to_glyphs_layer(self, ufo_layer, glyph, master):
-    if ufo_layer.name == "public.default":  # TODO: (jany) constant
+    if ufo_layer.name == self._sources[master.id].font.layers.defaultLayer.name:
```
madig, line 74:
```suggestion
    if ufo_layer is self._sources[master.id].font.layers.defaultLayer:
```

Maybe?

moyogo: Yes, like the other one.

madig (approved):

PR #666, opened 2021-02-27
Populate UFO lib "public.openTypeCategories"

moyogo:
This uses the same heuristic as `glyphsLib.builders.features._build_gdef()` to write the ufo.lib["public.openTypeCategories"] after to_ufo_propagate_font_anchors() may have been called.

### DIFF
--- a/Lib/glyphsLib/builder/glyph.py
+++ b/Lib/glyphsLib/builder/glyph.py
@@ -151,6 +151,43 @@ def to_ufo_glyph(self, ufo_glyph, layer, glyph):  # noqa: C901
         self.to_ufo_glyph_height_and_vertical_origin(ufo_glyph, layer)
 
 
+def to_ufo_openTypeCategories(self, ufo):
+    # See:
+    # * glyphsLib.builder.features._build_gdef
+    # * https://github.com/googlefonts/glyphsLib/issues/85
+    # * https://github.com/googlefonts/glyphsLib/pull/100#issuecomment-275430289
+    openTypeCategories = dict()
+    for ufo_glyph in ufo:
+        glyphinfo = glyphsLib.glyphdata.get_glyph(ufo_glyph.name)
+        category = ufo_glyph.lib.get(GLYPHLIB_PREFIX + "category") or glyphinfo.category
+        subCategory = (
+            ufo_glyph.lib.get(GLYPHLIB_PREFIX + "subCategory") or glyphinfo.subCategory
+        )
+        has_base_anchor = any(
+            not a.name.startswith("_") for a in ufo_glyph.anchors if a.name
+        )
+
+        if subCategory == "Ligature" and has_base_anchor:
+            openTypeCategory = "ligature"
+        elif category == "Mark" and (
+            subCategory in {"Nonspacing", "Spacing Combining"}
+        ):
+            openTypeCategory = "mark"
+        elif has_base_anchor:
+            openTypeCategory = "base"
+        else:
+            openTypeCategory = None
+
+        if openTypeCategory:
+            openTypeCategories[ufo_glyph.name] = openTypeCategory
+
+    if openTypeCategories:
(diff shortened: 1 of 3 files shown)


anthrotype:
so you want glyphsLib to both write out GDEF in features.fea _and_ populate the public.openTypeCategories? Why isn't one of the two sufficient?

If we needed both, at least we should try not to duplicate logic from _build_gdef but reuse as much as possible

moyogo:
> so you want glyphsLib to both write out GDEF in features.fea and populate the public.openTypeCategories? Why isn't one of the two sufficient?

What are you suggesting? This just adds public.openTypeCategories. The user controls whether GDEF is written to the features.fea.
Should we let the user control whether public.openTypeCategories are written?

anthrotype:
if we output both of them, the features.fea would always take precedence and the lib stuff would do nothing (at least that's I remember we agreed on in the ufo2ft PR that adds support for this).
We need to output only one of the two sets of data. Once ufo2ft becomes aware of the new lib key and can generate GDEF (including ligature carets), then we can switch to solely output the latter. The same option --generate_GDEF could be repurposed for controlling whether to write the lib key or not.

moyogo:
So we disable writing public.openTypeCategories when generate_GDEF is true.
```diff
diff --git a/Lib/glyphsLib/builder/builders.py b/Lib/glyphsLib/builder/builders.py
index 89594c5..888cad7 100644
--- a/Lib/glyphsLib/builder/builders.py
+++ b/Lib/glyphsLib/builder/builders.py
@@ -262,7 +262,8 @@ class UFOBuilder(_LoggerMixin):
             master = self.font.masters[master_id]
             if self.propagate_anchors:
                 self.to_ufo_propagate_font_anchors(ufo)
-            self.to_ufo_openTypeCategories(ufo)
+            if not self.generate_GDEF:
+                self.to_ufo_openTypeCategories(ufo)
             for layer in list(ufo.layers):
                 self.to_ufo_layer_lib(master, ufo, layer)
```

khaledhosny:
> This always overwrites categories, no matter what the UFO had before conversion to .glyphs,

Isn’t this code used when converting .glyphs to .ufo, not the other way around?

Lib/glyphsLib/builder/glyph.py
```diff
@@ -151,6 +151,43 @@ def to_ufo_glyph(self, ufo_glyph, layer, glyph):  # noqa: C901
         self.to_ufo_glyph_height_and_vertical_origin(ufo_glyph, layer)
 
 
+def to_ufo_openTypeCategories(self, ufo):
+    # See:
+    # * glyphsLib.builder.features._build_gdef
```
anthrotype, line 156: I think we should factor out the logic to build the GDEF categories from GlyphData from the existing _build_gdef to a new method that both it and this can use, we don't want to duplicate same logic

madig:
> > This always overwrites categories, no matter what the UFO had before conversion to .glyphs,
> 
> Isn’t this code used when converting .glyphs to .ufo, not the other way around?

I meant to say that I assume the code in this PR will always overwrite whatever was in the UFO when going UFO -> .glyphs -> UFO.

anthrotype:
> Isn’t this code used when converting .glyphs to .ufo, not the other way around?

yeah, this PR does not seem to be handling the other-way-around (ufo -> glyphs). But note that, alongside the UFO openTypeCategories, glyphsLib is still outputting similar (but not quite identical?) category/subCategory info in private glyph.lib keys, to ensure custom glyph categories will round-trip.

anthrotype:
> will always overwrite whatever was in the UFO when going UFO -> .glyphs -> UFO

indeed, it will discard any pre-existing public.openTypeCategories because that's not recorded in the generated .glyphs file. Could we store lossless-ly in the glyph category property?

madig:
https://github.com/googlefonts/glyphsLib/pull/792

Issue #588, opened 2020-02-07
error when exporting to UFO and duplicate layers contain background


alexmyczko:
```
# glyphs2ufo Inter.glyphs 
Inter Thin: Glyph asterisk, layer Dec 27 18, 16:33: Duplicate glyph layer name
Traceback (most recent call last):
  File "/usr/bin/glyphs2ufo", line 11, in <module>
    load_entry_point('glyphsLib==5.1.4', 'console_scripts', 'glyphs2ufo')()
  File "/usr/lib/python3/dist-packages/glyphsLib/cli.py", line 228, in _glyphs2ufo_entry_point
    return main(args)
  File "/usr/lib/python3/dist-packages/glyphsLib/cli.py", line 189, in main
    return options.func(options)
  File "/usr/lib/python3/dist-packages/glyphsLib/cli.py", line 220, in glyphs2ufo
    ufo_module=__import__(options.ufo_module),
  File "/usr/lib/python3/dist-packages/glyphsLib/__init__.py", line 123, in build_masters
    ufo_module=ufo_module,
  File "/usr/lib/python3/dist-packages/glyphsLib/builder/__init__.py", line 111, in to_designspace
    return builder.designspace
  File "/usr/lib/python3/dist-packages/glyphsLib/builder/builders.py", line 260, in designspace
    list(self.masters)  # Make sure that the UFOs are built
  File "/usr/lib/python3/dist-packages/glyphsLib/builder/builders.py", line 223, in masters
    self.to_ufo_glyph(ufo_glyph, layer, layer.parent)
  File "/usr/lib/python3/dist-packages/glyphsLib/builder/glyph.py", line 127, in to_ufo_glyph
    self.to_ufo_glyph_background(ufo_glyph, layer)
  File "/usr/lib/python3/dist-packages/glyphsLib/builder/glyph.py", line 265, in to_ufo_glyph_background
    new_glyph = ufo_layer.newGlyph(glyph.name)
  File "/usr/lib/python3/dist-packages/ufoLib2/objects/layer.py", line 120, in newGlyph
    raise KeyError("glyph %r already exists" % name)
KeyError: "glyph 'asterisk' already exists"
```

Find Inter at https://github.com/rsms/inter

https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=950099

madig:
Smells like a duplicate layer name. I though this was fixed 🤔

alexmyczko:
trying master

[...]

Issue #573, opened 2020-01-07
UFO->Glyphs should strip #<digit> suffix from 'Color' layer names 


anthrotype:
in #570, we fixed the error when exporting glyphs->UFO and glyphs contain duplicate layers with the same name by appending a `#<digit>` suffix to make the layer name unique (UFO layers must have unique names).

However, @schriftgestalt noted that in Glyphs.app there are some special glyph layers that are used for color fonts ("Color XX" where X are numbers indicating the index of a color palette) and which can occur multiple times with the same layer name string within the same glyph master:
https://github.com/googlefonts/glyphsLib/issues/566#issuecomment-561643874

When going back from UFO -> glyphs, we should check if the layer name starts with "Color XX" and strip any such suffixes.

khaledhosny:
This is already taken care of when handling color layers.

PR #667, opened 2021-02-27
Rename backgrounds of layers with the same name

moyogo:
Same as #570 but for background of layers with the same name.


moyogo:
I've updated the PR to store the layerId and the original layer name when there are multiple layers with the same name, and to keep the layer order as well when any layer has a name like "Color 1", fixing  both #573 and #588.

Lib/glyphsLib/builder/builders.py
```diff
@@ -517,6 +527,63 @@ def _copy_bracket_layers_to_ufo_glyphs(self, bracket_layer_map):
...
+            ufo_font = self._sources[master.id].font
+
+            # Collect layers in the order we first see them
+            layers_by_name = dict()
+            for glyph in self.font.glyphs:
+                for layer in glyph.layers:
+                    if master.id == layer.associatedMasterId:
+                        if layer.name not in layers_by_name:
+                            layers_by_name[layer.name] = []
+                        layers_by_name[layer.name].append(layer)
```
anthrotype, line 541: you could use `dict.setdefault` or a `defaultdict`

Lib/glyphsLib/builder/builders.py
```diff
@@ -517,6 +527,63 @@ def _copy_bracket_layers_to_ufo_glyphs(self, bracket_layer_map):
...
+                    if layer_name is not None:
+                        ufo_layer = ufo_font.layers[layer_name]
+                    else:
+                        n = 1
+                        layer_name = layer.name
+                        while layer_name in ufo_font.layers:
+                            layer_name = f"{layer.name} #{n!r}"
+                            n += 1
+                        ufo_layer = ufo_font.newLayer(layer_name)
+                        if layer.name != "Color 1":
```
anthrotype, line 575: what's this "Color 1"? A new magic value? Why only `1`?

Lib/glyphsLib/builder/layers.py
```diff
@@ -12,53 +12,83 @@
...
     else:
-        ufo_layer = ufo_font.layers[layer.name]
-    if self.minimize_glyphs_diffs:
+        # Find foreground layer and use the same name as a base for the
+        # background layer name.
+        layer_name = next(
+            (
+                l.name
+                for l in ufo_font.layers
+                if l.lib.get(LAYER_ID_KEY) == layer.layerId
```
anthrotype, line 41: maybe you could store a map from layerId to layer name in the builder instance instead of finding them like this each time

arrowtype:
I’m not quite sure whether this is due to this specific branch, or due to other code, but in a font I’ve use `glyphs2ufo` on, I’ve wound up with a couple thousand separate layer directories in the UFOs. As a result, the UFOs seem to be basically impossible to open with RoboFont or fontParts fontShell...

Is there a way to prevent the additional layers from being created? I don’t really need them. 

If this would be better as a separate issue, please let me know, and I can move this comment.

image

PR #668, opened 2021-03-08
Allow using custom GlyphData.xml with glyphs2ufo or to_ufos()

moyogo:


khaledhosny:
The regression tests failure is because we were previously adding empty subCategory to glyph.lib.

Issue #673, opened 2021-03-25
Completeness is wasteful


simoncozens:
In the common case of using glyphsLib to convert to UFO as part of the fontmake pipeline, all of this code is unnecessary:

```
    self.to_ufo_background_image(ufo_glyph, layer)
    self.to_ufo_guidelines(ufo_glyph, layer)
    self.to_ufo_glyph_background(ufo_glyph, layer)
    self.to_ufo_annotations(ufo_glyph, layer)
```

Converting any non-master, non-brace layers is also wasteful.

khaledhosny:
Not only wasteful, Glyphs files might have some garbage there (e.g. layers with duplicate names) that will not be used for anything but will break the build anyway.

moyogo:
Color layers can have the same name if they have the same color. https://github.com/googlefonts/glyphsLib/issues/573

simoncozens:
Incidentally the performance impact of doing less is pretty minimal in the grand scheme of things: https://daltonmaag.github.io/pipeline-perf-tracker/results/

madig:
Profile of fontmake when compiling Noto Sans to a VF from a .glyphs file:

![image](https://user-images.githubusercontent.com/380829/113184639-57e9a880-924d-11eb-8cca-e8b572dafb68.png)

A lot of time is wasted on 1. writing intermediate UFOs to disk because MutatorMath needs that 2. parsing the file.

simoncozens:
Use the new parser.

PR #674, opened 2021-03-28
Do less

simoncozens:
This skips over steps which are not necessary for font conversion when generating UFO files. See #673

### DIFF
--- a/Lib/glyphsLib/builder/__init__.py
+++ b/Lib/glyphsLib/builder/__init__.py
@@ -31,6 +31,7 @@ def to_ufos(
     generate_GDEF=True,
     store_editor_state=True,
     write_skipexportglyphs=False,
+    minimal=True
 ):
     """Take a GSFont object and convert it into one UFO per master.
 
@@ -44,6 +45,10 @@ def to_ufos(
 
     If generate_GDEF is True, write a `table GDEF {...}` statement in the
     UFO's features.fea, containing GlyphClassDef and LigatureCaretByPos.
+
+    If minimal is True, it is assumed that the UFOs will only be used in
+    font production, and unnecessary steps (e.g. converting background layers)
+    will be skipped.
     """
     builder = UFOBuilder(
         font,
@@ -54,6 +59,7 @@ def to_ufos(
         generate_GDEF=generate_GDEF,
         store_editor_state=store_editor_state,
         write_skipexportglyphs=write_skipexportglyphs,
+        minimal=minimal
     )
 
     result = list(builder.masters)
@@ -73,6 +79,7 @@ def to_designspace(
     generate_GDEF=True,
     store_editor_state=True,
     write_skipexportglyphs=False,
+    minimal=True,
 ):
     """Take a GSFont object and convert it into a Designspace Document + UFOS.
     The UFOs are available as the attribute `font` of each SourceDescriptor of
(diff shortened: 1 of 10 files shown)


Lib/glyphsLib/builder/__init__.py
```diff
@@ -44,6 +45,10 @@ def to_ufos(
 
     If generate_GDEF is True, write a `table GDEF {...}` statement in the
     UFO's features.fea, containing GlyphClassDef and LigatureCaretByPos.
+
+    If minimal is True, it is assumed that the UFOs will only be used in
+    font production, and unnecessary steps (e.g. converting background layers)
+    will be skipped.
```
anthrotype, line 51: maybe list exactly what is dropped

anthrotype (approved):
LGTM

can you fix the lint failure before merge?

Also can you tell us what the speed savings are for an average font?

simoncozens:
```python
import glyphsLib
from timeit import timeit

def convert_minimal():
    glyphsLib.load_to_ufos("SofiaSans-VF-New.glyphs", minimal=True)

def convert_maximal():
    glyphsLib.load_to_ufos("SofiaSans-VF-New.glyphs", minimal=False)

minimal = timeit(convert_minimal, number=15)
maximal = timeit(convert_maximal, number=15)
print("minimal=True: %f" % minimal)
print("minimal=False: %f" % maximal)
```

```
minimal=True: 68.512878
minimal=False: 44.625200
```

Yes, that's right. Avoiding running unnecessary code takes 54% longer. I don't understand it either.

simoncozens:
Looks like it's caching/data locality issues with the parser. Here's another test:

```python
import glyphsLib
from timeit import timeit

parsed = glyphsLib.GSFont("SofiaSans-VF-New.glyphs")

def convert_minimal():
    glyphsLib.to_ufos(parsed, minimal=True)

def convert_maximal():
    glyphsLib.to_ufos(parsed, minimal=False)

maximal = timeit(convert_maximal, number=10)
minimal = timeit(convert_minimal, number=10)
print("minimal=True: %f" % minimal)
print("minimal=False: %f" % maximal)

# Try other order, just to make sure.
minimal = timeit(convert_minimal, number=10)
maximal = timeit(convert_maximal, number=10)
print("minimal=True: %f" % minimal)
print("minimal=False: %f" % maximal)
```

```
minimal=True: 21.209306
minimal=False: 22.644560
minimal=True: 20.398469
minimal=False: 21.063948
```

So what - 5% faster generating UFOs?

simoncozens:
I'd like to get this merged, but it's kind of a breaking change.

For the majority use case (building fonts), it makes things faster and it also avoids certain stupid bugs: I have a build failing because there's a background glyph with a non-existent component - which shouldn't make any difference to the build process, but means that complete UFOs can't be generated, which stops everything dead.

But for the glyphs<->UFO<->glyphs workflow, changing the default to be "minimal=True" might muck things up. So I would like to check with @belluzj and @madig before merging.

(An obvious alternative is that we set default to "minimal=False" and turn on "minimal=True" in build systems. Yeah, that's clearly better...)

anthrotype:
> turn on "minimal=True" in build systems

I believe that's preferable

simoncozens:
@anthrotype Happy with this now?

anthrotype:
LGTM thanks

PR #679, opened 2021-04-05
WIP: Fix bracket glyph components self-referencing after roundtrip.

tomarcher101:
Currently there is an asymmetry in the implementation of [alternating glyph shapes](https://staging.glyphsapp.com/learn/alternating-glyph-shapes) between UFO and glyphs.app that results in roundtripping resulting in glyphs that reference themselves as a component. This self-referencing can then lead to recursion errors.

UFO handles alternating glyphs shapes by including the alternative glyph as a standalone glyph in its own right, found in the glif table. GSUB rules are then implemented to make the base glyph swap out for the alternate glyph when the correct conditions are met.

Glyphs handles alternating glyph shapes using the [bracket trick](https://staging.glyphsapp.com/learn/alternating-glyph-shapes), which instead stores the alternate glyph, not as a different glyph, but as a "bracket" layer of the base glyph.

As a result, if you have an alternate glyph that has its base glyph as a referenced component in a UFO, running `ufo2glyphs` combines the base glyph and alternate glyph into a single glyph and you get a self referencing glyph that Glyphs.app cannot display. On the return trip, `glyphs2ufo` does not unpack and separate these self referencing glyphs into their base and alternate glyph components correctly; instead creating glifs for the alternate glyphs that reference themselves as a component.

This fix is to fix the 2nd problem: allowing `glyphs2ufo` to correctly unpack the base and alternate glyphs correctly so the resulting UFO after a round trip isn't broken.


schriftgestalt:
Glyphs 3 can handle both, bracket layers and distinct glyphs+conditions in feature code.

madig:
> Glyphs 3 can handle both, bracket layers and distinct glyphs+conditions in feature code.

@schriftgestalt is there a tutorial for the latter? I only see the old tutorials that explain things for G2.

schriftgestalt:
the new tutorial is almost finished. The syntax is explained here https://glyphsapp.com/news/glyphs-3-make-things-you-love (search for "#ifdef VARIABLE")

PR #683, opened 2021-04-21
[builder] Respect "Propagate Anchors" custom parameter

khaledhosny:
The custom parameter was introduced in Glyphs 3, but I see no harm respecting it for Glyphs 2 as well.
https://forum.glyphsapp.com/t/stoping-anchor-propagation/18424


moyogo:
Thanks @khaledhosny. Could you add a test?

anthrotype (changes requested):
LGTM

anthrotype (approved):
LGTM

Issue #446, opened 2018-10-24
parse_hint_target arguments


schriftgestalt:
https://github.com/googlei18n/glyphsLib/blob/cfef8fb7d189c3e52ba89b6c6ddaa99a1e85ff8d/Lib/glyphsLib/classes.py#L243

`parse_hint_target()` should accept indexes with 2, 3 and 4 values. {1, 2}, {1, 2, 3}, {1, 2, 3, 4}. That happens with TrueType hints that are attached to interesting segments.

schriftgestalt:
I implemented something that fixes this. I don’t like regex enough to try to write one for that ;)
https://github.com/googlei18n/glyphsLib/pull/447

PR #690, opened 2021-06-28
Rework parser and writer (G3 prep)

simoncozens:
I'm going to do the Glyphs 3 branch again, now I know what I'm doing.

This PR is preparatory work for Glyphs 3 support, but there is nothing G3 specific about it at this stage. It consists of three commits, each of which passes tests independently when applied in sequence.

1. The first commit replaces the existing parser with one which uses openstep-plist. It has a facility which allows classes to deserialise their properties.
2. The second adds a facility to the writer allowing classes to serialize themselves, and then uses this to write out the data "by hand" instead of using magic properties (_classesForName/shouldWriteValueForKey/_keyOrder). This seems like a backwards step, but it really helps when adding Glyphs 3 support.
3. The third removes those properties which were there to support the automagic serialization.


simoncozens:
Well, the *first* commit is a bad rebase, but the other three are as described...

Lib/glyphsLib/classes.py
```diff
@@ -1338,25 +1324,15 @@ def plistValue(self):
...
 
-    _classesForName = {
-        "alignment": str,
-        "angle": parse_float_or_int,
-        "locked": bool,
-        "position": Point,
-        "showMeasurement": bool,
-        "filter": str,
-        "name": str,
-    }
```
khaledhosny, line 1359: Happy to see this stuff gone!

jenskutilek:
I don't think this is G3-specific, but TrueType hints can have 3 values in origin and target. That happens when the hinted point is one that will only be inserted when converting to TrueType, like an inflection point.

Here's an example file that fails when trying to parse the hints: [S.glyphs.zip](https://github.com/googlefonts/glyphsLib/files/6740618/S.glyphs.zip)

simoncozens:
The hints in your file don't parse using the main branch of glyphsLib, even if I convert it to Glyphs 2 file format, so the fact that the hints don't parse is not specific to this PR. I think what you're saying is a duplicate of #446.

jenskutilek:
Right! Yes, they didn’t ever parse. I just thought you might want to fix it for G3 files :)

simoncozens:
Sure. Happy to fix for G2 *and* G3, but I'd rather do things in separate steps to help reviewers; the "rejigging the parser" component is big enough to be a single PR by itself, we can fix unconnected open issues later...

PR #652, opened 2020-12-15
Glyphs 3, attempt 3

simoncozens:

PR #691, opened 2021-06-28
Glyphs 3 (attempt 4)

simoncozens:
See also #690, which contains the first few commits.

This is essentially the same code as before, but organised into "topic" commits and broken down into meaningful-enough-sized commits for ease of review (oh, and rebased onto main, of course).


Lib/glyphsLib/parser.py
```diff
@@ -14,232 +14,88 @@
...
+
+    def parse(self, d):
+        try:
+            if isinstance(d, str):
+                d = openstep_plist.loads(d, use_numbers=True)
+            elif isinstance(d, bytes):
+                d = openstep_plist.loads(d.decode(), use_numbers=True)
+            result = self._parse(d)
+        except openstep_plist.parser.ParseError:
+            raise ValueError()
```
madig, line 41:
```python
except openstep_plist.parser.ParseError as e:
    raise ValueError("Failed to parse file") from e
```

Lib/glyphsLib/classes.py
```diff
@@ -1477,37 +1477,62 @@ def parent(self):
...
         writer.writeObjectKeyValue(self, "iconName", "if_true")
         writer.writeObjectKeyValue(self, "id")
         writer.writeObjectKeyValue(self, "italicAngle", "if_true")
+        if writer.format_version == 3:
+            writer.writeKeyValue("metricValues", self.metrics)
 
-        if self._name and self._name != self.name:
+        if (
+            self._name and self._name != self.name
+        ) or writer.format_version == 3:  # Check this
```
madig, line 1518: Hm?

Lib/glyphsLib/classes.py
```diff
@@ -4279,39 +4313,60 @@ def gridLength(self):
...
+        if not kerntable:
             return self.EMPTY_KERNING_VALUE
         try:
-            return self._kerning[fontMasterId][leftKey][rightKey]
+            return kerntable[fontMasterId][leftKey][rightKey]
         except KeyError:
             return self.EMPTY_KERNING_VALUE
 
     def setKerningForPair(self, fontMasterId, leftKey, rightKey, value, direction=LTR):
         # TODO: (jany) understand and use the direction parameter
```
madig, line 4365: Is this comment still relevant?

Lib/glyphsLib/builder/tokens.py
```diff
@@ -2,6 +2,13 @@
 from collections import OrderedDict
 
 
+def _like(got, expected):
+    expected = expected.replace("?", ".")
+    expected = expected.replace("*", ".*")
+    # Technically we should be a bit stricter than this
```
madig, line 8: How so?

Lib/glyphsLib/builder/tokens.py
```diff
@@ -50,7 +57,7 @@ def parse_bare_number_value(self, number):
         # Find index of number
         try:
             index = [metric.name for metric in self.font.numbers].index(number)
-        except ValueError as e:
+        except ValueError:
```
madig, line 60: Please keep the `as e` and below do a `raise ValueError(...) from e`.

Lib/glyphsLib/classes.py
simoncozens: I've checked it, it's true! (Glyphs 2 omits the master name if it's "Regular", Glyphs 3 always writes it.)
We can remove that comment.

madig:
I had a superficial look (too much code to go through) and nothing weird stood out to me. I wish we had a serde for Python... `cattrs` exists, but requires a rewrite of classes.py with `attrs`.

I'd like some typing information so that Pylance or mypy could have a meaningful pass to find edge-cases, but maybe that can be part of ongoing maintenance.

Also, consider running https://pypi.org/project/flynt/ over the code, f-strings are nicer in general.

Lib/glyphsLib/classes.py
simoncozens: That is done now, comment can be removed.

Lib/glyphsLib/builder/tokens.py
simoncozens: Because if the RHS of the predicate (`expected`) might contain other characters which are meaningful within a regular expression match. e.g. `$[name like "haha this is an invalid re ("]` will cause it to explode.

simoncozens:
> Also, consider running https://pypi.org/project/flynt/ over the code, f-strings are nicer in general.

I think that's ongoing maintenance too. This PR is big enough already. :-)

madig:
Mhkay.

Does the token parsing warrant pulling in a real PEG parser or something?

simoncozens:
Eh, you're welcome to implement that if you think it would beneficial. What I have works.

madig:
Oh also, have you tried some things like opening open-source fonts in .glyphs format in Glyphs 3, saving in the v3 format and making sure it roundtrips perfectly here? I keep finding nuggets in FontLab 7 when doing that, for example.

madig:
Jany pointed out that he wrote `tests/run_various_tests_on_various_files.py` once. I'll go spelunking a bit. A first run shows me a sea of red Fs.

simoncozens:
They don't pass with main branch either.

madig:
I'm fiddling with the listed source files a bit and see that there's some tricky business going on. Sometimes, we intentionally change data when roundtripping to preserve our sanity... I'll try to look at other things...

schriftgestalt:
Is there anything I can help with?

madig:
My worry currently is that this code berg introduces bugs that will halt our production work in its tracks. Therefore, I'd like to see more testing with real-world fonts, i.e. load a v2 source into Glyphs 3 and save as v3, then take both v2 and v3 and make sure they roundtrip perfectly. This will shake out random oversights. I'll do some manual testing on our side. I'm not sure if a DS<->Glyphs roundtrip is useful here, as that code is unaware of v3 specifics...

madig:
So, Jany suggested I do the following:

madig:
@simoncozens can you add me as a collaborator? I want to push stuff to this PR.

belluzj:
> Jany pointed out that he wrote tests/run_various_tests_on_various_files.py once. I'll go spelunking a bit. A first run shows me a sea of red Fs.

> They don't pass with main branch either.

About these scripts: they're not meant to be unit tests, but more of an exploratory tool to spot recurring problems, and then one should turn each category of problem into a unit test. Also, since they haven't been run in a while, they probably need updates to point at newer font files and maybe ignore a wider range of "acceptable" diffs

simoncozens:
> @simoncozens can you add me as a collaborator? I want to push stuff to this PR.

Done.

madig:
@googlebot I consent.

madig:
@schriftgestalt Is there a certain version where Glyphs 2 switched from using `\\012` to `\n` in feature code in .glyphs files?

schriftgestalt:
It didn’t. There is a custom parameter "Use Line Breaks" that would just leave the line breaks. But otherwise it should write /012. Do you have a file that shows otherwise?

stone-zeng:
I have tried to use this PR for compiling [firamath](https://github.com/firamath/firamath), but I fails. One issue can be reproduced by running the following code:

```py
from glyphsLib import GSFont, to_ufos
from glyphsLib.parser import Parser
from fontmake.font_project import FontProject

with open('bug-0716.glyphs') as f:
    font = Parser(current_type=GSFont).parse(f.read())
master_ufos = to_ufos(font)
FontProject().build_otfs(master_ufos, output_dir='build/')
```

With output:

```
INFO:fontmake.font_project:Building OTF for Bug0716-Light
INFO:ufo2ft:Pre-processing glyphs
INFO:ufo2ft.filters.base:Running EraseOpenCornersFilter on Bug0716-Light
  File "/Users/xdzeng/Documents/Fonts/_Projects/firamath-main/venv/lib/python3.9/site-packages/fontmake/font_project.py", line 383, in _iter_compile
    yield compile_func(ufo, debugFeatureFile=debugFeatureFile, **options)
  File "/Users/xdzeng/Documents/Fonts/_Projects/firamath-main/venv/lib/python3.9/site-packages/ufo2ft/__init__.py", line 132, in compileOTF
    glyphSet = preProcessor.process()
  File "/Users/xdzeng/Documents/Fonts/_Projects/firamath-main/venv/lib/python3.9/site-packages/ufo2ft/preProcessor.py", line 62, in process
    func(ufo, glyphSet)
  File "/Users/xdzeng/Documents/Fonts/_Projects/firamath-main/venv/lib/python3.9/site-packages/ufo2ft/filters/base.py", line 195, in __call__
    if include(glyph) and filter_(glyph):
  File "/Users/xdzeng/Documents/Fonts/_Projects/firamath-main/venv/lib/python3.9/site-packages/glyphsLib/filters/eraseOpenCorners.py", line 117, in filter
    contour.draw(p)
  File "/Users/xdzeng/Documents/Fonts/_Projects/firamath-main/venv/lib/python3.9/site-packages/ufoLib2/objects/contour.py", line 153, in draw
    self.drawPoints(pointPen)
  File "/Users/xdzeng/Documents/Fonts/_Projects/firamath-main/venv/lib/python3.9/site-packages/ufoLib2/objects/contour.py", line 178, in drawPoints
    pointPen.endPath()
  File "/Users/xdzeng/Documents/Fonts/_Projects/firamath-main/venv/lib/python3.9/site-packages/fontTools/pens/pointPen.py", line 155, in endPath
    self._flushContour(segments)
  File "/Users/xdzeng/Documents/Fonts/_Projects/firamath-main/venv/lib/python3.9/site-packages/fontTools/pens/pointPen.py", line 239, in _flushContour
    pen.closePath()
  File "/Users/xdzeng/Documents/Fonts/_Projects/firamath-main/venv/lib/python3.9/site-packages/glyphsLib/filters/eraseOpenCorners.py", line 39, in closePath
    self.endPath()
  File "/Users/xdzeng/Documents/Fonts/_Projects/firamath-main/venv/lib/python3.9/site-packages/glyphsLib/filters/eraseOpenCorners.py", line 65, in endPath
    for i in segmentSegmentIntersections(segs[ix - 1], segs[next_ix])
  File "/Users/xdzeng/Documents/Fonts/_Projects/firamath-main/venv/lib/python3.9/site-packages/fontTools/misc/bezierTools.py", line 1186, in segmentSegmentIntersections
    intersections = lineLineIntersections(*seg1, *seg2)
  File "/Users/xdzeng/Documents/Fonts/_Projects/firamath-main/venv/lib/python3.9/site-packages/fontTools/misc/bezierTools.py", line 937, in lineLineIntersections
    slope34 = (e2y - s2y) / (e2x - s2x)
ZeroDivisionError: division by zero

The above exception was the direct cause of the following exception:

  File "/Users/xdzeng/Documents/Fonts/_Projects/firamath-main/scripts/build.py", line 34, in <module>
    FontProject().build_otfs(master_ufos, output_dir='build/')
  File "/Users/xdzeng/Documents/Fonts/_Projects/firamath-main/venv/lib/python3.9/site-packages/fontmake/font_project.py", line 223, in build_otfs
    self.save_otfs(ufos, **kwargs)
  File "/Users/xdzeng/Documents/Fonts/_Projects/firamath-main/venv/lib/python3.9/site-packages/fontTools/misc/loggingTools.py", line 372, in wrapper
    return func(*args, **kwds)
  File "/Users/xdzeng/Documents/Fonts/_Projects/firamath-main/venv/lib/python3.9/site-packages/fontmake/font_project.py", line 521, in save_otfs
    for font, ufo in zip(fonts, ufos):
  File "/Users/xdzeng/Documents/Fonts/_Projects/firamath-main/venv/lib/python3.9/site-packages/fontmake/font_project.py", line 385, in _iter_compile
    raise FontmakeError("Compiling UFO failed", ufo.path) from e
fontmake.errors.FontmakeError: Compiling UFO failed: division by zero
```

This is the example glyph file:

[bug-0716.zip](https://github.com/googlefonts/glyphsLib/files/6827824/bug-0716.zip)

anthrotype:
@stone-zeng this looks like a bug in the eraseOpenCorners filter, can you please file a separate glyphsLib issue? thank you

madig:
> It didn’t. There is a custom parameter "Use Line Breaks" that would just leave the line breaks. But otherwise it should write /012. Do you have a file that shows otherwise?

Oh, interesting, the file in question actually has that parameter. Is it used for Glyphs 2 _and_ 3? Does v3 use a different escape policy?

stone-zeng:
> @stone-zeng this looks like a bug in the eraseOpenCorners filter, can you please file a separate glyphsLib issue? thank you

But this file can be compiled successfully using #652, so I'm a little confused whether it's related to other things or does this pr make something wrong.

simoncozens:
It's also certainly a bug in fonttools, in my line/line intersection code. (Oops)

Lib/glyphsLib/classes.py
```diff
@@ -4075,7 +4075,8 @@ def _serialize_to_plist(self, writer):
             if self.gridSubDivisions != 1:
                 writer.writeKeyValue("gridSubDivision", self.gridSubDivisions)
 
-        writer.writeObjectKeyValue(self, "instances")
+        if self.instances:
+            writer.writeObjectKeyValue(self, "instances")
```
simoncozens, line 4079: You can also use a third parameter `"if_true"` instead of the `if`.

madig:
Some code is tossing out the Axes custom parameter in https://github.com/madig/cantarell-fonts/blob/f17124d041e6ee370a9fcddcc084aa6cbf3d5500/src/Cantarell.glyphs.

madig:
@schriftgestalt in GSInstance, what is the difference between `exports` and `active`? Is the first one v2 and the latter v3?

madig:
I now have a quoting problem (strings gaining double-quotes) in FiraCode.glyphs, specifically for the glyph name "infinity" and some background image path. It doesn't happen in a test file I made though. I'll continue next week...

schriftgestalt:
> @schriftgestalt in GSInstance, what is the difference between exports and active? Is the first one v2 and the latter v3?

`.exports` is the objectiveC API and the key that is used in the .glyphs file. `.active` is the python API. Don’t remember why we thought it is a good idea to use different names.

madig:
Hm, and which one is written to disk under which circumstances?

schriftgestalt:
It should always ever only write `exports = 0;`

schriftgestalt:
The difference between g2 and g3 is the ordering of the items.

madig:
@schriftgestalt is there an efficient way in Glyphs 2.6.6 of serially opening 460 .glyphs files, saying ok to "but this font is made with an old glyphs.app oh and also how bout I upgrade components" and saving them again?

schriftgestalt:
```
path = "~/Stuff/New Font.glyphs"
path_new = "~/Stuff/New Font_new.glyphs"
font = GSFont(path)
font.save(path=path_new)
```

madig:
That will do all the upgrading etc.?

schriftgestalt:
It uses the same code path as when you save manually.

simoncozens:
I think Nikolaus wants to upgrade G2 files to G3 schema programatically.

madig:
No, I want to do my roundtrip tests against the latest v2 format to reduce
scope. Glyphs.app 2.6 does some upgrading of certain older stuff and I want
the save function to do the same without popping up a dialog.

schriftgestalt:
The dialog is only shown when the font is shown in the UI. But as I said, the headless script will produce the same result.

simoncozens:
If the problem here is that the current test suite does not give you sufficient confidence to accept large code changes, can I *very strongly* recommend you address that in a new PR? I tried my hardest to present this one in a way that was as compact and logical as possible, and now it’s filling up with noise again.

schriftgestalt:
I think the task is to make sure all files use the latest structure.

simoncozens:
As of the above commit, all our test-suite tests pass, apart from the one @madig added to do with quoting strings. Additionally, the regression test suite I added in #697, of 50 Google Fonts .glyphs files, also passes.

madig:
I backed out all of my commits and will make a separate branch instead.

madig:
I think I figured it out. `customParameters["Axes"]` is synthesized if `Axis Mappings` exists, which it does in the test case (copied verbatim from the passed in, generated Designspace).

simoncozens:
Yes - see most recent commit.

madig:
Ok, so what to do about the test? Is what it tests still useful?

simoncozens:
Sure, the test is fine and useful, and as of [this comment](https://github.com/googlefonts/glyphsLib/pull/691#issuecomment-883425221) (before you force-pushed!) it was passing. Not sure what you've done but I'll try to fix it tomorrow.

madig:
Weird, I have a copy of the pre force tree and it has the same failure. Yes
let's check tomorrow.

simoncozens:
[Very weird](https://github.com/googlefonts/glyphsLib/runs/3114513337?check_suite_focus=true#step:5:105).

simoncozens:
Right. This passes all glyphs unit tests and the 50 regression tests and I'm going to stop fiddling with it again.

madig:
Thanks. I'll carry on in a different branch.

schriftgestalt (commented):
In Glyphs 3 I just added a ‘Axis location’ parameter to the instances (the structure is the same as in the masters). I found this much easier to define the axis mappings.

madig:
@schriftgestalt So, is that a new feature in an upcoming Glyphs.app version or something that's already shipping?

schriftgestalt:
This is shipping in 3.0.3

madig:
@simoncozens please rebase your branch on main (or have me do it).

simoncozens:
I couldn't rebase because of conflicts in the requirements-dev file, but I could merge.

madig:
Hrm, the rebase conflicts in requirements files can be worked around by regenerating them with pip-tools' pip-compile.

simoncozens:
Yeah, I did that in the merge commit, and now we have incompatible requirements. Aargh.

simoncozens:
Something's very wrong here: `openstep_plist` should be marked as a dependency. And the tests should fail if we're not installing it?!

madig:
It is a dependency? I can see it in setup.cfg and requirements.txt?

madig:
Oh I see, the regression test should fail but doesn't. I think that's because I marked it as can-fail. Let's try without...

simoncozens:
OK, I see it. We're checking out `main`, installing the dependencies from *that* (which does not have the new openstep dep), then checking out the branch later. We should probably add another step to the workflow which re-installs the dependencies after switching branches.

madig:
I just saw that, yes. Let's see.

madig:
Something is flipping smooth flags :o

simoncozens:
But only some of them. And I can't reproduce locally. Annoying.

simoncozens:
I can reproduce this now. I don't understand it, but some clues: it only happens in (i) background glyphs (ii) which are made up of components and (iii) in the regression test framework - not via glyphs2ufo.

madig:
- Sometimes, the flag is added: https://github.com/googlefonts/glyphsLib/pull/691/checks?check_run_id=3238636620#step:8:6536
- Other times, it is removed: https://github.com/googlefonts/glyphsLib/pull/691/checks?check_run_id=3238636620#step:8:6597

simoncozens:
The diff is the wrong way around.

madig:
Here it is added, no? https://github.com/googlefonts/glyphsLib/pull/691/checks?check_run_id=3239793037#step:8:2297

simoncozens:
That's not the point. The point is, which is right? Which is wrong?

simoncozens:
One which is added:

```
! /tmp/tmp0ujjlwis/Playball-Regular.ufo/glyphs.public.background/uhungarumlaut.glif - tests/data/gf/Playball/Playball-Regular.ufo/glyphs.public.background/uhungarumlaut.glif--- 
+++ 
@@ -62,7 +62,7 @@
       <point x="345" y="345" type="curve" smooth="yes"/>
       <point x="305" y="281"/>
       <point x="211" y="61"/>
-      <point x="148" y="59" type="curve"/>
+      <point x="148" y="59" type="curve" smooth="yes"/>
       <point x="127" y="59"/>
       <point x="120" y="75"/>
       <point x="120" y="97" type="curve" smooth="yes"/>
```

Glyphs source has `"148 59 CURVE SMOOTH"` so master is correct here. OK, I got confused earlier.

But I have no idea why this is happening. And why it is happening now, when it wasn't happening before.

simoncozens:
However, I am getting the same result (no "smooth" for that point) on `main` now locally!

simoncozens:
I bet you this is the problem:

```
-fonttools[ufo,unicode]==4.21.1
+fonttools[ufo,unicode]==4.25.2
```

simoncozens:
```
$ cat is-point-smooth.py
import glyphsLib
import logging
import fontTools

logging.basicConfig(level=logging.ERROR)
ufos = glyphsLib.load_to_ufos("tests/data/gf/Playball.glyphs")

print(fontTools.__version__)
print(ufos[0].layers["public.background"]["uhungarumlaut"][0][60].smooth)
```

```
$ python3 is-point-smooth.py
4.21.1
True
```

```
$ python3 is-point-smooth.py
4.25.2
False
```

Now what?

simoncozens:
Ship it!

madig:
Merging this now and tagging a pre-release. Thanks!

PR #696, opened 2021-07-16
unicodes of a glyph with multiple unicodes parsed as list of lists ?!

jansindl3r:
this solves a problem where unicodes were parsed as list of lists of one child in a file saved glyphs 3.0.3 (3091). Please check it

### DIFF
--- a/Lib/glyphsLib/types.py
+++ b/Lib/glyphsLib/types.py
@@ -376,8 +376,18 @@ class UnicodesList(list):
             unicodes = []
         elif isinstance(value, str):
             unicodes = value.split(",")
+        elif isinstance(value, list) or isinstance(value, tuple):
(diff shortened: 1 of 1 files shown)


jansindl3r:
sorry, my tests failed, but they failed even right after cloning it

jansindl3r:
@googlebot I signed it!

jansindl3r:
[Template.glyphs.zip](https://github.com/googlefonts/glyphsLib/files/6829963/Template.glyphs.zip)
This is the file

schriftgestalt (commented):
I wonder if this is the right place to fix this. Where does the nested list comes from?

jansindl3r:
@schriftgestalt you are right! But I didn't really find where it comes from

PR #697, opened 2021-07-20
Add a regression test framework

simoncozens:
This adds 50 OFL-licensed Glyphs files from the Google Fonts catalogue, and (in a GitHub actions runner) generates UFOs from them using a known-good branch of glyphsLib. It then checks out the pull request branch, generates new UFOs to a temporary directory and compares them.

Note that currently, the first branch checked out is this one (`regression-tests`) because the test framework and generator isn't in a "known-good" branch yet. So when this lands in main, the first thing to do will be uncomment the code which uses `main` as the reference branch.

"But you aren't comparing glyphs-to-glyphs or designspace-roundtripping!" Yeah, I'm not. I'm sure a few people are using glyphsLib for *generating* .glyphs files, but the thing we *really want to work* is the Glyphs-to-UFO pipeline.


madig:
@simoncozens I'm ready to merge this but googlebot seems to have a problem?!

madig:
Wait, everything is actually failing: https://github.com/googlefonts/glyphsLib/pull/697/checks?check_run_id=3113067799#step:7:15

simoncozens:
@googlebot I consent

simoncozens:
I’ll take another look. Looks like the second checkout might be trashing the working directory.

madig:
Maybe also consider adding https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstepscontinue-on-error to the test step to stop it from cancelling other CI jobs when we intentionally change something.

simoncozens:
I'm not sure we need that if we are talking about jobs in separate workflows. (ie. a fail in the regression test workflow will not stop jobs in the Test + Deploy workflow)

madig:
Thanks, here we go.

simoncozens:
Thanks; although we may need your "pass through exit code" (28fc96ab) commit in there for it to actually work.

simoncozens:
Also now we can set the "known-good" baseline to `main` in regression.yml. I'll do that in a separate PR.

madig:
Yeah, I'll rebase my branch on main next and do some fiddeling.

PR #698, opened 2021-07-21
Extend regression test framework to test Designspace -> Glyphs

madig:
This PR adds regression tests in the other direction, from Designspace to Glyphs.


madig:
@googlebot I consent.

madig:
I'm gonna try and look into using pytest-xdist to make the tests go faster. It seems to struggle with the way gf_regression.py is written though, with pytest being called from inside? Not sure. I'll look into using https://docs.pytest.org/en/6.2.x/example/simple.html#control-skipping-of-tests-according-to-command-line-option to make a normal regression_test.py file and mark tests accordingly.

madig:
@simoncozens whatchu think????

simoncozens:
GlyphsToDesignspace and DesignspaceToGlyphs are crying out to be refactored in terms of a common base class. Other than that, looks good!

madig:
You're right, and using parametrized tests makes everything even tidier. It's like pytest knows what I need.

PR #699, opened 2021-07-22
Use main as the comparison point for regression testing

madig:
Followup from https://github.com/googlefonts/glyphsLib/pull/697.


simoncozens:
Awesome, you're way ahead of me. :)

PR #706, opened 2021-08-20
Fix saving version 3 fonts

khaledhosny:
`Writer()` defaults to format version 2 and `GSFont.save()` did not set version on the `Writer` object making it think it is saving format 2 regardless of the font format.

### DIFF
--- a/Lib/glyphsLib/classes.py
+++ b/Lib/glyphsLib/classes.py
@@ -1354,7 +1354,7 @@ class GSCustomParameter(GSBase):
 
     def plistValue(self, format_version=2):
         string = StringIO()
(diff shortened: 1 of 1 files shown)


khaledhosny:
I don’t know how to add a test for this, but I caught it because glyphsLib would save Unicode values as unescaped hex strings instead of integers and reading that back failed when the hex strings didn’t look like valid integers, but I’m not sure that is not the only broken thing.

There is another instance of creating `Writer` object withpout setting `format_version` (in `GSCustomParameter`), probably it should be fixed too.

simoncozens (approved):

PR #707, opened 2021-08-20
Fix converting vendorID from version 3 fonts

khaledhosny:
`vendorID` is not a custom parameter any more.


madig:
And what about v2 files?

khaledhosny:
This tuple is for v2 custom parameters that became properties in v3, the first item in the sub-tuple is the v2 custom parameter, and the last is the v3 property.

simoncozens (approved):

PR #719, opened 2021-09-16
Accept FL7 trailing semicolons

chrissimpkins:
Workaround for the FontLab 7 glyphs source export trailing semicolon that raises an error on fontmake compiles as noted in https://github.com/googlefonts/fontmake/issues/806.

### DIFF
--- a/Lib/glyphsLib/parser.py
+++ b/Lib/glyphsLib/parser.py
@@ -48,15 +48,9 @@ class Parser:
         This method removes the semicolon before passing the string to the parser."""
         # see https://github.com/googlefonts/fontmake/issues/806
         if isinstance(d, str):
-            if d.endswith(";"):
-                d = d.rstrip(";")
-            elif d.endswith(";\n"):
-                d = d.rstrip(";\n")
+            d = d.rstrip(";\n")
         elif isinstance(d, bytes):
-            if d.endswith(b";"):
-                d = d.rstrip(b";")
-            elif d.endswith(b";\n"):
-                d = d.rstrip(b";\n")
+            d = d.rstrip(b";\n")
         return d
 
     def _parse(self, d, new_type=None):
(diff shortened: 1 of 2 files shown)


Lib/glyphsLib/parser.py
```diff
@@ -33,14 +33,32 @@ def __init__(self, current_type=OrderedDict, format_version=2):
...
 
+    def _fl7_format_clean(self, d):
+        """FontLab 7 glyphs source format exports include a final closing semicolon.
+        This method removes the semicolon before passing the string to the parser."""
+        # see https://github.com/googlefonts/fontmake/issues/806
+        if isinstance(d, str):
+            if d.endswith(";"):
+                d = d.rstrip(";")
+            elif d.endswith(";\n"):
+                d = d.rstrip(";\n")
```
belluzj, line 54:
```suggestion
            d = d.rstrip(";\n")
```

With one call to rstrip it would handle all the cases

Lib/glyphsLib/parser.py
```diff
@@ -33,14 +33,32 @@ def __init__(self, current_type=OrderedDict, format_version=2):
...
+        elif isinstance(d, bytes):
+            if d.endswith(b";"):
+                d = d.rstrip(b";")
+            elif d.endswith(b";\n"):
+                d = d.rstrip(b";\n")
```
belluzj, line 59:
```suggestion
            d = d.rstrip(b";\n")
```

same here

Lib/glyphsLib/parser.py
chrissimpkins: Ty!

https://github.com/chrissimpkins/glyphsLib/commit/56a428942ccbd4ae9b227dd69a779883ac6afa5d

Lib/glyphsLib/parser.py
chrissimpkins: https://github.com/chrissimpkins/glyphsLib/commit/56a428942ccbd4ae9b227dd69a779883ac6afa5d

chrissimpkins:
Addressed the unrelated flake8 lint CI fail in https://github.com/googlefonts/glyphsLib/pull/720

PR #721, opened 2021-09-22
Fix same-side-of-line test for open corners

### LINKED ISSUE #703: EraseOpenCornersFilter crashes on collinear segments, may produce incompatible outlines
https://daltonmaag.github.io/pipeline-perf-tracker/results/ shows failures for Sofia Sans and Noto Sans LCG.

> fontmake: Error: In '../../../downloads/narrow_build_vf_from_glyphs/Sofia_Sans__960_glyphs__7_masters_/SofiaSans/SofiaSans-VF-New.glyphs' -> 'master_ufo/SofiaSans.designspace': Generating fonts from Designspace failed: division by zero

> ERROR:cu2qu.ufo:Glyphs named 'uni1D21' have different number of segments
> Got exit code 0, exiting
> fontmake: Error: In '../../../downloads/narrow_build_vf_from_glyphs/Noto_Sans_LCG__3780_glyphs__8_masters_/NotoSans/NotoSans-MM.glyphs' -> 'master_ufo/NotoSans.designspace': Generating fonts from Designspace failed: fonts contains incompatible glyphs: 'uni1D21'

simoncozens:
This should fix #703 - or at least, *some* of the problems.


Lib/glyphsLib/filters/eraseOpenCorners.py
```diff
@@ -57,7 +57,7 @@ def endPath(self):
             # of the next segment both on the right side of the line?
             # (see discussion at https://github.com/googlefonts/glyphsLib/pull/663)
             pt1 = segs[ix - 1][0]
-            pt2 = segs[next_ix][-1]
+            pt2 = segs[next_ix][1]
```
madig, line 60: This indeed fixes the Sofia Sans `kgreenlandic` problem :) I don't understand the logic, though, is the comment above up to date?

simoncozens: Yes. The comment is fine. But the code before only worked for line segments, not curves. This is an example of the problem:



We know something is an open corner if both "points" are on the same side of the yellow line. But which points? Previously we were looking at the *ends* of the surrounding segments ( (444,362) and (432,213) ). But if you extend the yellow line, you'll find that (432,213) is actually on the other side of it! We should instead be checking the outgoing handle (353,296).

Hence, point 2 is either the outgoing handle for cubics or the end of the line segments for lines - in either case, `curve[1]`, not `curve[-1]`.

simoncozens: Sorry, the comment says "last point", and you're right, that's outdated.

simoncozens: Oh, this reveals another bug...

simoncozens:
Found a few more issues.

simoncozens:
There's a problem with line-curve intersection testing. (Aaargh.)

> Testing for intersections between ((330.0000432054082, 365.6789020602332), (330, 286)) and ((315, 296), (404, 296), (468, 251), (468, 183))
> No intersections

simoncozens:
It's a numerical precision/limits bug in geometry calculations in fontTools, my favourite.

simoncozens:
Test fails now, will pass when https://github.com/fonttools/fonttools/pull/2413 is merged.

Lib/glyphsLib/filters/eraseOpenCorners.py
```diff
@@ -53,25 +58,44 @@ def endPath(self):
                 ix = ix + 1
                 continue
 
+            logger.debug("Considering line segment (%i,%i)-(%i,%i)" % (*segs[ix][0], *segs[ix][1]))
```
anthrotype, line 61: if you're going to keep the logging in, at least try not to use the normal `%` string interpolation which is done all the time, but let logging module do it only when the message is actually emitted.
that is, replace:

```diff
- logging.debug("%s %s" % ("hello", "world"))
+ logging.debug("%s %s", "hello", "world")

File: tests/data/EraseOpenCornersIncompatibilityTest-Two.ufo/glyphs/incompatible.glif [hunk 3]
```diff
@@ -1,20 +1,20 @@
...
-  <lib>
-    <dict>
-      <key>com.schriftgestaltung.Glyphs.lastChange</key>
-      <string>2021/09/22 09:53:10</string>
-    </dict>
-  </lib>
+	<advance width="652"/>
+	<outline>
+		<contour>
+			<point x="464" y="189" type="curve"/>
```
anthrotype, line 6: why the inconsistent tap/space indentation here? I suggest you rewrite the test ufo so they match the original 2-space indent (open and save back using ufoLib2)

anthrotype: nevermind sorry, I was misled by Github's UI that showed the first indent level narrower than the second, making me believe you were mixing tab and spaces. I don't actually care whether you use one or the other, just not both. Ignore my comment

anthrotype:
> Test fails now, will pass when fonttools/fonttools#2413 is merged.

@simoncozens let me know when you are ready to tag a new fonttools release, I can do that for you

Lib/glyphsLib/filters/eraseOpenCorners.py
simoncozens: Oh, that's nice.

simoncozens:
fonttools#2411 is in there now too, so I think we're good to go for a new release. Thank you!

anthrotype:
done https://github.com/fonttools/fonttools/releases/tag/4.27.1

simoncozens:
As if by magic!

PR #725, opened 2021-09-27
drop Python 3.6, require >= 3.7

anthrotype:
3.10 is out next week, while 3.6 reaches end of life at the end of this year.
It's time to require 3.7 or greater, so we can finally use built-in dataclasses and more.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -27,7 +27,7 @@ jobs:
     if: "! contains(toJSON(github.event.commits.*.message), '[skip ci]')"
     strategy:
       matrix:
-        python-version: [3.9, 3.6]
+        python-version: [3.9, 3.7]
         platform: [ubuntu-latest, windows-latest]
     steps:
     - uses: actions/checkout@v2
--- a/Lib/glyphsLib/glyphdata.py
+++ b/Lib/glyphsLib/glyphdata.py
@@ -94,11 +94,7 @@ def get_glyph(glyph_name, data=None):
     if data is None:
         global GLYPHDATA
         if GLYPHDATA is None:
-            try:
-                from importlib.resources import open_binary
-            except ImportError:
-                # use backport for python < 3.7
-                from importlib_resources import open_binary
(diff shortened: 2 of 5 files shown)


khaledhosny:
Should we release 6.0.0 before or after this is merged?

anthrotype:
I think we could do it at the same time as we release 6

khaledhosny:
OK, it would be nice to cut a release today so we can start using it with fontmake (pip is refusing to install both fontmake and the beta, I think because of the major version number change).

khaledhosny:
> > pip is refusing to install both fontmake and the beta
> 
> that was intentional, to use the beta you have to use fontmake > 2.4.2 from main branch, there's a commit [googlefonts/fontmake@5f98037](https://github.com/googlefonts/fontmake/commit/5f980376120d9cdc328965aa170eeb108bdd0002) following v2.4.2 tag that unlocks glyphsLib again
> I wanted to make sure that installing fontmake==2.4.2 would install glyphsLib < 6

I see, but this would require installing from git repository, but I’d rather avoid that. Anyway, I still want to switch fontmake to use the `minimal` option and see what happens.

anthrotype:
we can also publish a pre-release for fontmake that uses the newer glyphsLib if you like. I just want to avoid unnecessary breakages following this major glyphsLib upgrade /cc @simoncozens @madig @belluzj

khaledhosny:
That would work, yes.

khaledhosny:
With `minimal` would be best, but even without it would be OK (I currently often use `glyphs2ufo` then `fontmake` in two separate steps anyway).

anthrotype:
I'll do it after your make the `minimal` PR. Maybe in the meantime you can install fontmake==2.4.1 (the one immediately preceding 2.4.2), as that does not force glyphsLib to use a "compatible version" `~=5`, but uses `>=...`

khaledhosny:
Sounds good.

madig:
Ok with me.

belluzj:
Yes fine with me too

Issue #715, opened 2021-09-03
[Glyphs3] support special layers


RosaWagner:
Font doesn't build if has intermediate or alternate layers

Screenshot from a collaborator (not practical sorry):
![bfb1fd4f-c2aa-4e91-ac6f-1ceeb5c93c45](https://user-images.githubusercontent.com/12222436/132017184-4508a67b-30f8-4349-bfed-944001eb4609.jpg)

Easy to go back to format 2 using intermediate layers, but not for alternate ones.
I'll try to find a test case in Google Fonts and link it here.

RosaWagner:
Forget about this, my collaborator was working with the last release which is not up to date.
So update: it works with intermediate layers. I can't confirm for alternate ones.

RosaWagner:
Update: when using format 3 and this main branch of GlyphsLib, I confirm the font builds if it has alternate layers, but the interpolation of the said glyphs (with alternate layers) don't work.

RosaWagner:
This is a test case for @simoncozens. 
Bold style should be a square (like when exporting with Glyphs), but with fontmake, it is still a circle.
[alternate.zip](https://github.com/googlefonts/glyphsLib/files/7285639/alternate.zip)

simoncozens:
Thank you!

RosaWagner:
> (It may be possible to have alternate layers on other axes than the first without too much trouble, but I deliberately want to do this in fairly small steps.) (#729)

[My branch of Literata](https://github.com/RosaWagner/literata) has sources with alternate layers for next step:
- alternate layers for 2 axes
- min or max value only on each axis

This is a minimal test-case:
[alt-2axes-test.zip](https://github.com/googlefonts/glyphsLib/files/7823618/alt-2axes-test.zip)

A: no layers
B: alternate layer with location `axis0: min, 120`, `axis1: min, 600`
C: alternate layer with location `axis0: 120, max`, `axis1: 600, max`
D: alternate layer with location `axis0: min, 120`, `axis1: 600, max `

anthrotype:
I think it's fine if at least initially we only support these when glyphs -> UFO.

madig:
Ok with me, we can add more support later.

schriftgestalt:
I removed the backwards bracket layer rules in Glyphs 3. Too complicated. What I do now in cases where one would use a backwards bracket (`]100]`) in Glyphs 2, I add the rule to the master layer and have an empty rule in the alternate.

[...]

PR #729, opened 2021-10-06
Support (some) Glyphs 3 alternate layers

simoncozens:
This PR partially addresses #715. Notably, it only deals with alternate layers:

* Which apply to one axis only
* Which apply only to the *first* axis
* Which have a max value or a min value but not both

I believe this is all that can be done without serious reworking of how alternate layers are handled by the builders, in particular the way that we name them on export. (It may be possible to have alternate layers on other axes than the first without too much trouble, but I deliberately want to do this in fairly small steps.)

This may be easier to review commit-by-commit as I have done things in small discrete changes

### DIFF
--- a/Lib/glyphsLib/builder/builders.py
+++ b/Lib/glyphsLib/builder/builders.py
@@ -432,16 +432,24 @@ class UFOBuilder(_LoggerMixin):
                 raise ValueError(
                     "For now, bracket layers can only apply to the first axis"
                 )
+
+            # Convert [500<wght<(max)] to [500<wght], etc.
+            if bracket_min == bracket_axis_min:
+                bracket_min = None
+            if bracket_max == bracket_axis_max:
+                bracket_max = None
+
+            if bracket_min is not None and bracket_max is not None:
+                raise ValueError("Alternate rules with min and max range not yet supported")
+            if bracket_min is None and bracket_max is None:
+                continue
+
             glyph_name = layer.parent.name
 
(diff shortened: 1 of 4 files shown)


anthrotype:
pardon my ignorance, what are "Glyphs 3 alternate layers"? Is it a new way to encode 'bracket' layers?

simoncozens:
Sort of.

Glyphs 2 bracket layers can specify a "switch point" and be either forwards or reverse - so `[300]` applies from 300-`axis_max` on the first axis and `]300]` applies from `axis_min-300` on the first axis.

Glyphs 3 alternate layers do a lot more than that; they directly specify a min and/or a max on each axis. So in G3 you could have an axis that applies `[wght<500]` or `[500<wght]` or even `[300<wght<500]` - or even `[500<wght,12<opsz<16]`.

The min-max range and multiple-axis cases are going to be very difficult to manage with our current setup; we will need new glyph naming conventions, and will need to tread carefully if we want to prioritise back-conversion from Designspace->glyphs.

moyogo:
It is documented on https://glyphsapp.com/learn/switching-shapes#g-1-alternate-layers-bracket-layers

RosaWagner:
Saira is a good candidate to test this PR (has several simple one-axis bracket layers) and the alternate shape got exported properly :)

@anthrotype yes, brace and bracket layers were quite confusing compare to "intermediate" and "alternate", and translation probably complicated.

Issue #717, opened 2021-09-13
[Glyphs3] NameID 2 is Regular when font is Italic.


RosaWagner:
Inconsistent name table when exporting from format 2 or format 3.
Italic fonts gets nameID2=Regular and Italic appended to nameID1 when exporting VF with format 3. 

I noticed Glyphs3 does the same thing, which I believe is wrong because it breaks style linking in desktop app such as Word (which creates then instances called "`FontName Italic Italic` cause Italic is in Name ID1 and in STAT table etc.)

When exporting with format 2:
```
    <namerecord nameID="1" platformID="3" platEncID="1" langID="0x409">
      Texturina Medium
    </namerecord>
    <namerecord nameID="2" platformID="3" platEncID="1" langID="0x409">
      Italic
    </namerecord>
```

When exporting with Format 3:
```
    <namerecord nameID="1" platformID="3" platEncID="1" langID="0x409">
      Texturina Medium
    </namerecord>
    <namerecord nameID="2" platformID="3" platEncID="1" langID="0x409">
      Italic
    </namerecord>
```

[Texturina-exports.zip](https://github.com/googlefonts/glyphsLib/files/7153083/Texturina-exports.zip)
cc @simoncozens @schriftgestalt

RosaWagner:
I just edited the issue cause the first time I mixed up the copy-pastas.
And reattach files here:

### [Texturina-Italic_Exports.zip](https://github.com/googlefonts/glyphsLib/files/7153179/Texturina-Italic_Exports.zip)

PR #730, opened 2021-10-06
Support Glyphs 3 GSInstance family names

### LINKED ISSUE #718: [Glyphs3] issue with FamilyName in instance setting and FVAR table
Fontmake has a Family Name argument. When specified, only the instances with that family name get into the FVAR table. It works when the font is Glyphs format 2. When I switch to format 3, ALL the instances set up with different family names (eg Texturina 12pt — Bold) get into the FVAR. The result is multiple instances of the same style name in the FVAR.

[TexturinaVF-2-3.zip](https://github.com/googlefonts/glyphsLib/files/7156448/TexturinaVF-2-3.zip)

simoncozens:
Fixes #718

### DIFF
--- a/Lib/glyphsLib/builder/instances.py
+++ b/Lib/glyphsLib/builder/instances.py
@@ -58,9 +58,7 @@ def _to_designspace_instance(self, instance):
     # at least according to https://docu.glyphsapp.com/#fontName
     for p in instance.customParameters:
         param, value = p.name, p.value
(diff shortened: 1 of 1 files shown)


anthrotype:
thanks, would be nice to also add test

simoncozens:
Not merging this yet - I think there's some relationship between this and #717, as the instance names in the designspace are turning out to different between G2 and G3.

PR #749, opened 2021-11-09
Fix saving quadratic curves in Glyphs 3

khaledhosny:


anthrotype (commented):
ouch! thank you Khaled. LGTM

PR #750, opened 2021-11-17
Handle Glyphs 3 layer attributes a bit better

khaledhosny:
Glyphs 3 has explicit layer attributes for things that Glyphs 2 used magic layer names for, and layer names has no special meanings. We often check for layer attributes and unconditionally fallback to layer name, but this can lead to undesired effects when layer names in Glyphs 3 happen to match magic Glyphs 2 names.

To fix this we now check for layer attributes in Glyphs 3 and layer names in Glyphs 2.

### DIFF
--- a/Lib/glyphsLib/builder/sources.py
+++ b/Lib/glyphsLib/builder/sources.py
@@ -151,30 +151,21 @@ def _to_designspace_source_layer(self):
 
     # First, collect all brace layers in the font and which glyphs and which masters
     # they belong to.
-    layer_name_to_master_ids = collections.defaultdict(set)
-    layer_name_to_glyph_names = collections.defaultdict(list)
+    layer_to_master_ids = collections.defaultdict(set)
+    layer_to_glyph_names = collections.defaultdict(list)
     for glyph in self.font.glyphs:
         for layer in glyph.layers:
-            if (
-                "{" in layer.name
-                and "}" in layer.name
-                and ".background" not in layer.name
-            ):
-                layer_name_to_master_ids[layer.name].add(layer.associatedMasterId)
-                layer_name_to_glyph_names[layer.name].append(glyph.name)
+            if layer._is_brace_layer():
+                key = (layer.name, tuple(layer._brace_coordinates()))
+                layer_to_master_ids[key].add(layer.associatedMasterId)
+                layer_to_glyph_names[key].append(glyph.name)
 
     # Next, insert the brace layers in a defined location in the existing designspace.
     designspace = self._designspace
     layers_to_insert = collections.defaultdict(list)
-    for layer_name, master_ids in layer_name_to_master_ids.items():
-        # Construct coordinates first...
-        brace_coordinates = [
-            float(c)
-            for c in layer_name[
-                layer_name.index("{") + 1 : layer_name.index("}")
-            ].split(",")
-        ]
-
+    for key, master_ids in layer_to_master_ids.items():
+        brace_coordinates = list(key[1])
+        layer_name = key[0]
         for master_id in master_ids:
             # ... as they may need to be filled up with the values of the associated
             # master.
@@ -189,7 +180,7 @@ def _to_designspace_source_layer(self):
                 logger.warning(
                     "Glyph(s) %s, brace layer '%s' defines more locations than "
                     "there are design axes.",
-                    layer_name_to_glyph_names[layer_name],
+                    layer_to_glyph_names[key],
                     layer_name,
                 )
 
--- a/Lib/glyphsLib/classes.py
+++ b/Lib/glyphsLib/classes.py
@@ -3783,6 +3783,28 @@ class GSLayer(GSBase):
         coordinates = name[name.index("{") + 1 : name.index("}")]
         return [float(c) for c in coordinates.split(",")]
 
+    def _is_color_palette_layer(self):
+        if self.parent.parent.format_version > 2:
+            return "colorPalette" in self.attributes  # Glyphs 3
+        return self.name.startswith("Color ")  # Glyphs 2
+
+    def _color_palette_index(self):
+        if not self._is_color_palette_layer():
+            return None
+
+        if self.parent.parent.format_version > 2:
+            # Glyphs 3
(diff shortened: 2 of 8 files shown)


Lib/glyphsLib/classes.py
```diff
@@ -3760,6 +3765,46 @@ def _bracket_info(self):
         else:
             return axis_index, bracket_crossover, None
 
+    def _is_brace_layer(self):
+        if self.parent.parent.format_version > 2:
+            return "coordinates" in self.attributes  # Glyphs 3
+        # Glyphs 2
+        return "{" in self.name and "}" in self.name and ".background" not in self.name
```
anthrotype, line 3772: or should it be `self.name.endswith(".background")`?

Lib/glyphsLib/classes.py
```diff
@@ -3760,6 +3765,46 @@ def _bracket_info(self):
...
+
+        # Glyphs 2
+        name = self.name
+        coordinates = name[name.index("{") + 1 : name.index("}")]
+        return [float(c) for c in coordinates.split(",")]
+
+    def _is_color_palette_layer(self):
+        if self.parent.parent.format_version > 2:
+            return "colorPalette" in self.attributes  # Glyphs 3
+        return self.name.startswith("Color ")  # Glyphs 2
```
anthrotype, line 3789: maybe a regex would be better, so you only match `^Color (?:\*|\d+)$`, instead of any layer that starts with "Color "

Lib/glyphsLib/builder/sources.py
```diff
@@ -151,30 +151,23 @@ def _to_designspace_source_layer(self):
...
-        brace_coordinates = [
-            float(c)
-            for c in layer_name[
-                layer_name.index("{") + 1 : layer_name.index("}")
-            ].split(",")
-        ]
-
+    for key, master_ids in brace_coordinates_to_master_ids.items():
+        brace_coordinates = list(key)
+        layer_name = "{" + ", ".join(f"{c:g}" for c in brace_coordinates) + "}"
```
anthrotype, line 170: hm so your're now discarding the original layer name.. why? I think we should try to retain the original brace layer names.

anthrotype (commented):
LGTM, only a couple questions (see below)

Lib/glyphsLib/builder/sources.py
schriftgestalt: I would not use the layer internally any more. Wouldn't it be easier to use the attributes and converting layer names to attributes for v2 files?

khaledhosny: > I would not use the layer internally any more. Wouldn't it be easier to use the attributes and converting layer names to attributes for v2 files?

This is what I wanted to do originally, but format version is not always known when layer name is set, so we can't set the attributes there, and I don't know how to make attributes some kind of dynamic dictionary that would check layer name when an attribute is accessed (and whether or not this would have performance imprecations).

khaledhosny: > hm so your're now discarding the original layer name.. why? I think we should try to retain the original brace layer names.

This is to make the tests pass with V3 file, since one of the tests is checking design space layer name and using V3 layer name makes no sense as it can be anything.

khaledhosny: This is also inline with Glyphs 3, where the displayed layer name is always constructed for layers with attributes.

Lib/glyphsLib/classes.py
khaledhosny: I copied the existing code and tried not to change it since I don't know exactly what it is matching.

Lib/glyphsLib/builder/sources.py
schriftgestalt: I have a read only property `nameUI` for that.

Lib/glyphsLib/classes.py
khaledhosny: Done.

Lib/glyphsLib/builder/sources.py
anthrotype: > inline with Glyphs 3, where the displayed layer name is always constructed for layers with attributes.

I see, but for Glyphs 2 brace layer names could contain some other string that added context to the numbers in the parentheses. I'd still consider it a (minor?) regression to lose that.

anthrotype (commented):
LGTM, with caveat about Glyphs2 brace layer names no longer roundripping (see [comment](https://github.com/googlefonts/glyphsLib/pull/750/files#r751100832))

Lib/glyphsLib/builder/sources.py
khaledhosny: I don’t mind reverting this part, but I’ll need a way to make `test_designspace_generation_brace_layers` test pass without it.

khaledhosny: Actually this change is brooken, since now designspace layer names don’t actually match UFO layer names. Let me try to fix it.

khaledhosny: Should be fixed now (I hop, I this the use of layer names in V3 files is wrong, but fixing this probably is a bigger change than this place.

Lib/glyphsLib/builder/sources.py
```diff
@@ -151,30 +151,21 @@ def _to_designspace_source_layer(self):
...
-    for layer_name, master_ids in layer_name_to_master_ids.items():
-        # Construct coordinates first...
-        brace_coordinates = [
-            float(c)
-            for c in layer_name[
-                layer_name.index("{") + 1 : layer_name.index("}")
-            ].split(",")
-        ]
-
+    for key, master_ids in layer_to_master_ids.items():
```
anthrotype, line 166: optional: you could unpack the `key` directly in the for loop variable as `for (layer_name, brace_coordinates), master_ids in ...`

khaledhosny: I need to convert `brace_coordinates` to a list anyway, so it wouldn’t save much lines.

PR #751, opened 2021-11-25
[builder] Support Glyphs 3 Color layers

khaledhosny:
Glyphs 3 supports a new kind of color layers that allows settings solid colors and gradients (among others) directly on the paths.

Glyphs supports exporting these layers to SVG table, but ufo2ft does not support building SVG table. We instead setup color layers in a way that allows ufo2ft to build a COLRv1 table from them.

There are few caveats:

* This is done only in minimal mode, since I have no idea how to setup these layers in a way that other UFO consumers can use, or that can round-trip back to Glyphs format.
* ~~Components in color layers will cause an assert, as I’m not sure what to do with the since Glyphs does not allow setting attributes on them and exports them with very thin stroked path in SVG table (COLRv1 does not yet support stroking).~~ Components are turned into `PaintColrGlyph` paints if component glyph has Color layers, otherwise they get decomposed and handled as attribute-less paths.
* ~~The only supported path attributes are `gradient` and `fillColor`, any other attribute will raise a `NotImplementedError`.~~ `strokeWidth` and `strokeColor` are also supported using skia-pathops stroking.
* ~~Paths without attributes also raise `NotImplementedError`, Glyphs seems to export them as stroked path.~~ Supported using skia-pathops stroking.


Lib/glyphsLib/builder/builders.py
```diff
@@ -134,7 +133,10 @@ def __init__(
         # Map Glyphs layer IDs to UFO layer names.
         self._layer_map = {}
 
-        # List of exploded color layer when building minimal UFOs.
+        # List of exploded color palette layers when building minimal UFOs.
+        self._color_palette_layers = []
```
anthrotype, line 137: I find the name "color palette layers" a bit confusing. I think you mean here the solid-colors-only kind of CORLv0, how about `self._solid_color_layers`?

Lib/glyphsLib/builder/color_layers.py
```diff
@@ -0,0 +1,151 @@
...
+
+                colorLine = dict(Extend="pad", ColorStop=colorStop)
+
+                gradientType = gradient.get("type")
+                if gradientType is None:
+                    # Linear gradient.
+                    # We set point 2 to point 1 rotated -90 degrees around
+                    # point 0, since Glyphs gradients don’t represent them.
+                    paint = dict(
+                        Format=4,
```
anthrotype, line 108: I suggest importing the PaintFormat enum from fontTools otTables module instead of hard coding the format number.

Lib/glyphsLib/builder/glyph.py
```diff
@@ -63,6 +68,47 @@ def to_ufo_glyph(self, ufo_glyph, layer, glyph):  # noqa: C901
...
+                for color_layer in color_layers:
+                    # XXX: What to do with components? Glyphs does not allow
+                    # setting attributes on them.
+                    assert not color_layer.components
+
+                    groups = {}
+                    for path in color_layer.paths:
+                        # XXX we need a smarter way to group paths with
+                        # identical attributes.
+                        key = json.dumps(path.attributes)
```
anthrotype, line 102: using json to get a string representation of the object is clever, but if it's just to use attributes as key of the groups, you could turn them in tuples of tuples I guess.

Lib/glyphsLib/builder/glyph.py
```diff
@@ -63,6 +68,47 @@ def to_ufo_glyph(self, ufo_glyph, layer, glyph):  # noqa: C901
...
+                and l.associatedMasterId == layer.associatedMasterId
+            ]
+            if color_layers:
+                import json
+
+                layers = []
+                for color_layer in color_layers:
+                    # XXX: What to do with components? Glyphs does not allow
+                    # setting attributes on them.
+                    assert not color_layer.components
```
anthrotype, line 96: if these are valid and possible in source files, you should not just assert but raise a proper exception with error message

anthrotype (commented):
Thanks Khaled for working on this!! I left a few comments.

Could you also summarize what is and what is not supported currently in Glyphs 3 color layers as far as COLRv1?

anthrotype:
> only in minimal mode

I think that's fine for the time being.

anthrotype:
> COLRv1 does not yet support stroking

but our nanoemoji compiler will turn stroked paths into outlines (in picosvg, through skia_pathops). If you need to do the same, you can resort to skia as well.

Lib/glyphsLib/builder/builders.py
khaledhosny: I can see how this can be confusing. The name is derive from the name of Glyphs 3 layer attribute. Color palette layers specify a palette index, when the other color layers set the colors directly on the path.

Lib/glyphsLib/builder/glyph.py
khaledhosny: They can occur in source files. That is a TODO item that I was planing to ask @schriftgestalt about.

Lib/glyphsLib/builder/glyph.py
khaledhosny: The problem is that attributes can contain lists and other tables.

khaledhosny:
> but our nanoemoji compiler will turn stroked paths into outlines (in picosvg, through skia_pathops). If you need to do the same, you can resort to skia as well.

Makes sense. Glyphs also supports setting stroke color and width and we don’t currently support then.

khaledhosny:
> > but our nanoemoji compiler will turn stroked paths into outlines (in picosvg, through skia_pathops). If you need to do the same, you can resort to skia as well.
> 
> Makes sense. Glyphs also supports setting stroke color and width and we don’t currently support then.

I have something that more or less works now, but I’m not sure how to handle the skia-pathops dependency. Should it be a hard-dependency or an optional one?

Lib/glyphsLib/builder/color_layers.py
```diff
@@ -141,6 +143,38 @@ def to_ufo_color_layers(self):
...
+                    import pathops
+                except ImportError as ex:
+                    raise RuntimeError(
+                        f"Stroked color layer path in '{glyph.name}' requires 'pathops' module."
+                    ) from ex
+
+                pos = attributes.get("strokePos", 0)
+                if pos:
+                    raise NotImplementedError(
+                        f"Unsupported color layer path attribute 'strokePos' in '{glyph.name}'"
```
anthrotype, line 161: what does strokePos mean/do?

anthrotype:
you can add skia-pathops to extras_require in setup.py, name the extra "colr"

Lib/glyphsLib/builder/color_layers.py
khaledhosny: It controls on which side of the path the stroke is drawn:
image

khaledhosny: Exporting font with stroke path set to anything other than the default currently crashes Glyphs anyway.

Lib/glyphsLib/builder/glyph.py
khaledhosny: What I really want is to group paths with same attributes together. Using dictionary is probably a hack in the first place, but I don’t know a better way.

anthrotype: have you tried itertools.groupby?

khaledhosny: I didn’t know about it, thanks. Taking a look, I’m not sure it can help as it requires the list to be already sorted, and I wouldn’t know how to sort the attributes.

schriftgestalt: I just keep the previous attributes in a variable and compare the current. You should not collate all layers with the same attributes as there might be something between them.

khaledhosny: Interesting! This would allow me to use itertools.groupby then.

schriftgestalt:
About the components. They do not have attributes on their own. They should just show whatever is on the base glyph. If that base glyph has a color layer, too you should link to it instead of the default layer.

khaledhosny:
Components are handled now (they are mapped to `PaintColrGlyph` paints if the component glyph has color layer, otherwise get decomposed).

I successfully used this PR to build a couple of fonts using Color layers, and I think it is ready.

anthrotype (approved):
Awesome, thanks Khaled!

Issue #752, opened 2021-11-29
codecov is showing stale coverage


khaledhosny:
It seems to not have been updated for over a year.

khaledhosny:
@anthrotype you removed codecov from tox.ini in 55ccc779222c228dd6be0b763877c3d947786a1d, so I assume this was intentional. Should the badge be removed from the README.md as well?

anthrotype:
I actually don't remember. That commit you pointed to was authored by @simoncozens

anthrotype:
Ha! looks like I inadvertently killed codecov support with this https://github.com/googlefonts/glyphsLib/commit/03392eb2053cb76f9659fcb5884bf15bb639e5dc#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03f where I merged the two test+publish workflows into a single one.

@khaledhosny do you mind trying to putting it back?

khaledhosny:
Sorry, I misread the commit author.

I’ll try putting it back.

PR #753, opened 2021-11-29
[ci] Restore Codecov upload

khaledhosny:
Ii was inadvertently removed in 03392eb.

Fixes https://github.com/googlefonts/glyphsLib/issues/752


anthrotype:
thanks!

PR #755, opened 2021-12-10
[cli] Make passing --glyph-data actually do something

khaledhosny:
We were not passing glyph_data to the builder functions.


anthrotype:
oops, thanks! 😅

PR #762, opened 2022-01-06
Don't write disabled feature prefixes

### LINKED ISSUE #761: Support "disabled" feature prefix code
Glyphs allows parts of the feature file to be disabled; glyphs2ufo ignores this setting and copies them to the `features.fea` anyway, with the result that:

```
featurePrefixes = (
{
code = "lookup IDontCare { sub nonsuchglyph by nonsuchglyph; } IDontCare;";
disabled = 1;
name = disabled;
}
);
```

produces a `features.fea`:

```
# Prefix: disabled
lookup IDontCare { sub nonsuchglyph by nonsuchglyph; } IDontCare;
```

When the feature is disabled because glyphs have been removed, this can cause the font to fail to build. (Also it's just *wrong* to include feature code that the designer has asked to be excluded.)

simoncozens:
Fixes #761.

### DIFF
--- a/tests/builder/features_test.py
+++ b/tests/builder/features_test.py
@@ -675,3 +675,22 @@ def test_comments_in_classes(ufo_module):
             ];
 """
     )
+
+
+def test_disabled_prefixes(ufo_module):
+    # https://github.com/googlefonts/glyphsLib/issues/562
+    font = to_glyphs([ufo_module.Font()])
+    feature = classes.GSFeaturePrefix(name="disabledTest")
+    feature.code = "Not even feature code"
+    feature.disabled = True
+    font.featurePrefixes.append(feature)
+
+    (ufo,) = to_ufos(font, ufo_module=ufo_module)
+    assert ufo.features.text == dedent(
+        """\
+        # Prefix: disabledTest
+        # disabled
+        # Not even feature code
+
+    """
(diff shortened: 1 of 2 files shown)


anthrotype:
I suppose that this will not roundtrip when ufo2glyphs, am I right? How important that would be?

simoncozens:
I think we already have that situation with disabled features.

schriftgestalt:
And don’t forget the classes. They can be disabled, too.

tests/builder/features_test.py
```diff
@@ -675,3 +675,22 @@ def test_comments_in_classes(ufo_module):
...
+
+    (ufo,) = to_ufos(font, ufo_module=ufo_module)
+    assert ufo.features.text == dedent(
+        """\
+        # Prefix: disabledTest
+        # disabled
+        # Not even feature code
+
+    """
+    )
```
belluzj, line 696: Could you check that when going back to Glyphs.app the commented-out feature prefix comes back uncommented? = in the same situation before going to UFO (if that sounds like a reasonable thing to have)

anthrotype: @simoncozens ping

khaledhosny:
Superseded by #1027

Issue #746, opened 2021-11-06
Multiaxis v3 Glyphs files fail to generate multiaxis designspace


jpt:
The error message suggests that glyphsLib is only looking at weight and width; the test font contains an optical axis, so the locations are unique

```
glyphs2ufo TestFont.glyphs
DesignSpace sources contain duplicate locations; varLib expects each master to define a unique location.
  ['Test Font Thin 8pt', 'Test Font Bold 8pt', 'Test Font Thin 144pt', 'Test Font Bold 144pt', 'Test Font Thin 8pt Expanded', 'Test Font Bold 8pt Expanded', 'Test Font Thin 144pt Expanded', 'Test Font Bold 144pt Expanded'] => {'Weight': 100, 'Width': 100, 'Custom': 0}
```

jpt:
FWIW, if I resave as a v2 file, it works fine.

I've attached the minimally mastered test font (one glyph - `n`) that I've been using to evaluate glyphsLib. It's reflective of font that properly compiles using Glyphs 3.0.4 build 3100

[TestFont.glyphs.zip](https://github.com/googlefonts/glyphsLib/files/7491263/TestFont.glyphs.zip)

simoncozens:
As of the PR above, glyphs2ufo completes successfully but gives an ambiguous designspace file because of #745. So I am closing this and moving there instead.

PR #766, opened 2022-01-14
Fix two small problems with localised values

simoncozens:
* The key "default" is sometimes used as a default language value
* You can't index into a `dict_values`.

Found while investigating #746

### DIFF
--- a/Lib/glyphsLib/classes.py
+++ b/Lib/glyphsLib/classes.py
@@ -3009,12 +3009,9 @@ class GSFontInfoValue(GSBase):  # Combines localizable/nonlocalizable properties
     def defaultValue(self):
         if not self.localized_values:
             return self.value
-        if "dflt" in self.localized_values:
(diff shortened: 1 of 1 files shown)


Lib/glyphsLib/classes.py
```diff
@@ -3011,9 +3011,11 @@ def defaultValue(self):
             return self.value
         if "dflt" in self.localized_values:
             return self.localized_values["dflt"]
+        if "default" in self.localized_values:
+            return self.localized_values["default"]
```
anthrotype, line 3015: or you could do

```python
for key in ("dflt", "default", "ENG"):
    if key in self.localized_values:
        return self.localized_values[key]
```

anthrotype (approved):

Lib/glyphsLib/classes.py
simoncozens: That does make sense. :-)

PR #769, opened 2022-01-25
Fix brace layers in minimal build

### LINKED ISSUE #741: Intermediate/brace layers are not supported anymore
https://github.com/emmamarichal/league-spartan

`fontmake.errors.FontmakeError: In 'LeagueSpartan.glyphs' -> 'master_ufo/LeagueSpartan.designspace': Generating fonts from Designspace failed: '{500}'`

- font is Glyphs version2
- has intermediate layer at 500 for /e and /a.ss01
- ufo2ft creates a source master at 500 in design space, which has the same name as the origin master at 200, which creates a conflict. 

I don't know at which commit exactly the change occurred though. Maybe it's not even glyphsLib the issue.

```
  <sources>
    <source filename="LeagueSpartan-ExtraLight.ufo" name="League Spartan ExtraLight" familyname="League Spartan" stylename="ExtraLight">
      <lib copy="1"/>
      <groups copy="1"/>
      <features copy="1"/>
      <info copy="1"/>
      <location>
        <dimension name="Weight" xvalue="200"/>
      </location>
    </source>
    <source filename="LeagueSpartan-ExtraLight.ufo" name="League Spartan ExtraLight {500}" layer="{500}">
      <location>
        <dimension name="Weight" xvalue="500"/>
      </location>
    </source>
```

[sources.zip](https://github.com/googlefonts/glyphsLib/files/7425902/sources.zip)

simoncozens:
Fixes #741

### DIFF
--- a/Lib/glyphsLib/builder/builders.py
+++ b/Lib/glyphsLib/builder/builders.py
@@ -326,7 +326,11 @@ class UFOBuilder(_LoggerMixin):
                 and ".background" not in layer.name
             ):
                 self.bracket_layers.append(layer)
-            elif self.minimal and layer.layerId not in master_layer_ids:
+            elif (
+                self.minimal
+                and layer.layerId not in master_layer_ids
+                and not layer._is_brace_layer()
+            ):
                 continue
             else:
                 ufo_layer = self.to_ufo_layer(glyph, layer)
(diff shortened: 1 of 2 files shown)


anthrotype (approved):
LGTM, thanks for the quick fix!

anthrotype:
I suggest we make a glyphsLib release after this, then also update fontmake's setup.py to require the new glyphsLib and release a new fontmake as well with it. Could you also take care of that?

Issue #770, opened 2022-01-31
Some glyph names get a different PostScript name in Glyphs.app


madig:
Edit: the following glyph names are not handled by glyphsLib at all, i.e. their PostScript name is `None`:

```
{'Ech_Vew-arm.liga',
 'Men_Ech-arm.liga',
 'Men_Ini-arm.liga',
 'Men_Now-arm.liga',
 'Men_Xeh-arm.liga',
 'Vew_Now-arm.liga',
 'aiMatra_anusvara-deva',
 'aiMatra_candraBindu-deva',
 'aiMatra_reph-deva',
 'aiMatra_reph_anusvara-deva',
 'ca_iMatra-tamil',
 'ca_uMatra-tamil',
 'ca_uuMatra-tamil',
 'ch_ya-deva',
 'd_dh_ya-deva',
 'da_rVocalicMatra-deva',
 'da_uMatra-deva',
 'da_uuMatra-deva',
 'dd_dda-deva',
 'dd_ddha-deva',
 'dd_ya-deva',
 'ddh_ddha-deva',
 'ddh_ya-deva',
 'eCandraMatra_anusvara-deva',
 'eCandraMatra_reph-deva',
 'eMatra_anusvara-deva',
 'eMatra_candraBindu-deva',
 'eMatra_reph-deva',
 'eMatra_reph_anusvara-deva',
 'eShortMatra_anusvara-deva',
 'eShortMatra_candraBindu-deva',
 'eShortMatra_reph-deva',
 'eShortMatra_reph_anusvara-deva',
 'ech_vew-arm.liga.sc',
 'finalkaf_qamats-hb',
 'finalkaf_sheva-hb',
 'finalkafdagesh_qamats-hb',
 'finalkafdagesh_sheva-hb',
 'fongman_fannu-thai',
 'h_ra_uMatra-deva',
 'h_ra_uuMatra-deva',
 'ha_iMatra-tamil',
 'ha_iiMatra-tamil',
 'ha_rVocalicMatra-deva',
 'ha_rrVocalicMatra-deva',
 'ha_uMatra-deva',
 'ha_uMatra-tamil',
 'ha_uuMatra-deva',
 'ha_uuMatra-tamil',
 'hatafpatah_siluqleft-hb',
 'hatafqamats_siluqleft-hb',
 'hatafsegol_siluqleft-hb',
 'headline-gurmukhi.shorter',
 'iMark_toandakhiat-khmer',
[...]

PR #771, opened 2022-01-31
[glyphdata] handle production names for ligatures with script suffixes

khaledhosny:
We had them handled in category detection but not in production names.
This change generalizes this and adds few more conditions:

1. Don’t add script suffix to ligature part that already have one.
2. Don’t add it if the part had an entry in GlyphData (to handle cases like "brevecomb_Dboldscript-math").

Should fix https://github.com/googlefonts/glyphsLib/issues/770


khaledhosny:
@madig feel free to add more tests. This probably does not fix all the differences you are seeing.

khaledhosny:
The regression tests failure is because we now detect proper names for more ligatures.

madig:
That was fast. Thanks! A ripgrep and sort through the diffs shows all new names are uni-names now, so that's good.

I added more testcases and found several diffs though. Not entirely sure if some of it is due to outdated info in GlyphData.xml? I used Glyphs 2.6.6 to generated the expected values.

khaledhosny:
With ef6fed2 the failures are down to 16 (was 70).

khaledhosny:
The last commit should have fixed a punch of failures, but it does not. We now generate uni names for the affected glyphs, but they are different from Glyphs’.

khaledhosny:
Remaining failures:
```
FAILED tests/glyphdata_test.py::test_prod_names[sh_r-deva-uni0936094D0930094D] - AssertionError: assert 'uni0936094D094D0930' == 'uni0936094D0930094D'
FAILED tests/glyphdata_test.py::test_prod_names[idotaccent.sc-i.sc.loclTRK] - AssertionError: assert 'i.loclTRK.sc' == 'i.sc.loclTRK'
FAILED tests/glyphdata_test.py::test_prod_names[t_r-deva-uni0924094D0930094D] - AssertionError: assert 'uni0924094D094D0930' == 'uni0924094D0930094D'
```

I think we should just accept the `idotaccent.sc` one as insignificant difference, but I have no idea what is causing the difference in the other two; we are splitting the ligatures correctly and getting the correct part names from GlyphData and I have no idea from where Glyphs is pulling the difference (AFAICT the relavant entries are the same in our copy of the data and Glyphs’).

anthrotype:
what's the status of this PR?

schriftgestalt:
I’m reworking the glyph data handling quite a bit right now so that is probably not needed any more.

anthrotype:
> I’m reworking the glyph data handling quite a bit right now so that is probably not needed any more.

how's that going?

khaledhosny:
I updated the expectations of the three remaining test failures. These should be investigated, but the failures are unrelated to splitting ligatures, we are doing this correctly but Glyphs is doing something more here.

If there are no objections, I’m going to merge this.

khaledhosny:
(As said previously, the regression failures are this PR working and handling more ligatures than before).

anthrotype:
maybe instead of making the test pass you could mark them as `xfail`?

tests/glyphdata_test.py
```diff
@@ -190,5 +193,221 @@ def test_glyphdata_no_duplicates(self):
...
+    "ro-khmer.pre.narrow": "uni17D2179A.narrow",
+    "rra_iMatra-tamil": "uni0BB10BBF",
+    "rra_iiMatra-tamil": "uni0BB10BC0",
+    "rra_uMatra-tamil": "uni0BB10BC1",
+    "rra_uuMatra-tamil": "uni0BB10BC2",
+    "sa_iMatra-tamil": "uni0BB80BBF",
+    "sa_iiMatra-tamil": "uni0BB80BC0",
+    "sa_uMatra-tamil": "uni0BB80BC1",
+    "sa_uuMatra-tamil": "uni0BB80BC2",
+    "sh_r-deva": "uni0936094D094D0930",  # uni0936094D0930094D
```
anthrotype, line 372: i'm not familiar with devanagari conjucts but looks like there's some reordering that is (or is not) happening here?
/cc @schriftgestalt 

```diff
- uni0936094D094D0930
+ uni0936094D0930094D
```

anthrotype:
+1 to merge this as improves current handling of ligatures' production names; as for any outstanding issues left, we could track them in their own issue and tackle later

schriftgestalt:
That should be handle with my pull request that was never excepted.

khaledhosny:
> That should be handle with my pull request that was never excepted.

Which one, https://github.com/googlefonts/glyphsLib/pull/818?

If you want to finish it in some reasonable time frame, I’m more than happy to close this one instead, otherwise I’m currently blocked on this issue and need a fix ASAP.

schriftgestalt:
It was mergable when opened the pull request.

schriftgestalt:
My PR fixes and improves several areas (names, indo, writing direction …). Adding more changes now will make it even more complicated to merge it. 
And that code is needed also for round tripping Glyphs 2 and 3 files.

khaledhosny:
So what is the resolution here? #818 is unmergable as it stands and I’m still blocked.

schriftgestalt:
I'll panned to work on glyphsLib next week. Finishing the Glyphs3 branch and rebasing the info branch.

khaledhosny:
I hate to keep saying this, but this PR have been open for over a tear and half and it is been 3 weeks since I rebased it, and I’m still blocked. I suggest merging this and I’ll try to help with rebasing #818 and solving the conflicts afterwords.

tests/glyphdata_test.py
khaledhosny: I think this is addressed by the mini Devanagri shaper in #818, so I’m not going to worry about it here.

anthrotype:
Feel free to merge this as soon as you make the CI pass, thanks

khaledhosny:
BTW, I tried to rebase #818 on current master, and there are significant conflicts, so I don’t think merging this is going to make it much worse.

Issue #773, opened 2022-02-10
Width of intermediate layer written to UFO is wrong


yanone:
The master layer’s width gets written instead. Outlines and anchors are okay, tho.

Please see attached minimal font. The file `OodleSans.Regular.ufo/glyphs.10. F_eb. 22, 08_39/hah-ar.init.swsh.glif` has width `1545` (same as its master layer) but should be `1045`.

I tried digging into the code but couldn't find it. I used the normal `fontmake -g sources/IntermediateLayerWidth.glyphs -o ufo` to convert.

[IntermediateLayerWidth.glyphs.zip](https://github.com/googlefonts/glyphsLib/files/8039131/IntermediateLayerWidth.glyphs.zip)

anthrotype:
somewhere here must be the hiding the bug:

https://github.com/googlefonts/glyphsLib/blob/57e66d21b71676a4e05b6ab02cb6204aad0ef81a/Lib/glyphsLib/builder/glyph.py#L254-L278

anthrotype:
regression probably introduced with this https://github.com/googlefonts/glyphsLib/pull/641

I think line 267 of glyphsLib/builder/glyph.py should be:

```diff
-      width = None 
+      width = layer.width 
```

@yanone can you try if that fixes it and send a PR if so? thanks

yanone:
I fixed it (see linked PR). The local tests run fine, only the regression test throws an error. Please help me make sense of it. I almost feel like thats expected now, because fonts have shipped with wrong layer widths since April 2021 now.

PR #774, opened 2022-02-10
Retain intermediate layer width correctly

yanone:
For https://github.com/googlefonts/glyphsLib/issues/773

```
# The width may be taken from another master via the customParameters 
# 'Link Metrics With Master' or 'Link Metrics With First Master'.
```

It says `may`, but then the metrics are copied regardless. I added a condition.

### DIFF
--- a/Lib/glyphsLib/builder/glyph.py
+++ b/Lib/glyphsLib/builder/glyph.py
@@ -254,7 +254,10 @@ def to_ufo_glyph(self, ufo_glyph, layer, glyph, do_color_layers=True):  # noqa: 
     # The width may be taken from another master via the customParameters
     # 'Link Metrics With Master' or 'Link Metrics With First Master'.
     master = self.font.masters[layer.associatedMasterId or layer.layerId]
-    if "Link Metrics With Master" in master.customParameters or "Link Metrics With First Master" in master.customParameters:
+    if (
+        "Link Metrics With Master" in master.customParameters
+        or "Link Metrics With First Master" in master.customParameters
+    ):
         metric_source = master.metricsSource.id
         metric_layer = self.font.glyphs[glyph.name].layers[metric_source]
         if metric_layer:
--- a/Lib/glyphsLib/builder/glyph.py
+++ b/Lib/glyphsLib/builder/glyph.py
@@ -254,17 +254,20 @@ def to_ufo_glyph(self, ufo_glyph, layer, glyph, do_color_layers=True):  # noqa: 
     # The width may be taken from another master via the customParameters
     # 'Link Metrics With Master' or 'Link Metrics With First Master'.
     master = self.font.masters[layer.associatedMasterId or layer.layerId]
-    metric_source = master.metricsSource.id
-    metric_layer = self.font.glyphs[glyph.name].layers[metric_source]
-    if metric_layer:
-        width = metric_layer.width
-        if layer.width != width:
-            logger.debug(
-                f"{layer.parent.name}: Applying width from master "
-                f"'{metric_source}': {layer.width} -> {width}"
-            )
+    if "Link Metrics With Master" in master.customParameters or "Link Metrics With First Master" in master.customParameters:
+        metric_source = master.metricsSource.id
+        metric_layer = self.font.glyphs[glyph.name].layers[metric_source]
+        if metric_layer:
+            width = metric_layer.width
+            if layer.width != width:
+                logger.warning(
+                    f"{layer.parent.name}: Applying width from master "
+                    f"'{metric_source}': {layer.width} -> {width}"
+                )
+        else:
+            width = layer.width
     else:
-        width = None
+        width = layer.width
 
     if width is None:
         pass
--- a/Lib/glyphsLib/builder/glyph.py
+++ b/Lib/glyphsLib/builder/glyph.py
@@ -254,23 +254,20 @@ def to_ufo_glyph(self, ufo_glyph, layer, glyph, do_color_layers=True):  # noqa: 
     # The width may be taken from another master via the customParameters
     # 'Link Metrics With Master' or 'Link Metrics With First Master'.
     master = self.font.masters[layer.associatedMasterId or layer.layerId]
-    if (
-        "Link Metrics With Master" in master.customParameters
-        or "Link Metrics With First Master" in master.customParameters
-    ):
-        metric_source = master.metricsSource.id
-        metric_layer = self.font.glyphs[glyph.name].layers[metric_source]
+    metrics_source = master.metricsSource
+    if metrics_source is None:
+        width = layer.width
+    else:
+        metric_layer = self.font.glyphs[glyph.name].layers[metrics_source.id]
         if metric_layer:
             width = metric_layer.width
             if layer.width != width:
                 logger.debug(
                     f"{layer.parent.name}: Applying width from master "
-                    f"'{metric_source}': {layer.width} -> {width}"
+                    f"'{metrics_source.id}': {layer.width} -> {width}"
                 )
         else:
-            width = layer.width
-    else:
-        width = layer.width
+            width = None
 
     if width is None:
         pass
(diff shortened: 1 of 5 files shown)


anthrotype:
Maybe @simoncozens or @jenskutilek can take a look as they last touched this code

Lib/glyphsLib/builder/glyph.py
```diff
@@ -254,17 +254,23 @@ def to_ufo_glyph(self, ufo_glyph, layer, glyph, do_color_layers=True):  # noqa:
...
-        width = metric_layer.width
-        if layer.width != width:
-            logger.debug(
-                f"{layer.parent.name}: Applying width from master "
-                f"'{metric_source}': {layer.width} -> {width}"
-            )
+    if (
+        "Link Metrics With Master" in master.customParameters
+        or "Link Metrics With First Master" in master.customParameters
+    ):
```
jenskutilek, line 260: > ```
> # The width may be taken from another master via the customParameters 
> # 'Link Metrics With Master' or 'Link Metrics With First Master'.
> ```
> 
> It says `may`, but then the metrics are copied regardless. I added a condition.

There already is a condition in https://github.com/googlefonts/glyphsLib/blob/main/Lib/glyphsLib/classes.py#L1592-L1616, so you're double checking the existence of the custom parameter now.

jenskutilek:
Nope, intermediate layers can't be kerned independently in Glyphs. I removed the related comment.

jenskutilek:
This is good to be merged as far as I can see, as the PR it depended on has been merged.

jenskutilek:
Don’t know what’s up with the regression tests, maybe they are valid changes that are caused by the fix in layer width handling, but then why didn’t they flag the changes in #777?

yanone:
I can explain. All except the bottom few glyphs are _brace layers_ as denoted by their `glyph.*` names, `*` being the layer name. Those are the actual targeted by this PR, so they're expected to change.

What I didn't expect to change were the substitution glyphs/bracket layers. I checked the example shown at the bottom of the regression test (Hahmlet.glyphs) manually to the result that earlier, `glyphsLib` indeed propagated the original layer’s width to the substitute layers/glyphs, which is of course nonsense. So this has been fixed in the same go.

It would be nice if you could independently look at Hahmlet again and confirm that the width propagation of `guarani` is now correct and formerly wasn't.

Lib/glyphsLib/classes.py
```diff
@@ -1602,7 +1602,7 @@ def metricsSource(self):
 
         # No custom parameters apply, go home
         if not source_master_id:
-            return self
+            return None
```
anthrotype, line 1605: the docstring above says "Normally this is the current master itself", hence this would `return self`, but now you changed it to None. `glyphsLib.classes` module is supposed to mimic the Glyphs.app's own Python Scripting API. Is there a corresponding `metricsSource` property in the latter, what is it supposed to return in this case? We should try to match that. I see you're using the `None` return value as a signal that there is no linked metrics master, maybe you could test `master.metricsSource is master`?

jenskutilek: There is no `GSMaster.metricsSource` property in the Glyphs API, it was added as a helper in #641 by @simoncozens.

jenskutilek: > I even just noticed something else: if source_master_id in self.font.masters: is comparing IDs against master objects, so that's nonsense.

Good catch, @yanone … I think we need to check all the master ids, like we already do for the names:

```python
    @property
    def metricsSource(self):
        # [...]

        # No custom parameters apply, go home
        if not source_master_id:
            return None

        # Try by master id
        for source_master in self.font.masters:
            if source_master.id == source_master_id:
                return source_master

        # Try by name
        for source_master in self.font.masters:
            if source_master.name == source_master_id:
                return source_master
```

jenskutilek: No, we can do it easier:

```python
        # Try by master id
        source_master = self.font.masterForId(source_master_id)
        if source_master is not None:
            return source_master

        # Try by name ...
```

jenskutilek: I guess this is confusing because the custom parameter doesn’t return the master ID, but the master name … so is the check by master ID needed at all?

yanone: Regarding @anthrotype’s inquiry about the return value: The `.metricsSource` attribute is only ever used once, here:https://github.com/googlefonts/glyphsLib/blob/1903622df5d31d056f4a7a2186d27a4d643f62bb/Lib/glyphsLib/builder/glyph.py#L257-L259

Since the value is checked for `None`, I conclude that returning `None` is the correct answer.

Otherwise I’ve added a test that checks the width of brace and bracket layers.
Sadly, nowadays, the Actions-based tests don't complete because of Codecov. I'm guessing it's not sending the coverage because this is a PR, which makes sense. So 2 out of 4 test are cancelled subsequently.

The test itself passes, though. Please have a look at the test routines and decide how to move forward.

jenskutilek:
> It would be nice if you could independently look at Hahmlet again and confirm that the width propagation of `guarani` is now correct and formerly wasn't.

I’ve checked Hahmlet, and the guarani layer widths are handled correctly now.

madig:
@yanone I reran then, now they're fine. Codecov is known to be flaky.

anthrotype:
The regression tests will need to be updated to match the changes.

yanone:
Please give me a hint on how to do that. I was under the impression that the print output of one test run of one code version is compared against the other and that the new output simply needs to be checked by a human and accepted.

Lib/glyphsLib/classes.py
anthrotype: ok, i'm fine with this returning `None`; can you then please also update the docstring which is still saying "Normally this is the current master itself..."

madig:
Inside your development venv (make sure you `pip install -e .` so you're actually running the code you're hacking on), run `python tests/tools/generate_regression_test_files.py` and commit all diffs it generates.

Lib/glyphsLib/builder/kerning.py
```diff
@@ -22,11 +22,12 @@
 
 def to_ufo_kerning(self):
     for master in self.font.masters:
-        master_id = master.id
-        kerning_source = master.metricsSource.id  # Maybe be a linked master
-        if kerning_source in self.font.kerning:
-            kerning = self.font.kerning[kerning_source]
-            _to_ufo_kerning(self, self._sources[master_id].font, kerning)
+        kerning_source = master.metricsSource  # Maybe be a linked master
```
anthrotype, line 25: you could do 
```python
        kerning_source = master.metricsSource or master  # Maybe be a linked master
```

and remove the following `if kerning_source is None:`

anthrotype (approved):
LGTM (provided that the regression tests gets updated)

Thanks @yanone for taking care of this!

anthrotype:
also thanks @jenskutilek of course!

yanone:
I need help please that's maybe not related to glyphsLib.

I ran `generate_regression_test_files.py` seemingly successfully, but the changes (if any) don't show up in my git (I use Github Desktop). Checking the data folder in VSCode shows them grayed-out, which to my knowledge hints at them being excluded from git, but I found no such evidence in `.gitignore`. Anyone have an idea what I’m looking at?

Bildschirmfoto 2022-05-04 um 10 47 54

anthrotype:
ah, so ok. these regression tests compare the glyphsLib from `main` branch against the results from the given PR branch and fail if there's any diff. The expected test data is re-generated each time (see https://github.com/googlefonts/glyphsLib/blob/main/.github/workflows/regression.yml) using the glyphsLib from `main` branch, so it doesn't need to be committed. Yes, it's weird it doesn't seem to be explicitly .gitignore'd, I don't know why it doesn't show up in the changed or untracked files.
Anyway, in this case, the diff is a good one, so we want to merge regardless of the regression test failures.

yanone:
Check. Thank you

PR #775, opened 2022-02-10
Accept Glyph.app’s ligature mark names

yanone:
Recognizes `shaddaFatha-ar` (Glyphs.app’s default naming) next to `shadda_fatha-ar` as a ligature mark and thus enables proper anchor propagation

### DIFF
--- a/Lib/glyphsLib/builder/anchors.py
+++ b/Lib/glyphsLib/builder/anchors.py
@@ -16,6 +16,7 @@
 from fontTools.misc.transform import Transform
 from .constants import COMPONENT_INFO_KEY
 import fontTools.pens.boundsPen
+import re
 
 from glyphsLib.types import Point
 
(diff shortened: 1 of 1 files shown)


yanone:
So, camelCase _can_ be a ligature.

yanone:
This is how I think it works: 

You're looking to find the final positions for a specific anchor, say `top`. You walk down a glyph’s components stack in reverse order (defined in the glyph in this order: base components first, mark components on top), and if any of the components has the anchor, you apply all components transformations and anchor positions from the first component to where you found the anchor.

Is this what you were asking for?

schriftgestalt:
> so any glyphs that uses camelCase would count as a ligature (e.g. "fontTools" would be a ligature of "font" and "tools"?)... That's surprising to me. Maybe we should summon @schriftgestalt to give his authoritative answer.

No. Ligatures are only glyph that might need a caret position and are made from more than one base "letter". And those are always (with a few exceptions that are explicitly marked in GlyphData, e.g.: allah-ar) written with underscores. Adieresis, shaddaFatha-ar and glyphs like that, that contain two or more components might be called Precomposed or Composites. 
The camelCase is used to improve readably. That is not very consistently used, I know. So it has no meaning. [^1]

The difference concerning anchor propagation is this:

- Composites: The components build a chain. Take the "_"-anchors from the first and the non underscore anchors from the last. So for shadda_fatha-ar (if build from components): use the "_top" anchor from the shadda and the "top" from fatha. 
- Ligatures: for each base glyph a set of "top_x" and "bottom_x" anchor is extracted. So for seen_reh-ar, you get the "top" and "bottom" from the seen, and because it is the first base, add a "_1". the same for the reh but add a "_2". It becomes a bit tricky if you have nested ligature parts: if you use that seen_reh in a "beh_seen_reh-ar". then you get the "top_1" form the beh and you need to rename the "top_1" and "top_2" form the seen_reh to "top_2" and "top_3". And then you can have marks interleaved. Then the rues from the Composites apply (use the position from the mark instead of the one from the base). 

Is that understandable? 

[^1]: And the shaddaFatha-ar is an old exception. Its existence is mostly there to reorder the parts (AFAIK, the trying order is not defied but the fatha should always be above) and position the parts. So it could be replace with some reordering lookup and mark positioning.

PR #777, opened 2022-02-17
Fix layer width handling when the width should not be taken from a master

jenskutilek:


anthrotype:
github says this branch has conflicts with main, please rebase on main, thanks

jenskutilek:
I wanted to merge to @yanone's branch first, that's why I based this on his branch. I resolved the conflict.

anthrotype:
ah sorry I missed that. So this PR goes into Yanone's branch, not main?

yanone (approved):
Jens’ solution works for my case, it adjusts the layer width as expected.
I'm not sure how to proceed: Is this PR getting merged and I close mine?

yanone:
I’m not the brightest with branches. It appears that it's based on `googlefonts` rather than my fork.

jenskutilek:
No, my branch is based on your branch in this `googlefonts/glyphsLib` repo. It should be merged into your branch, then your branch into main.

(It may be somewhat convoluted because I basically redid all your changes, but I wanted to avoid "invalidating" or ignoring your original pull request)

yanone:
I need clear instructions from an adult about what to do with this PR please.

yanone:
@madig Asking you here because I have no email address of yours. Cosimo suggested I ask you about two errors in the Github Action for this PR.

Black throws a traceback about an import, which is clearly a bug in the code. Can another Black version be used here?
What version is used? I could send them a bug report, but Black doesn't show up in the installation logs of that Action run.

And Flake8 throws two B020 errors about iterables being changed inside the loop, which I looked up in the code and can't confirm.

Any ideas?

https://github.com/googlefonts/glyphsLib/runs/6094266252?check_suite_focus=true

Also, a side question: Is there a way to find the respective Action run from inside this PR page? There is no link to it anywhere. Only the `cla` check shows up above.

yanone:
ps: When I manually run the latest flake8 on my machine on the files in question, the B020 doesn't show up, actually. A hint at an outdated version on the Action?

madig:
Outdated versions may be at fault here, I'll have a look.

madig:
I rejigged glyphsLib's linting process in https://github.com/googlefonts/glyphsLib/pull/784.

madig:
@yanone oh, regarding running actions: they might vanish once a PR is merged, not sure. Otherwise you can select each job separately.

yanone:
Thank you

PR #778, opened 2022-02-21
Support Glyphs3 RTL kerning

yanone:
Glyphs 3 kerning is sorted into the new `kerningLTR` and `kerningRTL` font data attributes instead of the previous plain `kerning`. The kerning for RTL is now stored in opposite attributes, contradicting the first/second logic of UFO and G2. `@MMK_L_*` becomes `@MMK_R_*` (and vv), and the visually right glyph side’s kerning class is now stored in `kernRight` as opposed to the previous `kernLeft` (=first/incoming side).

This means that correctly sorting glyphs into first and second UFO groups now involves checking a glyph’s script and whether or not it’s a number (that are always LTR even for RTL scripts). 

This in turn required a fix for the glyph data generator that previously didn't generate the script for unencoded derivative glyphs such as `hah-ar.init.swsh`. The algorithm then cycles through the glyph name variations by cutting off dot-separated suffices from the right until glyph data could be found (`hah-ar.init.swsh->hah-ar.init`), and adds only the script value, as the other values were found to be already correctly generated.


yanone:
One sec, I’m gonna rewrite the _script_ construction for the glyph data to be more aligned with the methods already in place for constructing the _production_ name and _categories_.

yanone:
Okay, I’m done. The script construction is now analog to the category construction, and also covers ligatures as requested in the `#TODO` comment.
The failure in the Windows tests I believe is unrelated to this PR.

schriftgestalt:
My comment was more general, not about your changes. I should make a separate issue.

Lib/glyphsLib/builder/groups.py
```diff
@@ -140,12 +140,28 @@ def _to_glyphs_kerning_group(self, name, glyphs):
...
+    group's side (1 or 2) and writing script.
     """
-    if int(side) == 1:
-        return "rightKerningGroup"
-    else:
-        return "leftKerningGroup"
+    if glyph.parent.format_version > 2:  # Glyphs 3
+        glyph_data = glyphdata.get_glyph(glyph.name)
+        if (
+            glyph_data.script in ("arabic", "hebrew")
```
anthrotype, line 148: why only "arabic" and "hebrew"?

yanone: These are the two RTL scripts to my knowledge. This point is arguably the critical one, also about the numbers. I chose those to my best knowledge. Some algorithm needs to be in place to sort Glyphs’ kerning classes into the correct first/second pairs, and I didn’t know any better.

yanone: I see, there are more RTL scripts. Are those already defined somewhere in the toolchain for fonts?

simoncozens: You can get script direction from a function in ufo2ft.utils.

(I’m on the road this week so won’t be able to review.)

yanone: > I can add the writing direction to the glyphdata.xml but there might be a way to query unicode to get the writing direction of each code. Then we don’t need to check for `script == arabic and category != Number`.

Yes, I could try this approach. @anthrotype Seems like a good idea?

khaledhosny:
This is getting dangerously convoluted. The designer manually determines which kerning is RTL and which is LTR, then glyphsLib turns this into a single kerning table, then ufo2ft’s kern feature writer will try to deduce which kerning is RTL and which is LTR.

khaledhosny:
Also, some of the work being done here conflicts with #771.

Lib/glyphsLib/builder/kerning.py
```diff
@@ -24,25 +24,34 @@ def to_ufo_kerning(self):
...
     """Add .glyphs kerning to an UFO."""
 
     warning_msg = "Non-existent glyph class %s found in kerning rules."
 
     for left, pairs in kerning_data.items():
-        match = re.match(r"@MMK_L_(.+)", left)
+        if direction == "LTR":
+            match = re.match(r"@MMK_L_(.+)", left)
+        elif direction == "RTL":
+            match = re.match(r"@MMK_R_(.+)", left)
```
anthrotype, line 44: wait.. does Glyphs 3 really calls the class on the _left_-hand side `@MMK_R_` and the one on the _right_-hand side `@MMK_L_`?!

yanone:
> This is getting dangerously convoluted. The designer manually determines which kerning is RTL and which is LTR, then glyphsLib turns this into a single kerning table, then ufo2ft’s kern feature writer will try to deduce which kerning is RTL and which is LTR.

We’re hoping for a remedy with UFO 4 to support separate kerning tables for each writing direction, if I understand correctly? Today, G3 sources don’t produce any Arabic kerning at all, so there is an urgency.

> Also, some of the work being done here conflicts with https://github.com/googlefonts/glyphsLib/pull/771.

They way I see it doesn't conflict with my issue logically, so I could easily wait it out and apply my changes to the updated code after #771. Is this PR under active development?

Lib/glyphsLib/builder/kerning.py
yanone: I’ll try to answer: The kerning pairs are still defined in logical order. So the code logic doesn’t change here. First stays first, second stays second. Additionally, for the new RTL-specific kerning table, Georg has switched the `L` and `R` to now be visual rather than logical, (as he did with `kernLeft` and `kernRight` classes) which I believe shouldn't have happened. We should have silently accepted that `L` and `R` were never appropriate names but are used to describe 1st and 2nd.

FYI, you can check the two added `RTL_kerning_v2.glyphs` and `RTL_kerning_v3.glyphs` files for reference, in particular the kerning pair `reh-ar,hah-ar.init.swsh` which is defined in both.

yanone:
My personal take, although it belongs more in the UFO 4 thread: I’m in favour of having separate kerning tables for all writing directions, as this removes all ambiguity of sorting. Yes, the user decides that by switching the writing direction in Glyphs. So yes, this either needs to be carried all the way through to ufo2ft, or as a workaround, maybe we can use ufo2ft’s exact same algorithm to decide on the writing direction in this PR, which would mean to leave all other code changes of mine in place and only replace the sorting bit.

But please, in gods name, keep the data definitions as first and second always, including user-defined kerning classes. They should be called incoming and outgoing side rather than visual left and right.

schriftgestalt:
> Additionally, for the new RTL-specific kerning table, Georg has switched the L and R to now be visual rather than logical, (as he did with kernLeft and kernRight classes) which I believe shouldn't have happened.

That is probably try. I can explain why I did that. If you open a font in Glyphs 2 and switch from LTR to RTL, the kerning classes would switch in the UI. That can lead to very unexpected results when you accidentally edit the classes in the wrong mode (or in font view). So I decided to actually use the visual order as the keys where suggesting all along.

yanone:
What would help here is if editors employed a correct text composition engine so that a user wouldn’t be required to choose a direction for their kerning pairs, and the editor automatically sorted the kerning by the writing direction as calculated by the composition engine.

simoncozens:
> What would help here is if editors employed a correct text composition engine so that a user wouldn’t be required to choose a direction for their kerning pairs, and the editor automatically sorted the kerning by the writing direction as calculated by the composition engine.

This is true: and if you go down that road, the editor doesn’t need to store a direction for the kerning pair in the font source because this direction is computed when the font is compiled. kerningLTR / kerningRTL is just an overcomplication.

schriftgestalt:
We are thinking about adding proper BIDI support to the edit view. But I fear the navigation will not get any easier if we do so. Dealing with mixed text is a nightmare everywhere I encounter it (try adding something to the beginning or end of the Arabic part in cases like this: `key = "جبأ"`. I copy this to Glyphs and do the edit there because of the missing BIDI processing).

But I get the point of the classes. I can change that back for the next version.

yanone:
We can close this. Naming mark ligatures with an underscore is a good habit, and sources can be changed accordingly.

anthrotype:
@yanone I believe you meant to close https://github.com/googlefonts/glyphsLib/pull/775 instead of this one

yanone:
Oh sorry. Reopening

yanone:
> I can change that back for the next version.

Well, it's too late now. 

Does anyone have an idea on how to move forward? I see three options:

1) Not support RTL kerning for G3 sources (further complicated by the fact that G3 can't save backward-compatible G2 files)
2) Accept the changes more or less as-is
3) Propose another solution

The only way I can see this both work & not rely on my weak _Arabic/Hebrew & not a number_ condition is to employ the same algorithm that `ufo2ft` uses to sort the kerning pairs into writing directions. If that's interesting, I would need some advice on this. The way I see it, `ufo2ft` uses a glyph’s Unicode value to determine writing direction, which is insufficient for unencoded glyphs such as many glyphs that are present in my own Arabic font I’m currently working. The approach in this PR relies on GlyphInfo which I’ve fixed to report the correct script for all types of unencoded glyphs.

But I have no solution for how to work around the awkward `L`/`R` switches other than what I’ve already done.

anthrotype:
Ufo2ft doesn't simply use unicode codepoints but compiles a temporary GSUB table and runs the fontTools.subset glyph "closure" over GSUB to find all the glyphs that are reachable from an initial set of glyphs whose codepoints is associated with a given BIDI type (strong LTR or RTL, etc.)

anthrotype:
I am not suggesting that you use the same approach here, I was simply clarifying the ufo2ft method. I haven't fully thought this through and right now can't advice on a way forward I'm afraid.

schriftgestalt:
The writing direction for a glyph should be derived the same way as the other glyph info properties. As the base info objects mostly have a Unicode, you can use the Unicode properties to get the writing direction, or we add the direction to glyph data. I have that in Glyphs so I could export the glyph data with this information.

yanone:
I have a font on my table now for Google Fonts that is a COLRv1 Arabic font so it's both 1) not backwards compatible to Glyphs2 and 2) needs RTL kerning to work. So I wanted to ask if this issue could be revisited?

tests/data/RTL_kerning_v3.glyphs
```diff
@@ -0,0 +1,241 @@
...
+);
+instanceInterpolations = {
+m01 = 1;
+};
+name = Regular;
+}
+);
+kerningRTL = {
+m01 = {
+"@MMK_R_leftAlef" = {
```
anthrotype, line 202: is it right ("MMK_R"), is it left ("leftAlef"), first, second?! My mind just exploded 🤯

anthrotype:
> As the base info objects mostly have a Unicode, you can use the Unicode properties to get the writing direction, or we add the direction to glyph data. I have that in Glyphs so I could export the glyph data with this information.

@schriftgestalt what did you decide to do in the end about this?

schriftgestalt:
As I said, I’m reworking the glyphData and I can add the direction data form Glyphs, too.

anthrotype:
thanks, that would help I guess. Though I'm still very much confused as we can't even agree on what is "left" or "right".

anthrotype:
Can I find a blogpost, handbook or help page, whatever that explains what exactly changed from Glyphs 2 to 3 as regards RTL kerning?
I'm also available to arrange a video call if anyone volunteers to teach me.

tests/data/RTL_kerning_v3.glyphs
anthrotype: perhaps I get it now, please confirm. In the Glyphs3's `kerningRTL` dictionary, if the _first_ key is a group it's prefixed `@MMK_R_`, whereas the second key in the (nested) dict, is prefixed `@MMK_L_` if a group. Hence, in RTL kerning, the (logical) first glyph is the one placed on the right-hand side, the second glyph is the one on the left-hand side. The opposite is true for LTR kerning, where first is left and second is right. All good so far.

Now, the meaning of "left" and "right" in the `leftKerningGroup` and `rightKerningGroup` glyph properties is _not_ "the group this glyph belong when it occurs in the left|right position", but rather the _side_ or _direction_ towards which this glyph is kerned against: so, for RTL kerning (like your `leftAlef` above), "left" means the group pertains to the left (in visual sense) side of the alef glyph, when this is placed in the first (logical) position next to second glyph, whose right (visual) side is the one that participate in the kerning...

schriftgestalt:
I have some time tomorrow morning from 9–12 my time.

tests/data/RTL_kerning_v3.glyphs
schriftgestalt: There are several stages where the kerning data is accessed. For me the more important part is during editing the font data in the UI, not the generating of the feature code. 
I think the confusion comes form the fact that the classes are defined on a single glyph. The classes (left/right/top/bottom) collects all glyphs that are similar on that side. The kerning is defined on pairs, so here the left/right mean the glyph on that side of the pair. The right side of the left glyph is kerned to the left side of the right glyph. I know that this can be confusing but this is how I managed to get it together.

Just for reference a comparison between LTR and RTL:
```
kerningLTR = {
	m01 = {
		"@MMK_L_A_R" = {
			"@MMK_R_B_L" = -20;
		};
	};
};
kerningRTL = {
	m01 = {
		"@MMK_R_alef_L" = {
			"@MMK_L_beh_R" = -50;
		};
	};
};
```

Lib/glyphsLib/builder/kerning.py
```diff
@@ -24,25 +24,34 @@ def to_ufo_kerning(self):
...
         master_id = master.id
         kerning_source = master.metricsSource.id  # Maybe be a linked master
-        if kerning_source in self.font.kerning:
-            kerning = self.font.kerning[kerning_source]
+        if kerning_source in self.font.kerningLTR:
+            kerning = self.font.kerningLTR[kerning_source]
             _to_ufo_kerning(self, self._sources[master_id].font, kerning)
+        if kerning_source in self.font.kerningRTL:
+            kerning = self.font.kerningRTL[kerning_source]
+            _to_ufo_kerning(self, self._sources[master_id].font, kerning, "RTL")
```
anthrotype, line 32: since both kerningLTR and kerningRTL are merged into the same ufo.kerning dictionary, is it ever possible that a clash might occur? I.e. the same kern pair defined in both kerningLTR and kerningRTL (potentially with different kern values), which would then mean that the RTL coming last would overwrite the LTR kern pair with the same (first, second) key?
I suppose this is possible in theory given Glypsh3 kerning is now defined in two split dictionaries, so glyphsLib should either raise some error or issue a warning about the ambiguous kern pair.

anthrotype:
> The right side of the left glyph is kerned to the left side of the right glyph

that's quite a mouthful 😄 but I get it now, thanks for the clarification!

So I guess once Georg finishes reworking glyph_data by adding script/direction, this PR will need to be updated accordingly.

I know it's not ideal that glyphsLib merges the two kerning dicts and then ufo2ft splits them again following its own logic, but I'm not sure what what the alternative could be at this stage.

Honestly, I still don't get _why_ the new kerningRTL was added in Glyphs3 to begin with; what problem is it trying to fix?

Lib/glyphsLib/builder/groups.py
```diff
@@ -140,12 +140,28 @@ def _to_glyphs_kerning_group(self, name, glyphs):
...
+            and glyph_data.category != "Number"
```
anthrotype, line 149: btw, there other characters belonging to RTL scripts (e.g. Arabic) that are _not_ "Numbers" (in the unicode category sense of Nd, No, etc.) but which have a (weak) left-to-right bidirectional type: e.g. arabic decimal/thousand separators (Po: Punctuation other), or arabic number sign (Cf: Format), and some others.

In ufo2ft we don't check the category is Number, but we check if the bidiType (`unicodedata.bidirectional`) is either "L" (strong left to right), "AN" (Arabic numbers) or "EN" (extended arabic-indic digits)

anthrotype: Note (to self, nostly) that in ufo2ft, the kerning pairs for the (LTR) arabic numbers are added to the same `kern_rtl` lookup where the rest of the (RTL) kern pairs between arabic non-numbers are; but they are treated as left-to-right, hence only the x-advance is modified (not also the x-origin):

E.g. see this test case:
https://github.com/googlefonts/ufo2ft/blob/ad28eea062e0dd48678309bd9ef86dfcc85fa85a/tests/featureWriters/kernFeatureWriter_test.py#L691-L695

Whereas here, it looks like the `kerningRTL` dict only contains the RTL kerning pairs.

So ufo2ft's `kern_rtl` lookup and Glyphs3 `kerningRTL` don't exactly overlap; the former is a superset of the latter.

anthrotype:
> So I guess once Georg finishes reworking glyph_data by adding script/direction, this PR will need to be updated accordingly.

@schriftgestalt any updates on this?

yanone:
@anthrotype I want to reinvigorate this project as Arabic fonts in Glyphs 3 sources with kerning are slowly piling up.

Georg’s `GlyphsInfo3` branch brings a new implementation of the glyph data and forms the basis of the RTL kerning decisions. I’ve cleaned up the functionality of the branch which fixed almost all issues of the tests. Namely:
- Production name is now available under both `.production` as well as `.production_name` in of the `GlyphInfo` class. Both variants are used in `glyphsLib`, which is probably historically grown. Maybe we should address this and unify all references to either one, but it could break third-party implementations. Glyphs’ own API uses `.productionName` by the way.
- `get_glyph()` returns an unpropagated `GlyphInfo` object as opposed to `None`, as several places in `glyphsLib` (or maybe just tests) are referencing attributes directly and not checking for the object to not be `None`. This is probably the existing behaviour of the main branch, so I've just restored this.

15 tests are now failing, but I can attribute only one of them to glyph data, if at all:
- `tests/builder/builder_test.py:638`: In the GDEF code comparison, `t_e_s_t.alt` is not categorized as a ligature. I’ve added a test to `glyphdata_test.py` to make sure that it is, so I’m not sure where this stems from. In the same check, `wigglylinebelowcomb.alt` isn't categorized as a mark. My guess is that Georg’s glyph data isn’t involved in this, as it doesn’t contain any code to identify marks heuristically, but rather that the mark is identified somewhere else by its `_bottom` anchor. Please have a look at this.

anthrotype:
What about vertical kerning (when layout is top to bottom)? Does glyphsLib support exporting that data already? I presume not, given that vertical kerning in UFO is kind of undefined (see https://github.com/googlefonts/ufo2ft/issues/668)

It would be nice if at the same time as adding support for RTL kerning for Glyphs3 sources, we also supported the vertical kerning..

anthrotype:
I know Glyphs.app does support vertical kerning, my question was more if glyphsLib supports exporting it to UFO and how does it store it (if at all)

yanone:
I had only recently implemented `kerningRTL` in glyphsLib in this PR here and haven't seen any mention of anything else yet, so no

khaledhosny:
Ah, I misread. It does not seem that `kerningVertical` is used anywhere.

anthrotype:
ok. I filed a separate issue https://github.com/googlefonts/glyphsLib/issues/824

eliheuer:
@schriftgestalt I have a few fonts waiting on this, is there something I can do to help move this forward? 

As you know I'm not a big fan of Glyphs becoming the default source format for Google Fonts, I would prefer to work with open-source editors and source files in an open format that is designed to be application independent.

Most likely Google Fonts will continue to use Glyphs as the default format, but I would like to acknowledge that this issue becoming a long drawn out thing that can only be worked on by the internal Glyphs team is a good example of why Google Fonts might want to use more open editors and source formats in the future.

schriftgestalt:
I have finished the GlyphsInfo stuff. Please have a look at the GlyphsInfo3 branch and pull request.

simoncozens:
I've said this privately to Yan, but I think it's worth talking about here as well:

I have some misgivings about the direction of this PR, but I’m aware that some of that may be to do with the fact that I have some misgivings about the whole concept of a separate RTL kerning dictionary. So it’s hard for me to separate my feelings about the PR implementation from my feelings about what it’s implementing.

I think one of the reasons why you're not getting traction on this PR (at least from me), is that conceptually I can't get my head around it. Splitting kerning by direction at the source level doesn't make sense to me. Most glyphs have inherent directionality so you don't need to store this separately. For those which do not, like kerning two common script glyphs together, then what? Two numbers will be displayed in logical order regardless of direction, so that doesn't need a separate RTL dictionary. So really the only case this deals with is punctuation against punctuation. And then why split by *direction*? Perhaps your Hebrew is more tightly spaced than Arabic - then what? I'm sure there are reasons why separate LTR and RTL kerning is a thing. But to be honest, none of them convince me. So it's not something I *really* want to be spending my time thinking about anyway.

And when I do think about it, in terms of the implementation, I feel like this could be much, much simpler. glyphsLib is part of a UFO-based compilation approach where we convert glyphs files to UFO for compilation. There is only one kerning dictionary in UFO. So no matter what we do, everything is going to end up in a single kerning dictionary anyway, at which point it is the responsibility of ufo2ft to turn the contents of that dictionary into GPOS rules. ufo2ft *already does* the clever stuff about working out whether a kern is LTR or RTL.

So my conclusion is that glyphsLib does not need to do that - it just needs to gather all the kerns wherever they live and put them into a UFO kerning file. And now we have a situation where this is all based on a GlyphsInfo file? That makes even less sense to me.

schriftgestalt:
It is not Jan’s fault that their file structure is how it is. 

In Glyphs, there are different aspects about the kerning.
- It needs to be able to produce proper GPOS.
- It needs to display the data in the UI. 
- should the pairs be stored in typing or visual order. Same goes for classes. 

Splitting it into two kerning dicts made the handling of the UI much easier and fixed some edge cases that were in Glyphs 2.

I’m happy to discuss this in detail and if you can explain how to handle it properly, I can change it for the next version.

simoncozens:
I think the way to handle it properly is to combine kerningLTR and kerningRTL into one dictionary and then let ufo2ft do its thing. But I haven't tested this.

yanone:
I'm eager to try.

`_glyph_kerning_attr()` (https://github.com/googlefonts/glyphsLib/blob/support_glyphs3_rtl_kerning-3/Lib/glyphsLib/builder/groups.py#L141, this is my latest branch, not this one) uses the glyph direction to determine the group. You're saying, the group and order should be determined strictly by the origin dict of the pair (LRT or RTL)? I'm not sure where to begin here.

This would work, but will break round-tripping, as for UFO to Glyphs conversion we wouldn't know in which dict to move the pairs then.

yanone:
So I think it's really the group where the problem lies here. The kerning dict conversion 
 in [kerning.py](https://github.com/googlefonts/glyphsLib/blob/support_glyphs3_rtl_kerning-3/Lib/glyphsLib/builder/kerning.py#L35) probably won't need changing.

But the two separate kerning dicts are only half of the story. Since Georg also decided to switch the kerning groups from `leftKerningGroup` to `rightKerningGroup` and vv for RTL, we need to know the writing direction of each glyph for [groups.py](https://github.com/googlefonts/glyphsLib/blob/support_glyphs3_rtl_kerning-3/Lib/glyphsLib/builder/groups.py#L141), which is how we ended up here in the first place.

yanone:
The only solution to this that I can see without Georg’s updated glyph data is to decide the group sides based on whether or not a glyph can be found as part of the RTL kerning, so walk the dict for each group decision.

simoncozens:
**ufo2ft already does the clever stuff about working out whether a kern is LTR or RTL.**

anthrotype:
> So I think it's really the group where the problem lies here... Since Georg also decided to switch the kerning groups from leftKerningGroup to rightKerningGroup and vv for RTL, we need to know the writing direction of each glyph for groups.py, which is how we ended up here in the first place.

Yes, Yanone is right, this is the crux of the problem. It's not sufficient to simply merge the two kerning dicts. We need to find a way to deal with the groups as well.

simoncozens:
The groups are going to be split by the kern splitter, though.

anthrotype:
the problem is `_glyph_kerning_attr` method, which determines what is to be put in public.kern1 and public.kern2 kerning groups, used here:

https://github.com/googlefonts/glyphsLib/blob/f0128057019daf0453b2f7dbc06665637ba2cb1e/Lib/glyphsLib/builder/groups.py#L75-L82

This returs "leftKerningGroup" or "rightKerningGroup" depending on the UFO group's logical position (1 or 2). For Glyphs-2 kerning, this would always be "rightKerningGroup" for public.kern1 and "leftKerningGroup" for public.kern2.
Now, since Glyphs3, for RTL kerning, these have been swapped:  i.e. "leftKerningGroup" for public.kern1 and "rightKerningGroup" for public.kern2.

anthrotype:
> will break round-tripping, as for UFO to Glyphs conversion we wouldn't know in which dict to move the pairs then

we could work around this by storing stuff in the lib, but at this stage is more urgent to be able to compile these fonts, we can add support for roundtripping at a later stage

yanone:
You have no idea how delighted I am that someone utters the word _urgency_ in this context other than me. It is indeed the case that this whole disaster is currently leading to Arabic and Hebrew typography being slowed down and disadvantaged. I know of several projects by several studios that are on hold since a while now, they're just not speaking out about it even though I asked them to.

simoncozens:
So how I would attempt to approach this would be to put everything into two groups, `public.kern1.A`, `public.kern2.RTL.A` and vice versa, and when merging the two dictionaries together, rewrite the groups depending on whether you're processing `kerningLTR` or `kerningRTL`.

anthrotype:
hm you're onto something, keep going!

yanone:
So basically [here](https://github.com/googlefonts/glyphsLib/blob/support_glyphs3_rtl_kerning-3/Lib/glyphsLib/builder/kerning.py#L35)?

yanone:
> So how I would attempt to approach this would be to put everything into two groups, e.g. `public.kern1.A` and `public.kern2.RTL.A` if it's a left glyph, or `public.kern2.A` and `public.kern1.RTL.A` if it's a right glyph, and when merging the two dictionaries together, rewrite the groups depending on whether you're processing `kerningLTR` or `kerningRTL`.

I understand the first part, but I don't understand where the merging is happening that Simon is speaking of.

simoncozens:
OK, I'm going to avoid using real examples because if I do, I'll get horribly confused with `MMK_L_` and `MMK_R_` and muck it up, and that won't help explain anything.

So the first step is to store each glyph into `@MMK_some_kerning_group` and *also* into `@MMK_some_other_kerning_group_RTL`.

After we've built all those groups, what we *currently* do is:
*  iterate through each pair defined in `kerning` or `kerningLTR`
* for each pair, if it contains a group, transform the groups into `@public.kernX.whatever` where `X` depends on whether it was left or right
* then emit the kern pair into the UFO kerning.

We keep doing that. But then next, we also do the same thing but for `kerningRTL`:
* for each pair, if it contains a group, transform the groups into `@public.kernX.RTL.whatever` where `X` depends on whether it was *right* or *left*. (i.e. switching it round here)
* then emit the kern pair into the UFO kerning.

That will work for all cases except for situations where the same glyph pair has both LTR and RTL kerning, with different values in each table. But we were never going to get that right anyway.

anthrotype:
wow.. and the result of this very complicated process is basically to _undo_ Glyphs3's LTR vs RTL kerning split such that ufo2ft can later on redo that in its own way (ie. using cmap+GSUB closure to identify script and pair direction).

schriftgestalt:
> That will work for all cases except for situations where the same glyph pair has both LTR and RTL kerning, with different values in each table. But we were never going to get that right anyway.

That is also one reason why I did the split.

yanone:
Gentlemen, I’ve implemented Simon’s idea and am happy to announce that I think it works. I made it into a separate PR for clarity: https://github.com/googlefonts/glyphsLib/pull/838

yanone:
This is the first of the three related PRs and has now been fixed in https://github.com/googlefonts/glyphsLib/pull/865, so closing this too.

PR #781, opened 2022-02-23
Remap G3 width class to user value

### LINKED ISSUE #780: Width class mapping in G3
When mapping user-space to design-space in the designspace file, glyphsLib uses the "Width Class" in the instances of the source file, not the axis coordinates. However, Glyphs does not write a `widthClass` for the default value (5). For some reason, glyphsLib then infers a `widthClass` of 100, leading to a designspace like this:

```
    <axis tag="wdth" name="Width" minimum="2" maximum="100" default="100">
      <map input="2" output="70"/>
      <map input="3" output="79"/>
      <map input="4" output="89"/>
      <map input="100" output="100"/>
    </axis>
```

simoncozens:
Ensures that `class_to_value` is run on the width value coming from Glyphs 3 source files. Fixes #780.

### DIFF
--- a/Lib/glyphsLib/builder/axes.py
+++ b/Lib/glyphsLib/builder/axes.py
@@ -73,20 +73,23 @@ def user_loc_string_to_value(axis_tag, user_loc):
     87.5
     >>> user_loc_string_to_value('wdth', 'Clearly Not From Glyphs UI')
     """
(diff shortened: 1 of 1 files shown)


anthrotype (approved):
LGTM. As usual, thanks for taking of it 👍

Issue #694, opened 2021-07-05
ufo2glyphs unhandled AttributeError: 'MarkClassName' object has no attribute 'glyphclass'


DavidRaymond:
When running ufo2glyphs I'm seeing the following traceback:
```Traceback (most recent call last):
  File "/home/david/src/glyphsLib/Lib/glyphsLib/builder/features.py", line 595, in _process_glyph_class_definition
    if st.glyphs.original:
AttributeError: 'MarkClassName' object has no attribute 'original'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/local/bin/ufo2glyphs", line 11, in <module>
    load_entry_point('glyphsLib', 'console_scripts', 'ufo2glyphs')()
  File "/home/david/src/glyphsLib/Lib/glyphsLib/cli.py", line 293, in _ufo2glyphs_entry_point
    return main(args)
  File "/home/david/src/glyphsLib/Lib/glyphsLib/cli.py", line 189, in main
    return options.func(options)
  File "/home/david/src/glyphsLib/Lib/glyphsLib/cli.py", line 266, in ufo2glyphs
    font = glyphsLib.to_glyphs(
  File "/home/david/src/glyphsLib/Lib/glyphsLib/builder/__init__.py", line 145, in to_glyphs
    return builder.font
  File "/home/david/src/glyphsLib/Lib/glyphsLib/builder/builders.py", line 749, in font
    self.to_glyphs_features()
  File "/home/david/src/glyphsLib/Lib/glyphsLib/builder/features.py", line 298, in to_glyphs_features
    _to_glyphs_features(
  File "/home/david/src/glyphsLib/Lib/glyphsLib/builder/features.py", line 317, in _to_glyphs_features
    processor.to_glyphs(font)
  File "/home/david/src/glyphsLib/Lib/glyphsLib/builder/features.py", line 496, in to_glyphs
    self._process_file()
  File "/home/david/src/glyphsLib/Lib/glyphsLib/builder/features.py", line 508, in _process_file
    or self._process_glyph_class_definition()
  File "/home/david/src/glyphsLib/Lib/glyphsLib/builder/features.py", line 612, in _process_glyph_class_definition
    elements.append("@" + st.glyphs.glyphclass.name)
AttributeError: 'MarkClassName' object has no attribute 'glyphclass'
```
[...]

PR #782, opened 2022-02-24
Handle features.fea code where a mark class is used as a glyph class

nrsiward:
see issue googlefonts/glyphsLib#694

### DIFF
--- a/Lib/glyphsLib/builder/features.py
+++ b/Lib/glyphsLib/builder/features.py
@@ -744,21 +744,30 @@ class FeatureFileProcessor:
             if st.glyphs.original:
                 for glyph in st.glyphs.original:
                     try:
-                        # Class name (GlyphClassName object)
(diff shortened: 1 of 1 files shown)


DavidRaymond:
What is the status of this?  It would be great to have this PR merged.

anthrotype:
thanks for the PR.

> Is anything now preventing the merge?

the whole block of code feels a bit hackish, with all the FIXME and the all-catching AttributeErrors. I know you simply tried to fix your particular issue by adapting to the style of the neighbouring lines.

Also there's no accompanying tests to cover the newly added lines.

anthrotype:
@nrsiward i'm OK to merge if you can add a small test for this in https://github.com/googlefonts/glyphsLib/blob/main/tests/builder/features_test.py

anthrotype:
I will add the test, thanks

PR #783, opened 2022-04-01
Fix brace layer parsing to force int where there may be strings

jpt:
I was running into this error while converting a file that uses brace layers:

```
    brace_layers.sort(key=lambda x: tuple(x.location.values()))
TypeError: '<' not supported between instances of 'str' and 'int'
```

The issue was that my Glyphs file had some integers in its brace layer axis locations, along with some strings. I'm not sure how or why this divergence occurred, but I think it would make sense to force an `int` for these values.

### DIFF
--- a/Lib/glyphsLib/builder/sources.py
+++ b/Lib/glyphsLib/builder/sources.py
@@ -164,7 +164,7 @@ def _to_designspace_source_layer(self):
     designspace = self._designspace
     layers_to_insert = collections.defaultdict(list)
     for key, master_ids in layer_to_master_ids.items():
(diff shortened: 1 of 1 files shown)


anthrotype:
i'm not sure the fix is correct. Can you provide a reproducer?

jpt:
I'll work on putting together a minimal file that can reproduce this, thanks for looking into it

bramstein:
For what it is worth: we also ran into this issue with one of our families. It seems the most recent (unreleased) Glyphs version 3.2 automatically fixes this upon saving.

PR #786, opened 2022-04-28
Don't write special "Variable Font Setting" instances to designspace

jenskutilek:
Glyphs 3 stores export settings for variable fonts in special `GSInstance` instances.

Those are currently not special-cased by glyphsLib, and are written as a normal instance into designspaces files, which leads to duplicate instances in the exported variable font.

This PR avoids exporting them to a designspace, but more work needs to be done to preserve them in roundtripping. Probably they should be written to the designspace lib and when going from DS to Glyphs, be reconstructed from there.

### DIFF
--- a/Lib/glyphsLib/classes.py
+++ b/Lib/glyphsLib/classes.py
@@ -141,6 +141,9 @@ CURVE = "curve"
 OFFCURVE = "offcurve"
 QCURVE = "qcurve"
 
+INSTANCETYPESINGLE = 0
+INSTANCETYPEVARIABLE = 1
+
 TAG = -2
 TOPGHOST = -1
 STEM = 0
@@ -3020,6 +3023,7 @@ class GSInstance(GSBase):
         writer.writeObjectKeyValue(self, "active", condition=(not self.active))
         if writer.format_version > 2:
             writer.writeObjectKeyValue(self, "axes", keyName="axesValues")
+            writer.writeObjectKeyValue(self, "type", keyName="type", condition="if_true")
         writer.writeObjectKeyValue(self, "exports", condition=(not self.exports))
         writer.writeObjectKeyValue(self, "customParameters", condition="if_true")
         if writer.format_version == 2:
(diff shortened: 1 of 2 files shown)


anthrotype:
I don't know what '"Variable Font Setting" instance' mean, could you elaborate? Is there any link to documentation?

anthrotype:
oh, I just discovered what used to be named "Instances" tab is now named "Exports" and the plus button has 

image

jenskutilek:
Exactly, those are what I mean. They can be used as a vehicle for custom parameters, filters, and other properties for a VF export.

Programmatically, they can be differentiated by `GSInstance.type`:
- https://docu.glyphsapp.com/#GSInstance.type
- https://docu.glyphsapp.com/#instance-types

In the Glyphs tutorial:
- https://glyphsapp.com/learn/creating-a-variable-font#g-step-5-add-a-variable-font-setting

Lib/glyphsLib/classes.py
```diff
@@ -3090,12 +3095,32 @@ def __init__(self):
...
 
     customParameters = property(
         lambda self: CustomParametersProxy(self),
         lambda self, value: CustomParametersProxy(self).setter(value),
     )
 
+    @property
+    def type(self):
+        if self._type == INSTANCETYPEVARIABLE:
+            return "variable"  # "variable" is stored in plist
```
anthrotype, line 3110: it's weird this `type` property returns either a constant string literal "variable" or a False bool value. What does Glyphs.app scripting API do in this case?

anthrotype: > the plist the value is stored as str "variable" or not at all

then the conversion from `1` to `"variable"` should be done inside `_serialize_to_plist` method before writing the `type` key

anthrotype:
@jenskutilek I see this somehow conflicts/overlaps with @yanone's PR that we merged recently: https://github.com/googlefonts/glyphsLib/pull/812 -- if there's anything from this PR that needs to be included, please rebase and adjust accordingly, otherwise we can close this

jenskutilek:
I've added some leftovers of this to #815.

PR #790, opened 2022-07-01
Fix for #789: don't drop 'name' attribute of first point of open contour when drawing to pen

### LINKED ISSUE #789: [pen/ufo] First point of open contour doesn't transmit its "name" attribute
An open contour where the first point has a "name", when written to UFO loses the "name" attribute.

I have located the problem and will submit a PR shortly.

justvanrossum:
Make the first point of an open contour behave the same wrt. the 'name' attribute as other points.

This fixes #789.


justvanrossum:
Here is a test font. I'm not familiar enough with the code base to quickly see how/where such a test case could be added.

[OpenContourFirstPointName.glyphs.zip](https://github.com/googlefonts/glyphsLib/files/9031249/OpenContourFirstPointName.glyphs.zip)

justvanrossum:
Anyone willing to review this? It is a blocker for our [Nabla COLRv1 GF](https://github.com/justvanrossum/nabla)

anthrotype (approved):

PR #796, opened 2022-07-18
Set feature for designspace rules

simoncozens:
Glyphs has a mechanism (the "Feature for Feature Variations" custom parameter) for setting which feature should be used for processing bracket rules. This implements that custom parameter, passing the feature information to designspace. If the user requests `rclt`, we set `processing=last` on the rules tag; otherwise, we use the equivalent designspace lib key `com.github.fonttools.varLib.featureVarsFeatureTag`.

### DIFF
--- a/Lib/glyphsLib/builder/builders.py
+++ b/Lib/glyphsLib/builder/builders.py
@@ -497,7 +497,7 @@ class UFOBuilder(_LoggerMixin):
         feat = self.font.customParameters["Feature for Feature Variations"]
         if feat == "rclt":
             self._designspace.rulesProcessingLast = True
-        elif feat != "rvrn":
+        elif feat and feat != "rvrn":
             self._designspace.lib[FEAVAR_FEATURETAG_LIB_KEY] = feat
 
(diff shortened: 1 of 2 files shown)


Lib/glyphsLib/builder/builders.py
```diff
@@ -492,6 +493,13 @@ def _apply_bracket_layers(self):
                 )
                 self._designspace.addRule(rule)
 
+        # Set feature for rules
+        feat = self.font.customParameters["Feature for Feature Variations"]
+        if feat == "rclt":
+            self._designspace.rulesProcessingLast = True
+        elif feat and feat != "rvrn":
+            self._designspace.lib[FEAVAR_FEATURETAG_LIB_KEY] = feat
```
anthrotype, line 501: how about always using the FEAVAR_FEATURETAG_LIB_KEY and forget about the rulesProcessingLast?

anthrotype: when the lib key is set, the `processing` attribute is ignored (it's always implicitly "last" for anything but 'rvrn' which is special and always applied first).

PR #804, opened 2022-09-01
Builder refactor, part 1

simoncozens:
This is a set of (hopefully uncontroversial) refactors to the builder code. The main changes are

* Moving *any* constants to the `constants` module.
* Moving a number of utility functions to the `utils` module.
* Converting the tests from `unittest` to `pytest`.

The idea behind this is to prepare for a more invasive PR which starts to split the Glyphs file parsing/representation, UFO building, and Glyphs building functionality into separate Python modules.


Lib/glyphsLib/builder/axes.py
```diff
@@ -17,24 +17,7 @@
 
 from glyphsLib import classes
 from glyphsLib.classes import WEIGHT_CODES, WIDTH_CODES
-from .constants import GLYPHLIB_PREFIX
-
-# This is a key into GSFont.userData to store axes defined in the designspace
-AXES_KEY = GLYPHLIB_PREFIX + "axes"
```
anthrotype, line 23: was this AXES_KEY unused? looks like you removed it

Lib/glyphsLib/builder/constants.py
```diff
@@ -11,20 +11,84 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+import re
 
 
 PUBLIC_PREFIX = "public."
+GLYPH_ORDER_KEY = PUBLIC_PREFIX + "glyphOrder"
```
anthrotype, line 18: maybe we could even sort these constant alphabetically (or maybe not)

anthrotype (approved):
Thanks for this! I haven't looked closely at all the test modules that you converted from unittest to pytest (sorry, it's too many 😅) but I trust that you kept them all and test coverage hasn't decreased.

anthrotype:
> split the Glyphs file parsing/representation, UFO building, and Glyphs building functionality into separate Python modules.

i'm still not 100% convinced that this won't actually increase the maintenance cost by triplicating all the chores of setting things up for testing, deployment, issue tracking, releasing (sometimes in locksteps for they will continue to depend on one another).

simoncozens:
Here's how I see it:

* Separating parsing/representation from building is a no-brainer to me; these are separable operations and we can test each element separately.
* The builder itself is very complicated and hard to maintain.
* I'd like to experiment with going Glyphs straight to fontTools, but it's very hard to follow the flow of logic in glyphsLib as is.
* Part of the reason why the builder is complicated is that we try to round-trip every submodule. It would be much less complicated to see just the flow of the Glyphs->UFO operation if that was clearly marked out.
* We want to put most maintenance effort into Glyphs->UFO but the round-tripping requirement means we also have to put equivalent effort into UFO->Glyphs.
* Most maintainers are not all that interested in UFO->Glyphs. Those who *are* interested in UFO->Glyphs are only interested in certain functionality, not all of it. So maintaining a full round-tripping toolchain for every bit of functionality is not actually something *anyone* actually wants.

So separating UFO->Glyphs and Glyphs->UFO will actually *decrease* maintainer load, by allowing these projects to have different goals and different levels of functionality support and allowing maintainers to focus on the parts of the problem that they actually need.

Lib/glyphsLib/builder/axes.py
simoncozens: Yeah, removed unused constant in e1d1e2e3. I've been running the test suite between commits and it works.

simoncozens:
Gonna do some more.

simoncozens:
OK, I'm going to stop here with this one. Changing the `minimal`/`minimize_glyphs_differences` thing will cause regression tests to fail, because there are things we're currently storing in the UFOs that we don't actually need to. From a tests/regression tests perspective, this PR is at least clean.

simoncozens:
Ping @anthrotype? I know this has become massive, but most bits of it are just small refactors. There's no behaviour changes, and the regression tests are clean. It would unlock work on bracket layers. I could split off the refactors to tests/ if it would be more reassuring.

Lib/glyphsLib/builder/features.py
```diff
@@ -21,7 +21,15 @@
...
 
-from .constants import GLYPHLIB_PREFIX
+from glyphsLib.util import PeekableIterator
+from .constants import (
+    GLYPHLIB_PREFIX,
+    ANONYMOUS_FEATURE_PREFIX_NAME,
+    ORIGINAL_FEATURE_CODE_KEY,
+    ORIGINAL_CATEGORY_KEY,
+    _LANGUAGE_MAPPING,
+    _REVERSE_LANGUAGE_MAPPING,
```
belluzj, line 31: Maybe these two should not have an underscore anymore?

belluzj (approved):
LGTM, just a minor comment about naming two constants, thanks a lot @simoncozens

Lib/glyphsLib/builder/glyph.py
```diff
@@ -146,6 +146,20 @@ def to_ufo_glyph(self, ufo_glyph, layer, glyph, do_color_layers=True):  # noqa:
         self.to_ufo_layer_user_data(ufo_glyph, layer)  # .user_data
         self.to_ufo_smart_component_axes(ufo_glyph, glyph)  # .components
 
+    # Optimization: profiling glyphs2ufo of NotoSans-MM.glyphs (6000 glyphs) on a Mac
+    # mini late 2014, Python 3.6.8, revealed that a whopping 17% of the time was spent
+    # converting lastChange to UFO timestamps. I could not reproduce this on a Windows
+    # 10/Python 3.7 setup, so this might be a platform thing. If-guarding anyway
+    # because these timestamps are useless in a UFO scenario if you use Git.
+
+    # This *should* be gated with not self.minimal, but that breaks regression tests
```
anthrotype, line 155: regression tests breaking is fine when the change is desired (will auto-fix themselves upon merging to main but need to be reviewed manually and accepted as good).
But maybe it's fine to do this change in a separate PR

anthrotype (approved):
LGTM, thank you for taking on the ungrateful task 🙏

PR #808, opened 2022-09-15
Support "meta Table" custom parameter

### LINKED ISSUE #693: support for `meta` table needed
ufo2ft [now supports](https://github.com/googlefonts/ufo2ft/pull/509) the `meta` table [defined in UFO3](https://unifiedfontobject.org/versions/ufo3/lib.plist/#publicopentypemeta).  

However, it doesn't look as though glyphsLib supports it yet: I tried round-tripping a UFO to glyphs and back, and the meta table info was removed -- but maybe I did something wrong?

simoncozens:
Closes #693.

- Convert "meta Table" to public.openTypeMeta
- Tests for meta table implementation

### DIFF
--- a/Lib/glyphsLib/builder/custom_params.py
+++ b/Lib/glyphsLib/builder/custom_params.py
@@ -28,6 +28,7 @@ from .constants import (
     REVERSE_CODEPAGE_RANGES,
     PUBLIC_PREFIX,
     UFO_FILENAME_CUSTOM_PARAM,
+    UFO2FT_META_TABLE_KEY,
 )
 from .features import replace_feature, replace_prefixes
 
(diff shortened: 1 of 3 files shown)


Lib/glyphsLib/builder/custom_params.py
```diff
@@ -529,6 +530,43 @@ def to_glyphs_gasp_table(value):
 )
 
 
+# convert Glyphs' meta Table to UFO openTypeMeta
+def to_ufo_meta_table(value):
+    meta = {}
+    # In:  {data = "de-Latn"; tag = dlng; }, {data = "sr-Cyrl"; tag = slng; }
+    # Out: { "dlng": [ "de-Latn" ], "slng": [ "sr-Cyrl" ] }
```
anthrotype, line 537: fonttools and UFO spec currently store meta data as a dictionary thus disallowing entries with duplicate keys, whereas Glyphs.app follows OT spec and allows that:

https://github.com/fonttools/fonttools/issues/2345
https://github.com/unified-font-object/ufo-spec/issues/187

Maybe that will change if somebody bothers to change that, but for now that's how it is, I suggest you log a warning in the case of duplicate keys. Right now you are taking the last, either that or taking first is fine, both are equally "wrong" anyway.

anthrotype (approved):
Thanks! (see comment about duplicate meta keys)

PR #810, opened 2022-09-22
Also use instance "Axis Location" params for building axis mappings

### LINKED ISSUE #714: Axis mappings are missing in designspace
It seems that glyphsLib only adds axis mappings from the "Axis Location" custom parameter of masters, but not those of exports. Thus the avar table when building with fontmake/glyphsLib is different than when exporting the VF directly from Glyphs.

jenskutilek:
Fixes #714


anthrotype:
thanks Jens, looks good to me. Is the test failure expected? Please double check and if so adjust the expectations. If we don't have any instances with Axis Location custom parameter among the tests we should add one for this.

jenskutilek:
Not sure why the tests fail; they fail locally for me too, but in a different way, and I can't make sense of the output. Any idea what's broken on my machine? Example output:

```
__________________________________________________________ DesignspaceTest.test_basic __________________________________________________________

self = <tests.builder.interpolation_test.DesignspaceTest testMethod=test_basic>

    def test_basic(self):
        masters, instances = makeFamily()
        font = makeFont(masters, instances, "DesignspaceTest Basic")
        doc = to_designspace(font, instance_dir="out")
>       self.expect_designspace(doc, "DesignspaceTestBasic.designspace")

tests/builder/interpolation_test.py:179:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
tests/builder/interpolation_test.py:155: in expect_designspace
    return self._expect_designspace(doc, expected_path)
tests/builder/interpolation_test.py:167: in _expect_designspace
    self.fail("*.designspace file is different from expected")
E   AssertionError: *.designspace file is different from expected
------------------------------------------------------------- Captured stderr call -------------------------------------------------------------
DesignspaceTestBasic.designspace discrepancies (per xmldiff):
  [update-attribute, /designspace[1], format, "5.0"]  [move, /designspace/instances/instance[2]/kerning[1], /designspace[1], 2]  [rename, /designspace/kerning[1], instances]  [move, /designspace/instances[2]/instance[1], /designspace/instances[1], 0]  [insert, /designspace/instances[1], instance, 1]  [insert-attribute, /designspace/instances[1]/instance[2], familyname, "DesignspaceTest Basic"]  [insert-attribute, /designspace/instances[1]/instance[2], filename, "out/DesignspaceTestBasic-Semibold.ufo"]  [insert-attribute, /designspace/instances[1]/instance[2], name, "DesignspaceTest Basic Semibold"]  [insert-attribute, /designspace/instances[1]/instance[2], stylemapfamilyname, "DesignspaceTest Basic Semibold"]  [insert-attribute, /designspace/instances[1]/instance[2], stylemapstylename, "regular"]  [insert-attribute, /designspace/instances[1]/instance[2], stylename, "Semibold"]  [move, /designspace/instances[2]/instance[2], /designspace/instances[1], 2]  [move, /designspace/instances[2]/instance[1], /designspace/instances[1], 3]  [update-attribute, /designspace/instances[1]/instance[4], filename, "out/DesignspaceTestBasic-Black.ufo"]  [update-attribute, /designspace/instances[1]/instance[4], name, "DesignspaceTest Basic Black"]  [update-attribute, /designspace/instances[1]/instance[4], stylemapfamilyname, "DesignspaceTest Basic Black"]  [update-attribute, /designspace/instances[1]/instance[4], stylename, "Black"]  [move, /designspace/instances[1]/instance[1]/kerning[1], /designspace/instances[1]/instance[2], 0]  [rename, /designspace/instances[1]/instance[2]/kerning[1], location]  [move, /designspace/instances[1]/instance[1]/info[1], /designspace/instances[1]/instance[3], 0]  [rename, /designspace/instances[1]/instance[3]/info[1], location]  [move, /designspace/instances[1]/instance[4]/location/dimension[1], /designspace/instances[1]/instance[2]/location[1], 0]  [move, /designspace/instances[1]/instance[3]/location[2]/dimension[1], /designspace/instances[1]/instance[3]/location[1], 0]  [move, /designspace/instances[2]/instance/location/dimension[1], /designspace/instances[1]/instance[4]/location[1], 0]  [delete, /designspace/instances[2]/instance/info[1]]  [delete, /designspace/instances[2]/instance/kerning[1]]  [delete, /designspace/instances[2]/instance/location[1]]  [delete, /designspace/instances[2]/instance[1]]  [delete, /designspace/instances[2]]  [delete, /designspace/instances/instance[4]/info[1]]  [delete, /designspace/instances/instance[3]/info[1]]  [delete, /designspace/instances/instance[3]/kerning[1]]  [delete, /designspace/instances/instance[3]/location[2]]
```

anthrotype:
hugh.. that xmldiff thing was contributed by some user but really it's way less readable than a dump textual diff

anthrotype:
maybe yours fail because it's using designspace 5.0, which produces slightly different output/. Try using the same requirements.txt as the test environment

anthrotype:
or just use `tox` as the ci does

jenskutilek:
I run `tox` locally, but still get different results, and test failures even on the main branch!?

jenskutilek:
Something was broken; after removing the `.tox` dir everything works as on CI. Maybe because I accidentally ran `tox` from a venv previously?

anthrotype:
so yeah, probably you had created that local .tox environment before the upstream requirements.txt was updated, and tox isn't smart enough to figure out that it needs to re-create the local venv when that changes. Also `tox -r` would force tox recreate the venv.

anthrotype:
actually according to https://github.com/tox-dev/tox/issues/149 that old tox issue had been fixed with tox 4, what version are you using?

jenskutilek:
I have tox 3.26; tox 4 is still in beta, that's why I didn't get it automatically

anthrotype:
nevermind, the latest tox on pypi is 3.26 but that thread is tagged "fixed-by-tox4" and the latter is still in "rewrite"

anthrotype:
@jenskutilek what's the status of this?

jenskutilek:
I'll continue to work on it shortly.

jenskutilek:
I think the old behaviour was wrong and have updated the test expectations accordingly.

The test font BraceTestFontV3 has "Axis Location" CPs on the masters:

```
Width          Weight
User  Design   User  Design
  75      50    100     100
 100     100    700    1000
```

The instances don't have custom parameters, so the locations would be deduced from the weight class and width values.

With my PR, the instance locations are always evaluated, so the mapping changes as follows:

```
Width          Weight
User  Design   User  Design
  75      50    100     100
  87.5    75    400     500 <- from "Semi Consensed" instance
 100     100    700    1000
```

The mapping 87.5 to 75 is exactly halfway between the masters, so it doesn't change the result (it could be argued that it should be optimized away). But the 400 to 500 mapping on the weight axis does change the result, and was missing before. I believe the new behaviour is correct.

WDYT @anthrotype?

jenskutilek:
Not sure how to interpret the results of the regression tests:

```
! /tmp/tmpfcd0gzbt/RasaItalics-MM.designspace - tests/data/gf/RasaItalics-MM/RasaItalics-MM.designspace--- 
86
+++ 
87
@@ -3,9 +3,6 @@
88    <axes>
89      <axis tag="wght" name="Weight" minimum="300" maximum="700" default="300">
90        <map input="300" output="165"/>
91 -      <map input="400" output="214"/>
92 -      <map input="500" output="263"/>
93 -      <map input="600" output="312"/>
94        <map input="700" output="361"/>
95      </axis>
96    </axes>
```

Does that mean that the lines with the minus sign are removed in the new or the old version? Because when I convert to designspace manually with `glyphs2ufo`, the marked lines are there.

Anyway it doesn't make a difference in result, as the intermediate mappings are linear between the masters, and thus could be removed.

Lib/glyphsLib/builder/axes.py
```diff
@@ -403,9 +415,8 @@ def set_user_loc(self, master_or_instance, value):
                         pass
             return
 
-        # For masters, set directly the custom parameter (old way)
-        # and also the Axis Location (new way).
-        # Only masters can have an 'Axis Location' parameter.
```
anthrotype, line 408: you modified this comment removing "Only masters can have an 'Axis Location'" but the code around the comment is the same as before; when master_or_instance is an instance and not a master, the code above returns early and skips the following block. So I'm not sure whether the comment or the code should be right

tests/builder/designspace_gen_test.py
```diff
@@ -164,8 +164,8 @@ def test_designspace_generation_brace_layers(datadir, filename, ufo_module):
         (a.name, a.minimum, a.default, a.maximum, a.map) for a in designspace.axes
     ]
     assert axes_order == [
-        ("Width", 75, 100, 100, [(75, 50.0), (100, 100.0)]),
-        ("Weight", 100, 100, 700, [(100, 100.0), (700, 1000.0)]),
+        ("Width", 75, 100, 100, [(75, 50), (87.5, 75), (100, 100)]),
+        ("Weight", 100, 100, 700, [(100, 100), (400, 500), (700, 1000)]),
```
anthrotype, line 168: I presume these additional mappings coming from intermediate instances. Do they contain an explicit Axis Location parameter? If not, aren't these redundant? It looks like they are exactly the same value as it would be linearly interpolated between the previous and next mappings so they have no effect. I don't think varLib will prune those when making avar (maybe it could/should).

anthrotype:
> the minus sign

yeah that's a bit confusing. I think it means the original reference source file does not contain those lines, whereas the new updated one does (proabably would be clearer if minus/plus were flipped..)

anthrotype:
> Anyway it doesn't make a difference in result, as the intermediate mappings are linear between the masters, and thus could be removed.

hm so we're now considering instances when making wght/wdth mappings even if they do not contain Axis Location custom parameter. Is that actually needed? Was the previous behavior to only consider the masters for the axes maps wrong?
The tile of the PR says 'Also use instance "Axis Location" params for building axis mappings', but I don't see that tested explicitly.

tests/builder/designspace_gen_test.py
jenskutilek: The width axis mapping is redundant, the weight axis one not.

Regular has weight class 400, so IIUC in absence of any Axis Location parameters (on masters and instances), glyphsLib would deduce the mapped value from the weight class. Ignoring the weight class would put Regular at 366.6667.

Using the weight class to map Regular seems more correct to me.

But comparing fonts generated by Glyphs and glyphsLib, it seems that Glyphs doesn't bother mapping instances either in this constellation.

anthrotype: >  it seems that Glyphs doesn't bother mapping instances either

which would be an argument for _not_ doing it here.. unless an explicit Axis Location custom parameter was used on the instance (which is what this PR presumably was about)?

jenskutilek: Yep. I'll change it…

jenskutilek:
I've changed the logic so that if the font has Axis Location parameters on all masters, only locations from this parameter (for masters and instances) will be used for building the mapping.

I'll add some tests tomorrow.

Lib/glyphsLib/builder/axes.py
```diff
@@ -165,12 +165,16 @@ def to_designspace_axes(self):
...
+            # If all masters have an "Axis Location" custom parameter, only the values
+            # from this parameter will be used to build the mapping of the masters and instances
             # TODO: (jany) use Virtual Masters as well?
+            #       (jenskutilek) virtual masters can't have an Axis Location parameter.
             mapping = {}
             for master in self.font.masters:
                 designLoc = axis_def.get_design_loc(master)
-                userLoc = axis_def.get_user_loc(master)
+                userLoc = axis_def.get_user_loc_from_axis_location_cp(master)
+                if userLoc is None:
```
anthrotype, line 176: isn't this check for None-ness ineffective, given that `font_uses_axis_locations` already ensures that all the masters have Axis Location custom parameter and as such will have a non-None userLoc?

anthrotype: if anything we want to `assert userLoc is not None`

Lib/glyphsLib/builder/axes.py
jenskutilek: The code doesn't differentiate between master and instance, so as I understand it, it has set the custom parameter on masters and instances alike all along.

Lib/glyphsLib/builder/axes.py
```diff
@@ -165,12 +165,16 @@ def to_designspace_axes(self):
...
         elif font_uses_axis_locations(self.font):
-            # Build the mapping from the "Axis Location" of the masters and instances
+            # If all masters have an "Axis Location" custom parameter, only the values
+            # from this parameter will be used to build the mapping of the masters and instances
             # TODO: (jany) use Virtual Masters as well?
+            #       (jenskutilek) virtual masters can't have an Axis Location parameter.
             mapping = {}
             for master in self.font.masters:
                 designLoc = axis_def.get_design_loc(master)
-                userLoc = axis_def.get_user_loc(master)
```
anthrotype, line 173: hm wait a sec.. are you now completely byassing `get_user_loc` method that would previously do the guessing based on weight class and only considering Axis Location? I think before the latter would override implicit weight/width class settings if not None, but if None those implicit settings would still count.. now it seems you simply skip them in the None case..

Lib/glyphsLib/builder/axes.py
anthrotype: No, look at the `if hasattr(master_or_instance, "instanceInterpolations"):` block just above; that is True for any GSInstance, by definition (if I an not mistaken) and it `return`s early thus skipping anything that comes afterwards which is then assumed to only apply for non-instances, hence masters...

Lib/glyphsLib/builder/axes.py
anthrotype: but it may be present while also omit some axis coordinates, can it not? which is why you are checking for None.. whereas before it was never None, because even if an axis was omitted from an Axis Location parameter, there would still be a userLoc defined (inferred from other values)

jenskutilek: > but it may be present while also omit some axis coordinates, can it not? which is why you are checking for None.. whereas before it was never None, because even if an axis was omitted from an Axis Location parameter, there would still be a userLoc defined (inferred from other values)

I haven't checked in detail, but it seems that via the UI in Glyphs it is impossible to omit coordinates for some axes.

Lib/glyphsLib/builder/axes.py
schriftgestalt: checking for `instanceInterpolations` is quite weak. Wouldn’t be `isinstance(master_or_instance, GSInstance)` be better?

anthrotype: yes probably.. i'm not sure to be honest why @belluzj did it like that.

Lib/glyphsLib/builder/axes.py
jenskutilek: If the Axis Location parameter is disabled in a master or instance, it will return `None`. I’ve updated my code accordingly.

Lib/glyphsLib/builder/axes.py
```diff
@@ -165,40 +192,46 @@ def to_designspace_axes(self):
...
+            # from this parameter will be used to build the mapping of the masters and
+            # instances
             # TODO: (jany) use Virtual Masters as well?
+            #       (jenskutilek) virtual masters can't have an Axis Location parameter.
             mapping = {}
             for master in self.font.masters:
                 designLoc = axis_def.get_design_loc(master)
                 userLoc = axis_def.get_user_loc(master)
+                if userLoc is None:
+                    # May happen if the custom parameter is disabled
```
anthrotype, line 205: the method `get_user_loc` can never return None, see how at line 367 the variable is initialized to something that is not None:

user_loc = self.default_user_loc

Even if the master custom parameter is missing or disabled, it will still return something. Please remove this `if userLoc is None:` bloc

Lib/glyphsLib/builder/axes.py
```diff
@@ -165,40 +192,46 @@ def to_designspace_axes(self):
...
                 if userLoc in mapping and mapping[userLoc] != designLoc:
                     logger.warning(
-                        "Axis location (%s) was redefined by '%s'", userLoc, master.name
+                        f"Axis {axis_def.tag}: Master '{master.name}' redefines "
+                        f"the mapping for user location {userLoc} "
+                        f"from {mapping[userLoc]} to {designLoc}"
                     )
                 mapping[userLoc] = designLoc
 
+            update_mapping_from_instances(
```
anthrotype, line 215: maybe add a comment why in here we do `cp_only=True` for the instances

anthrotype:
you added a test file but I don't see that used in any tests, I suppose you're working on that. Ideally I prefer when test data for some unit tests is defined in the same test module as close as possible to where it is being tested, as opposed to checking in a big source file that the test loads from disk, which no human is ever going to parse.. Unless we can't avoid it, in which case please try make the test data file as minimalist as possible, removing all the garbage. Thanks

jenskutilek:
Yep, I'm working on it. The test glyphs file will disappear.

tests/builder/axes_test.py
```diff
@@ -342,6 +342,145 @@ def test_mapping_using_axis_location_custom_parameter_on_instances(ufo_module):
     assert doc.axes[1].map == [(50, 224), (60, 384), (100, 448)]
 
 
+def test_mapping_using_axis_location_cp_on_masters_none(ufo_module):
+    # https://github.com/googlefonts/glyphsLib/issues/714
+    # https://github.com/googlefonts/glyphsLib/pull/810
+
+    # If custom parameters are disabled, they return None.
+    # The master Axis Location CP are disabled, but the ones on the instances
```
anthrotype, line 350: I'm curious, how common would be to have this Axis Location custom parameter in a master _and_ have it marked as "disabled"... Why would one want to do that? Wouldn't it be sufficient to not have it there/delete it?
I ask because you added two tests that specifically test for that None-ness

tests/builder/axes_test.py
```diff
@@ -342,6 +342,145 @@ def test_mapping_using_axis_location_custom_parameter_on_instances(ufo_module):
     assert doc.axes[1].map == [(50, 224), (60, 384), (100, 448)]
 
 
+def test_mapping_using_axis_location_cp_on_masters_none(ufo_module):
+    # https://github.com/googlefonts/glyphsLib/issues/714
+    # https://github.com/googlefonts/glyphsLib/pull/810
+
+    # If custom parameters are disabled, they return None.
```
anthrotype, line 349: I think the customParameters proxy object always returns None when a custom parameter that is looked up is not actually there (instead of throwing KeyError like a dict would normally do). So I'm not exactly sure what you're trying to do with your setting them explicitly to a `None` value below.

anthrotype: The comment should maybe say "If custom parameters are missing, they return None."

tests/builder/axes_test.py
jenskutilek: I set them to "disabled" all the time, to compare versions or test something. But I wasn't aware that it always returns None, no matter if it is disabled or not there at all. Will adapt the code.

jenskutilek:
CI/CD – I love but hate them. I have no idea why the test does fail here now, and on my local machine it is also broken again in a different way:

```
$ tox
.package create: /Users/jens/Code/glyphsLib/.tox/.package
.package installdeps: setuptools>=45, wheel, setuptools_scm[toml]>=6.0
/Users/jens/Code/glyphsLib/.tox/.package/lib/python3.10/site-packages/setuptools/config/setupcfg.py:508: SetuptoolsDeprecationWarning: The license_file parameter is deprecated, use license_files instead.
  warnings.warn(msg, warning_class)
/Users/jens/Code/glyphsLib/.tox/.package/lib/python3.10/site-packages/setuptools/config/setupcfg.py:508: SetuptoolsDeprecationWarning: The license_file parameter is deprecated, use license_files instead.
  warnings.warn(msg, warning_class)
lint create: /Users/jens/Code/glyphsLib/.tox/lint
lint installdeps: -rrequirements-dev.txt
lint installed: appdirs==1.4.4,attrs==21.4.0,black==22.3.0,booleanOperations==0.9.0,cffsubr==0.2.9.post1,click==8.1.2,coverage==6.3.2,cu2qu==1.6.7.post1,defcon==0.10.0,execnet==1.9.0,flake8==4.0.1,flake8-bugbear==22.3.23,fonttools==4.34.4,fs==2.4.15,importlib-metadata==4.2.0,iniconfig==1.1.1,lxml==4.8.0,mccabe==0.6.1,mypy-extensions==0.4.3,packaging==21.3,pathspec==0.9.0,platformdirs==2.5.2,pluggy==1.0.0,py==1.11.0,pyclipper==1.3.0.post2,pycodestyle==2.8.0,pyflakes==2.4.0,pyparsing==3.0.8,pytest==7.1.1,pytest-forked==1.4.0,pytest-randomly==3.11.0,pytest-xdist==2.5.0,pytz==2022.1,six==1.16.0,skia-pathops==0.7.2,tomli==2.0.1,typed-ast==1.5.3,typing_extensions==4.2.0,ufo2ft==2.27.0,ufonormalizer==0.6.1,unicodedata2==14.0.0,xmldiff==2.4,zipp==3.8.0
lint run-test-pre: PYTHONHASHSEED='2285796586'
lint run-test: commands[0] | black --check --diff Lib tests
All done! ✨ 🍰 ✨
82 files would be left unchanged.
lint run-test: commands[1] | flake8 Lib tests
py37 create: /Users/jens/Code/glyphsLib/.tox/py37
SKIPPED: InterpreterNotFound: python3.7
py38 create: /Users/jens/Code/glyphsLib/.tox/py38
py38 installdeps: -rrequirements.txt, -rrequirements-dev.txt
py38 inst: /Users/jens/Code/glyphsLib/.tox/.tmp/package/1/glyphsLib-6.1.1.dev28+g3fae8dc4.d20221010.tar.gz
py38 installed: appdirs==1.4.4,attrs==21.4.0,black==22.3.0,booleanOperations==0.9.0,cffsubr==0.2.9.post1,click==8.1.2,coverage==6.3.2,cu2qu==1.6.7.post1,defcon==0.10.0,execnet==1.9.0,flake8==4.0.1,flake8-bugbear==22.3.23,fonttools==4.34.4,fs==2.4.15,glyphsLib @ file:///Users/jens/Code/glyphsLib/.tox/.tmp/package/1/glyphsLib-6.1.1.dev28%2Bg3fae8dc4.d20221010.tar.gz,importlib-metadata==4.2.0,iniconfig==1.1.1,lxml==4.8.0,mccabe==0.6.1,mypy-extensions==0.4.3,openstep-plist==0.3.0,packaging==21.3,pathspec==0.9.0,platformdirs==2.5.2,pluggy==1.0.0,py==1.11.0,pyclipper==1.3.0.post2,pycodestyle==2.8.0,pyflakes==2.4.0,pyparsing==3.0.8,pytest==7.1.1,pytest-forked==1.4.0,pytest-randomly==3.11.0,pytest-xdist==2.5.0,pytz==2022.1,six==1.16.0,skia-pathops==0.7.2,tomli==2.0.1,typed-ast==1.5.3,typing_extensions==4.2.0,ufo2ft==2.27.0,ufoLib2==0.13.1,ufonormalizer==0.6.1,unicodedata2==14.0.0,xmldiff==2.4,zipp==3.8.0
py38 run-test-pre: PYTHONHASHSEED='2285796586'
py38 run-test: commands[0] | coverage run --parallel-mode -m pytest
=============================================================================== test session starts ================================================================================
platform darwin -- Python 3.8.8, pytest-7.1.1, pluggy-1.0.0
cachedir: .tox/py38/.pytest_cache
Using --randomly-seed=2990034994
rootdir: /Users/jens/Code/glyphsLib, configfile: pyproject.toml
plugins: xdist-2.5.0, forked-1.4.0, randomly-3.11.0
collected 0 items / 1 error

====================================================================================== ERRORS ======================================================================================
__________________________________________________________________________ ERROR collecting test session ___________________________________________________________________________
.tox/py38/lib/python3.8/site-packages/_pytest/config/__init__.py:607: in _importconftest
    mod = import_path(conftestpath, mode=importmode, root=rootpath)
.tox/py38/lib/python3.8/site-packages/_pytest/pathlib.py:556: in import_path
    raise ImportPathMismatchError(module_name, module_file, path)
E   _pytest.pathlib.ImportPathMismatchError: ('tests.conftest', '/Users/jens/Code/glyphsLib/tests/conftest.py', PosixPath('/Users/jens/Code/glyphsLib/glyphsLib-5.2.1.dev224+g9271c32e/tests/conftest.py'))
============================================================================= short test summary info ==============================================================================
ERROR  - _pytest.pathlib.ImportPathMismatchError: ('tests.conftest', '/Users/jens/Code/glyphsLib/tests/conftest.py', PosixPath('/Users/jens/Code/glyphsLib/glyphsLib-5.2.1.dev224...
!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!! Interrupted: 1 error during collection !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
================================================================================= 1 error in 0.37s =================================================================================
/Users/jens/Code/glyphsLib/.tox/py38/lib/python3.8/site-packages/coverage/inorout.py:519: CoverageWarning: Module glyphsLib was never imported. (module-not-imported)
  self.warn(f"Module {pkg} was never imported.", slug="module-not-imported")
/Users/jens/Code/glyphsLib/.tox/py38/lib/python3.8/site-packages/coverage/control.py:793: CoverageWarning: No data was collected. (no-data-collected)
  self._warn("No data was collected.", slug="no-data-collected")
ERROR: InvocationError for command /Users/jens/Code/glyphsLib/.tox/py38/bin/coverage run --parallel-mode -m pytest (exited with code 2)
```

anthrotype:
the tests didn't fail, codecov faild to upload the results, maybe we need to do the same thing I did in fonttools for https://github.com/fonttools/fonttools/issues/2763

try to run with -r option to recrete the local envs

jenskutilek:
> try to run with -r option to recrete the local envs

Alas, that doesn't fix it, and it is broken only on one of my computers :-/

anthrotype:
regarding the codecov failures, can you please try to rebase on `main`, I just pushed a commit that I hope should fix that: https://github.com/googlefonts/glyphsLib/commit/1974c4425b4ba588f00f12c4d5cfb5f45ca982a3

jenskutilek:
Ok, all tests pass now

anthrotype:
Thanks. Can you please summarize in a sentence or two what this PR changes are so I can include in the next release when we cut that?

PR #812, opened 2022-09-29
exclude GSInstance.type == "variable" when building DS axis maps

yanone:
Glyphs 3 implemented a new type of instance for VF settings that sits alongside the known type of "normal" instances.

This messed with the `instance_mapping` creation in `axes.py` as it would overwrite `designLoc` back to `0` for the `Regular` instance, causing the designspace to have erroneous default locations.

This PR now excludes the "variable" instance from being processed for the `instance_mapping` creation.

Edit: This is kind of urgent, as a new version of Cairo Play needs to be published. I can't temporarily use this branch for the font because I already have to use other temporary branches in order to be able to generate RTL kerning. Issues are piling up for Arabic. This PR here is a quick one.

### DIFF
--- a/Lib/glyphsLib/builder/axes.py
+++ b/Lib/glyphsLib/builder/axes.py
@@ -184,7 +184,9 @@ def to_designspace_axes(self):
             # a user location and a design location.
             instance_mapping = {}
             for instance in self.font.instances:
-                if is_instance_active(instance) or self.minimize_glyphs_diffs:
+                if (
+                    is_instance_active(instance) or self.minimize_glyphs_diffs
+                ) and instance.type != "variable":
                     designLoc = axis_def.get_design_loc(instance)
                     userLoc = axis_def.get_user_loc(instance)
                     if (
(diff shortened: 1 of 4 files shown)


anthrotype (approved):

anthrotype:
(ideally we should have a test)

yanone:
I added a test

anthrotype (approved):

anthrotype:
wait a second, did you check in the entire CairoPlay source file as the VariableInstances.glyphs test file? I see it contains almost 3K lines. It doesn't need to be a whole file, you should be able to reproduce the issue with a minimal source. Can you please update that?

yanone:
I had already deleted all glyphs. It was only 50KB, mostly feature code. After deleting the feature code it comes down to 22KB. I can do that. But do I push it to the branch or directly to main now that it's merged?

anthrotype:
I think it's still a lot of bytes wasted.. It's not the actual file size that matter, but the noise in when reviewing the tests, the fact that the relevant details get obscured in a wall of text. 
So can you please reduce it to the bare minimum to be able to test this particular feature? I presumed this was the case already, but let's establish this as a general rule for any coming features. 
You can open a new PR.

Issue #813, opened 2022-09-30
Caution: GSInstance.type == "variable"


yanone:
I just submitted a [PR](https://github.com/googlefonts/glyphsLib/pull/812) that implements `GSInstance.type`.

Glyphs 3 unexpectedly implemented a "variable" instance that is a pseudo instance for VF settings.
In my particular case, this messed with the `instance_mapping` as that instance would overwrite the `designLoc` back to 0, preventing correct entry of an axis’ default location. The PR fixes that.

I'm writing this issue as a heads-up about other fallout from this new instance type.

For instance, it shows up in designspace documents as a separate instance. Interestingly, the instance won't end up in the `fvar` table of an actual font. There may be code in place to prevent a duplicate instance (the new instance has a name of "Regular") from making it into the font. But in my case the "variable" instance is located at the end of the instance list. What if it's at the beginning and the rightful Regular instance would get ignored instead?

The extent of fallout of this is unfathomable for me. Someone else needs to dive into this. What I know is that Glyphs 3 VF sources that I get on my table increasingly include this "variable" instance for settings.

anthrotype:
thanks for the heads up. Maybe it's simply a matter of filtering those out somewhere here so they don't end up in the designspace instances:

https://github.com/googlefonts/glyphsLib/blob/d93c234f03a6b7452931151797ec8ec50227078e/Lib/glyphsLib/builder/instances.py#L47-L54

but we also need to think about:

- how these new psudo-instances map to designspace v5 concepts of a variable-font target /cc @belluz
- how we should round-trip these when going .glyphs => ufo

yanone:
There's more: The `Variable Font Origin` custom parameter can be defined in two different places (once in the font tab of the Font Info and once in that "variable" instance). And the same is probably true for all other custom parameters.

Here’s the thing: They actually get saved into the source file twice in two locations, and you can give them two separate values for maximum ambiguity.

anthrotype:
I hope @schriftgestalt can help us better understand how this is supposed to work.

anthrotype:
for reference, this other PR from @jenskutilek https://github.com/googlefonts/glyphsLib/pull/786 is also related to this issue

jenskutilek:
> There's more: The `Variable Font Origin` custom parameter can be defined in two different places (once in the font tab of the Font Info and once in that "variable" instance). And the same is probably true for all other custom parameters.
> 
[...]

PR #815, opened 2022-10-05
VF settings fixup for #812/#813

jenskutilek:
Small fixup for #812 and #813.

- Use constants in API, as Glyphs does (see https://docu.glyphsapp.com/#instance-types)
- Check against constant when deciding to include an "instance" (i.e. VF setting) in the axis mappings

### DIFF
--- a/Lib/glyphsLib/classes.py
+++ b/Lib/glyphsLib/classes.py
@@ -3300,6 +3300,7 @@ GSInstance._add_parsers(
         {"plist_name": "axesValues", "object_name": "axes"},
         {"plist_name": "manualInterpolation", "converter": bool},
         {"plist_name": "properties", "type": GSFontInfoValue},
+        {"plist_name": "type", "converter": bool},
     ]
 )
 
--- a/Lib/glyphsLib/classes.py
+++ b/Lib/glyphsLib/classes.py
@@ -20,6 +20,7 @@ import os
 import re
 import uuid
 from collections import OrderedDict
+from enum import IntEnum
 from io import StringIO
 
 import openstep_plist
@@ -141,9 +142,12 @@ CURVE = "curve"
 OFFCURVE = "offcurve"
 QCURVE = "qcurve"
 
+
 # Instance types; normal instance or variable font setting pseudo-instance
-INSTANCETYPESINGLE = 0
-INSTANCETYPEVARIABLE = 1
+class InstanceType(IntEnum):
+    SINGLE = 0
+    VARIABLE = 1
+
 
 TAG = -2
 TOPGHOST = -1
@@ -242,8 +246,9 @@ WIDTH_CODES = {
 }
 
 
-def instance_type_converter(value):
-    return {"variable": INSTANCETYPEVARIABLE}.get(value, INSTANCETYPESINGLE)
+def instance_type(value):
+    # Convert the instance type from the plist ("variable") into the integer constant
+    return getattr(InstanceType, value.upper())
 
 
 class OnlyInGlyphsAppError(NotImplementedError):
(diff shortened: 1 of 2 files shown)


Lib/glyphsLib/classes.py
```diff
@@ -141,6 +141,9 @@
 OFFCURVE = "offcurve"
 QCURVE = "qcurve"
 
+INSTANCETYPESINGLE = 0
+INSTANCETYPEVARIABLE = 1
```
anthrotype, line 145: should this be defined in https://github.com/googlefonts/glyphsLib/blob/main/Lib/glyphsLib/builder/constants.py maybe? The same probably applies for the other constants defined in here, at least those that are imported by other modules and not private to the current one.

Lib/glyphsLib/classes.py
```diff
@@ -3300,6 +3300,7 @@ def customValue3(self, value):
         {"plist_name": "axesValues", "object_name": "axes"},
         {"plist_name": "manualInterpolation", "converter": bool},
         {"plist_name": "properties", "type": GSFontInfoValue},
+        {"plist_name": "type", "converter": bool},
```
anthrotype, line 3303: is it a bool or an int? (more like an IntEnum)

jenskutilek: In the plist (Glyphs file) it is given as `type = variable`, or not at all. I figured interpreting the string as bool is the easiest way to interpret the value in sync with the constants INSTANCETYPESINGLE (0)/INSTANCETYPEVARIABLE (1).

I'm not familiar with IntEnum, how would the conversion be done?

Lib/glyphsLib/classes.py
jenskutilek: Moving it there leads to circular imports. It's not immediately obvious to me how to avoid that ...

Lib/glyphsLib/classes.py
anthrotype: these two constants, INSTANCETYPESINGLE and INSTANCETYPEVARIABLE, _what_ are they in the Glyphs Python API? Are they integers? Strings?

if one does `type(instance.type)` what do you get?

IIUC, when serializing to plist, the the type is either omitted when == default INSTANCETYPESINGLE, or set to "variable" for INSTANCETYPEVARIABLE.

Depending on what the INSTANCETYPESINGLE/VARIABLE constants are, whether int or str, we can either use an IntEnum or a StrEnum: i.e. an Enum mixed in with an int or a str type which compares equal to those and can be used as drop-in replacement, see https://docs.python.org/3.11/library/enum.html#enum.IntEnum and https://docs.python.org/3.11/library/enum.html#enum.StrEnum) 

StrEnum was only added with 3.11 to the enum module, but one can achieve the same with 

```
from enum import Enum
class StrEnum(str, Enum):
    __str__ == str.__str__

class InstanceType(StrEnum):
    SINGLE = "single"
    VARIABLE = "variable"

assert InstanceType("variable") == InstanceType.VARIABLE == "variable"

with pytest.raises(ValueError):
    InstanceType("unknown")
```

anthrotype: If the constants are integers, we can use IntEnum like so:

```python
from enum import IntEnum

class InstanceType(IntEnum):
    SINGLE = 0
    VARIABLE = 0

assert InstanceType.SINGLE == InstanceType(0) == 0
assert InstanceType.VARIABLE == InstanceType(1) == 1
with pytest.raises(ValueError):
    InstanceType(2)

assert InstanceType.VARIABLE.name.lower() == "variable"
```

schriftgestalt: > if one does type(instance.type) what do you get?

in Glyphs: 
```
print(type(instance.type))

>> <class 'int'>
```

schriftgestalt: I like the idea of using enums like this. I might adjust the Glyphs API in that direction.

anthrotype:
LGTM

PR #818, opened 2022-10-07
WIP: Glyph info 3

anthrotype:
I just created a PR from the 'GlyphInfo3' branch. @yanone @schriftgestalt what's the status of this?
Can you give this a better PR title and add a description of why/what this is about? thanks

related to https://github.com/googlefonts/glyphsLib/pull/778 and https://github.com/googlefonts/ufo2ft/issues/658


yanone:
The `GlyphInfo3` branch is Georg’s attempt to update `glyphsLib` to support glyph data as seen in Glyphs3, namely the writing direction (`glyph_info.direction = GSLTR/GSRTL`), which is necessary as a precursor to update `glyphsLib` to support RTL kerning from Glyphs3 sources, as those store RTL in a new attribute called `kerningRTL`.

https://github.com/googlefonts/glyphsLib/pull/778 is my attempt at this second step, but it's incomplete, or rather, needs to be rebased to the `GlyphInfo3` branch. I'd rather wait with any of that until this PR here is merged.

schriftgestalt:
I try to finish this in the next couple of days.

Lib/glyphsLib/glyphdata.py
```diff
@@ -21,21 +21,126 @@
...
+GSRTL = "RTL"
+GSVertical = 4
+
+
+def debug(*string):
+    # print(*string)
+    pass
+
+
+class GlyphInfo:
```
madig, line 60: Given that we use Python 3.7 as the minimum, GlyphsInfo can be made a https://devdocs.io/python~3.7/library/dataclasses, which saves some boilerplate code.

schriftgestalt:
I finished what I had planed to do.

madig:
Note to self:

1. Iterate over all names and ensure that I get the same info from glyphsLib as from Glyphs 3
2. Grab a bunch of .glyphs files and iterate over their names and test the same

Lib/glyphsLib/glyphdata.py
schriftgestalt: I tried that when I was working on the glyphData stuff. But I didn’t figure out how to implement the fallback of the production name property. And because of that property, the custom copy() method is still needed and so it saves not as much as it could.

schriftgestalt:
> Iterate over all names and ensure that I get the same info from glyphsLib as from Glyphs 3
when you iterate over all names in the xml file you basically compare the .xml files. More interesting (and potentially error prone) is the info construction from ligatures and composite glyphs.

madig:
> More interesting (and potentially error prone) is the info construction from ligatures and composite glyphs.

Could you maybe share the code? Did you reference it for making this PR

schriftgestalt:
The code in Glyphs and in glyphsLib is very different (the code in Glyphs is very old and does slightly more). But the behavior should be the same as shown by the tests. 

What I meant is that when you test the new code, one should focus in glyphs that are not in the glyphData and those are mostly ligatures. 

One thing I might need to explain:
There is a difference between `f_f.ss01` and `f_f._ss01`. The first glyph is a stylistic alternate of the `f_f` and the later is a ligature of `f` and `f.ss01`. So if the suffix has the same number of underscores, each substring between the underscores is the suffix for the corresponding glyph name, including empty strings.

yanone:
@anthrotype

I'm cleaning Georg’s work up a bit. The linting test passes now. Could you please take a look at the 3.7 test? `postscriptNames.get("A")` returns `uni0041` now instead of `None` which doesn't seem like a mistake to me.

And then there's the regression check. That's whole bunch.

anthrotype:
> postscriptNames.get("A") returns uni0041 now instead of None which doesn't seem like a mistake to me.

the test expects None because the CustomGlyphData.xml does not contain an entry for glyph 'A'. So something has changed in this PR that makes the test fail. I think we want to keep the current behaviour as regards custom glyph data.

schriftgestalt:
My code should set the production name to "A", not the uni-name. I added a test to make sure. So the new behavior seems to come somewhere else?

schriftgestalt:
Can this be revived? The RTL<>LTR stuff is needed for g3 files round tripping and the other improvements are also good to have.

PR #819, opened 2022-10-07
Vf settings fixup 2

jenskutilek:
#813 resp. its PR #812 did only skip "variable font setting" instances when building the axis mappings.

This PR actually keeps them out of the designspace file. I've also adapted the test.

### DIFF
--- a/tests/builder/axes_test.py
+++ b/tests/builder/axes_test.py
@@ -425,7 +425,9 @@ def test_variable_instance(ufo_module):
     """
     source_path = os.path.join("tests", "data", "VariableInstance.glyphs")
     font = GSFont(source_path)
+    assert len(font.instances) == 28
     doc = to_designspace(font)
 
     assert doc.axes[0].map[2] == (400, 80)
     assert doc.axes[0].default == 400
+    assert len(doc.instances) == 27
--- a/tests/builder/axes_test.py
+++ b/tests/builder/axes_test.py
@@ -425,9 +425,9 @@ def test_variable_instance(ufo_module):
     """
     source_path = os.path.join("tests", "data", "VariableInstance.glyphs")
     font = GSFont(source_path)
-    assert len(font.instances) == 28
+    assert len(font.instances) == 28  # Including the VF setting
     doc = to_designspace(font)
 
     assert doc.axes[0].map[2] == (400, 80)
     assert doc.axes[0].default == 400
-    assert len(doc.instances) == 27
+    assert len(doc.instances) == 27  # The VF setting should not be in the DS
--- a/Lib/glyphsLib/builder/instances.py
+++ b/Lib/glyphsLib/builder/instances.py
@@ -19,7 +19,7 @@ import logging
 from fontTools.varLib.models import piecewiseLinearMap
 
 from glyphsLib.util import build_ufo_path
-from glyphsLib.classes import WEIGHT_CODES, GSCustomParameter
+from glyphsLib.classes import WEIGHT_CODES, GSCustomParameter, InstanceType
 from .constants import (
     UFO_FILENAME_CUSTOM_PARAM,
     EXPORT_KEY,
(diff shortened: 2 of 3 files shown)


tests/builder/axes_test.py
```diff
@@ -425,7 +425,9 @@ def test_variable_instance(ufo_module):
     """
     source_path = os.path.join("tests", "data", "VariableInstance.glyphs")
     font = GSFont(source_path)
+    assert len(font.instances) == 28
     doc = to_designspace(font)
 
     assert doc.axes[0].map[2] == (400, 80)
     assert doc.axes[0].default == 400
+    assert len(doc.instances) == 27
```
anthrotype, line 433: a comment as to why there were 28 and now 1 less would be nice, for the casual passer-by

jenskutilek:
Filtering them out is [what you suggested](https://github.com/googlefonts/glyphsLib/issues/813#issuecomment-1263283684). Keeping them in the designspace in the current form will lead to additional invalid or at best, duplicate instances in a VF, and going back from DS to glyphs, there is no way to recognize them and act accordingly.

anthrotype:
yes, they they should not end up among the designspace instances when going .glyphs => DS, because they are not actually instances. But we do store additional private glyphs-specific stuff in the designspace lib (e.g. to minimize diffs when round-tripping), we could do the same here for the time being. Until we figure out a way to map between Glyphs.app instances-as-variable-type and DSv5 variable-font elements.

tests/builder/axes_test.py
jenskutilek: Done.

anthrotype:
In any case, ignoring the existence of these instance.type="variable" may not be the right approach even when building.. First of all, they may contain important metadata that one may want to incorporate in the fontmake's VF build (not sure exactly what, you know better than me). Secondly, because I suppose there may be multiple such variable-type instances. Which one should fontmake build? Probably the right answer is all of them, by using Designspace v5 `<variable-font>` elements.
https://fonttools.readthedocs.io/en/latest/designspaceLib/xml.html#variable-fonts-element

jenskutilek:
Yes sure, they should be fully supported, but in the short term they break VF builds by adding bogus instances, so a quick solution would be nice.

anthrotype:
e.g. in the font you're trying to build, what does that variable "instance" actually contains? Can fontmake safely ignore the stuff in it ?

anthrotype:
if fontmake doesn't need that to build a correct VF, a temporary solution would be to simply remove it from the source file? If, on the other hand, it serves some purpose, then we need to find out what that is, and try to support it.

schriftgestalt:
The info that is in the variable font setting will mostly be in the design space. When round tripping, that info can be sued to make a new vfs 

man’s there can be more that one of them. Each one would need it be converted into its own design space. Not sure how to round trip that. 

> Can fontmake safely ignore the stuff in it ?

probably not. There are vital settings in it. Most of them where custom parameters on the font level before (the ones with a ‘variable font’ prefix.

jenskutilek:
@schriftgestalt I've seen some Glyphs files with "Variable Font Origin" in the variable font setting. But it is not possible to add this via UI in the cutting edge versions. Is that supposed to be in there or not?

schriftgestalt:
It will be supported in 3.2.

anthrotype:
> Each one would need it be converted into its own design space.

but new DesignSpace version 5 allows to specify multiple `<variable-fonts>` each with specific axis configurations and settings, we should be able to write a single DSv5 from a .glyphs source containing multiple variable "instances"

schriftgestalt:
I didn't know about that new capability. Even better, then.

anthrotype:
yes, see

https://fonttools.readthedocs.io/en/latest/designspaceLib/index.html#version-5-0
https://fonttools.readthedocs.io/en/latest/designspaceLib/xml.html#variable-fonts-element

Issue #754, opened 2021-12-01
Problems going from UFOs to .glyphs using v3 file format


DavidRaymond:
Using v6.0.0 I've been looking at Glyphs v3 file format support.  

I see that ufo2glyphs does not currently have an option to output in v3 file format.  I created an edited version to test with by adding `font.format_version = 3` prior to the `font.save` section and successfully created a v3 format .glyphs file from a design space file.  Is that the correct approach to use to create v3 format files?

Round-tripping using that and then glyphs2ufo produced many differences.  Some examples are:
- postscriptBlueValues, postscriptStemSnapH and postscriptStemSnapV removed from fontinfo.plist
- styleName in fontinfo.plist changed to Regular for all styles
- com.schriftgestaltung.weight in lib.plist changed to Regular for all styles
- public.verticalOrigin added to the lib of all .glif files

Note with the postscript keys, this data does end up in a .glyphs file if it is produced by saving from the Glyphs App - for example blue values end up in Alignment Zones.

anthrotype:
thanks for reporting this. Yes, we should add an option to set format_version to 3 for ufo2glyphs.
Regarding the difference you see when roundtripping, at least the first three look like bugs which we should fix; the last one about verticalOrigin I'm not sure. What value do you get? Did the original ufo glyphs contain public.verticalOrigin in their lib?

DavidRaymond:
The original .glifs did not have public.verticalOrigin.  It was added with a value of 2100 - which is the same value as ascender in fontinfo.

DavidRaymond:
I've done some further testing on this, and discovered that the addition of verticalOrigin is not connected with v3 format support - it happens round-tripping using the standard glyphs2ufo as well.

DavidRaymond:
To reproduce the other changes I'd reported:

- Clone https://github.com/silnrsi/font-andika-mtihani
- Round-trip the designspace files in `source/` using standard ufo2glyphs and glyphs2ufo and save the resulting UFOs
- Edit ufo2glyphs in cli.py and add `font.format_version = 3` after `font.disablesAutomaticAlignment = options.enable_automatic_alignment`
- Repeat the round-trip and compare the new UFOs with those saved above

Note that there will be other changes to com.schriftgestaltung... values in addition to those mentioned in my original report.  Those reported were just examples.

arrowtype:
[...]

PR #825, opened 2022-10-17
Implements part of #754 via getter/setters for proxying the file format 3 attributes

kontur:
This implements writing `postscriptBlueValues`, `postscriptOtherBlues`, `postscriptStemSnapV` and `postscriptStemSnapH` for glyphs > ufo.

I added [./tests/glyphs3_test.py](tests) and other tests are passing as well.

I'm not sure how elegant or clumsy it is to have those `GSFontMaster._...` underscore internals and accessing them via the getter/setter proxies, but it seemed the most practical way synthesize them from the `GSFont` attributes required for the offsets.

There are [this](https://github.com/kontur/glyphsLib/blob/fc2b551ddd02da4df0f0b2290fc526a554108847/Lib/glyphsLib/classes.py#L1765), [this](https://github.com/kontur/glyphsLib/blob/fc2b551ddd02da4df0f0b2290fc526a554108847/Lib/glyphsLib/classes.py#L1799) & [this](https://github.com/kontur/glyphsLib/blob/fc2b551ddd02da4df0f0b2290fc526a554108847/Lib/glyphsLib/classes.py#L1819) rather blunt stop gaps in the getters which mainly function to return nothing for cases where the masters are constructed via code, e.g. in the tests (about 50 or so fail otherwise). In a way this could mask a real issue as well, but not sure how realistic it is to rewrite the tests, or add some kind of "defaulting" structures to keep the indexes for constructed objects in sync between GSFont and GSFontMaster.

Should I squash my commits into one? The first commit attempts were a bit hit and miss...

### DIFF
--- a/Lib/glyphsLib/classes.py
+++ b/Lib/glyphsLib/classes.py
@@ -1445,6 +1445,8 @@ class GSAlignmentZone(GSBase):
         )
 
     def __lt__(self, other):
+        if not hasattr(other, "position") or not hasattr(other, "size"):
+            return False
         return (self.position, self.size) < (other.position, other.size)
 
     def __eq__(self, other):
@@ -1760,8 +1762,8 @@ class GSFontMaster(GSBase):
 
     @property
     def alignmentZones(self):
-        # If there are values that are parsed from file format 2 or explicitly
-        # set return those
+        # If there are values that are parsed from file format 2 or which are
+        # explicitly set return those
         if self._alignmentZones is not None:
             return self._alignmentZones
 
@@ -1786,12 +1788,21 @@ class GSFontMaster(GSBase):
         return zones
 
     @alignmentZones.setter
-    def alignmentZones(self, value):
-        assert type(value) == list
+    def alignmentZones(self, entries):
+        if type(entries) is not list:
+            raise ValueError(
+                "alignmentZones expected as list, got %s (%s)"
+                % (entries, type(entries))
+            )
         zones = []
-        for val in value:
-            if val not in zones and val is not False:
-                zones.append(val)
+        for zone in entries:
+            if type(zone) is not tuple and type(zone) is not GSAlignmentZone:
+                raise ValueError(
+                    "alignmentZones values expected as tuples of (pos, size) "
+                    "or GSAligmentZone, got: %s (%s)" % (zone, type(zone))
+                )
+            if zone not in zones:
+                zones.append(zone)
         self._alignmentZones = zones
 
     @property
--- a/Lib/glyphsLib/classes.py
+++ b/Lib/glyphsLib/classes.py
@@ -1445,13 +1445,13 @@ class GSAlignmentZone(GSBase):
         )
 
     def __lt__(self, other):
-        if not hasattr(other, "position") or not hasattr(other, "size"):
-            return False
+        if not isinstance(other, GSAlignmentZone):
+            return NotImplemented
         return (self.position, self.size) < (other.position, other.size)
 
     def __eq__(self, other):
-        if not hasattr(other, "position") or not hasattr(other, "size"):
-            return False
+        if not isinstance(other, GSAlignmentZone):
+            return NotImplemented
         return (self.position, self.size) == (other.position, other.size)
 
     def plistValue(self, format_version=2):
@@ -1789,15 +1789,15 @@ class GSFontMaster(GSBase):
 
     @alignmentZones.setter
     def alignmentZones(self, entries):
-        if type(entries) is not list:
-            raise ValueError(
+        if not isinstance(entries, tuple) and not isinstance(entries, list):
+            raise TypeError(
                 "alignmentZones expected as list, got %s (%s)"
                 % (entries, type(entries))
             )
         zones = []
         for zone in entries:
-            if type(zone) is not tuple and type(zone) is not GSAlignmentZone:
-                raise ValueError(
+            if not isinstance(zone, tuple) and not isinstance(zone, GSAlignmentZone):
+                raise TypeError(
                     "alignmentZones values expected as tuples of (pos, size) "
                     "or GSAligmentZone, got: %s (%s)" % (zone, type(zone))
                 )
--- a/Lib/glyphsLib/classes.py
+++ b/Lib/glyphsLib/classes.py
@@ -1447,6 +1447,11 @@ class GSAlignmentZone(GSBase):
     def __lt__(self, other):
         return (self.position, self.size) < (other.position, other.size)
 
+    def __eq__(self, other):
+        if not hasattr(other, "position") or not hasattr(other, "size"):
+            return False
+        return (self.position, self.size) == (other.position, other.size)
+
     def plistValue(self, format_version=2):
         return '"{{{}, {}}}"'.format(
             floatToString5(self.position), floatToString5(self.size)
@@ -1574,30 +1579,36 @@ class GSFontMaster(GSBase):
         "ascender": 800,
         "descender": -200,
         "italic angle": 0,
+        "baseline": 0,
     }
 
     _axis_defaults = (100, 100)
 
     def _parse_alignmentZones_dict(self, parser, text):
+        """
+        For glyphs file format 2 this parses the aligmentZone parameter
+        directly. For file format 3 alignmentZones will be infered from
+        metricValues.
+        """
         _zones = parser._parse(text, str)
-        self.alignmentZones = [GSAlignmentZone().read(x) for x in _zones]
+        self._alignmentZones = [GSAlignmentZone().read(x) for x in _zones]
 
     def __init__(self):
         self._customParameters = []
         self._name = None
         self._userData = None
-        self.alignmentZones = []
+        self._alignmentZones = None
+        self._horizontalStems = None
+        self._verticalStems = None
         self.axes = list(self._axis_defaults)
         self.metrics = []
         self.customName = ""
         self.font = None
         self.guides = []
-        self.horizontalStems = 0
         self.iconName = ""
         self.id = str(uuid.uuid4()).upper()
         self.numbers = []
         self.stems = []
-        self.verticalStems = 0
         self.visible = False
         self.weight = self._defaultsForName["weight"]
         self.width = self._defaultsForName["width"]
@@ -1746,6 +1757,82 @@ class GSFontMaster(GSBase):
             # Pad array with ... zeroes?
             self.metrics.append(GSMetricValue(position=0))
         self.metrics[metricIndex] = GSMetricValue(position=value)
+
+    @property
+    def alignmentZones(self):
+        # If there are values that are parsed from file format 2 or explicitly
+        # set return those
+        if self._alignmentZones is not None:
+            return self._alignmentZones
+
+        if len(self.metrics) == 0:
+            return []
+
+        validZoneMetrics = (
+            "ascender",
+            "cap height",
+            "x-height",
+            "baseline",
+            "descender",
+        )
+
+        zones = []
+        for index, fontMetric in enumerate(self.font.metrics):
+            if fontMetric.type not in validZoneMetrics:
+                continue
+            metric = self.metrics[index]
+            zone = GSAlignmentZone(pos=metric.position, size=metric.overshoot)
+            zones.append(zone)
+        return zones
+
+    @alignmentZones.setter
+    def alignmentZones(self, value):
+        assert type(value) == list
+        zones = []
+        for val in value:
+            if val not in zones and val is not False:
+                zones.append(val)
+        self._alignmentZones = zones
+
+    @property
+    def horizontalStems(self):
+        if self._horizontalStems is not None:
+            return self._horizontalStems
+
+        if not hasattr(self.font, "stems"):
+            return []
+
+        horizontalStems = []
+        for index, font_stem in enumerate(self.font.stems):
+            if not font_stem.horizontal:
+                continue
+            horizontalStems.append(self.stems[index])
+        return horizontalStems
+
+    @horizontalStems.setter
+    def horizontalStems(self, value):
+        assert type(value) == list
+        self._horizontalStems = value
+
+    @property
+    def verticalStems(self):
+        if self._verticalStems is not None:
+            return self._verticalStems
+
+        if not hasattr(self.font, "stems"):
+            return []
+
+        verticalStems = []
+        for index, font_stem in enumerate(self.font.stems):
+            if font_stem.horizontal:
+                continue
+            verticalStems.append(self.stems[index])
+        return verticalStems
+
+    @verticalStems.setter
+    def verticalStems(self, value):
+        assert type(value) == list
+        self._verticalStems = value
 
     @property
     def ascender(self):
--- a/Lib/glyphsLib/classes.py
+++ b/Lib/glyphsLib/classes.py
@@ -1581,7 +1581,6 @@ class GSFontMaster(GSBase):
         "ascender": 800,
         "descender": -200,
         "italic angle": 0,
-        "baseline": 0,
     }
 
     _axis_defaults = (100, 100)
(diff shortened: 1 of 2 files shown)


kontur:
Any chance to get this reviewed? :)

tests/glyphs3_test.py
```diff
@@ -22,3 +22,24 @@ def test_glyphspackage_load(datadir):
...
+
+def test_glyphs3_alignment_zones(datadir):
+    font = glyphsLib.load(str(datadir.join("GlyphsUnitTestSans3.glyphs")))
+    master = font.masters[0]
+
+    assert master.alignmentZones != []
+
+    # As per classes_tests::GSAlignmentZoneFromFileTest
+    for i, zone in enumerate([(800, 10), (700, 10), (470, 10), (0, -10), (-200, -10)]):
+        pos, size = zone
```
anthrotype, line 35:
```suggestion
    for i, (pos, size) in enumerate([(800, 10), (700, 10), (470, 10), (0, -10), (-200, -10)]):
```

Lib/glyphsLib/classes.py
```diff
@@ -1744,6 +1755,82 @@ def _set_metric(self, metricname, value):
...
+            if fontMetric.type not in validZoneMetrics:
+                continue
+            metric = self.metrics[index]
+            zone = GSAlignmentZone(pos=metric.position, size=metric.overshoot)
+            zones.append(zone)
+        return zones
+
+    @alignmentZones.setter
+    def alignmentZones(self, value):
+        assert type(value) == list
```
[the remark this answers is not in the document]
kontur: Yes, makes sense! It actually expects a list of tuples; I'll add tests for this, too.

tests/glyphs3_test.py
kontur: Ah, nicer!

kontur:
Rebased on `main` again, but there appear to be a lot of unrelated tests failing. What's the deal with those, are those residue from merged in WIP and the respective features will still fix those, or is there an effort to just tackle them as separate task?

anthrotype:
> there appear to be a lot of unrelated tests failing.

the CI checks seem to be passing now, not sure what you're referring to: https://github.com/googlefonts/glyphsLib/actions/runs/3627890015

I'll take another review pass

anthrotype:
the regression tests failing is expected when adding new features from a PR branch, but once that's merged to main, they fix themselves (the regression test CI job compares output from PR branch with that from main branch).

One needs to manually check that the changes are as expected:
https://github.com/googlefonts/glyphsLib/actions/runs/3627890014/jobs/6119377961

anthrotype:
the CI also runs tox the same way as you do so that's unexpected.. let me try reproduce locally from a copy of your PR branch

anthrotype:
the regression tests failures seem all benign, I can see that postscriptBlueValues are now being compiled which is what this PR is about

kontur:
Ah, perfect, thanks for the `-r` hint, not very familiar with tox. Tests run locally without issues as well 🙇

Lib/glyphsLib/classes.py
```diff
@@ -1747,6 +1760,91 @@ def _set_metric(self, metricname, value):
...
+            "x-height",
+            "baseline",
+            "descender",
+        )
+
+        zones = []
+        for index, fontMetric in enumerate(self.font.metrics):
+            if fontMetric.type not in validZoneMetrics:
+                continue
+            metric = self.metrics[index]
```
anthrotype, line 1785: I'm a bit confused.. Where are the metrics (or the stems in a similar piece of code below) actually stored? Inside the GSFont.metrics or in this GSFontMaster.metrics?
Why do you have to iterate over the font.metrics if you then take the item from the master's metrics?

Lib/glyphsLib/classes.py
```diff
@@ -1747,6 +1760,91 @@ def _set_metric(self, metricname, value):
...
+            zone = GSAlignmentZone(pos=metric.position, size=metric.overshoot)
+            zones.append(zone)
+        return zones
+
+    @alignmentZones.setter
+    def alignmentZones(self, entries):
+        if type(entries) is not list:
+            raise ValueError(
```
anthrotype, line 1793: usually it's preferable to use isinstance instead of `type() is ...`, unless you really have a good reason to restrict the exact type. And raise TypeError instead of ValueError when the type is not as expected

Lib/glyphsLib/classes.py
anthrotype: ah ok it's explained in https://github.com/schriftgestalt/GlyphsSDK/blob/Glyphs3/GlyphsFileFormat/GlyphsFileFormatv3.md#differences-between-version-2

kontur:
@anthrotype yep, those type checks are better now, thanks for pointing that out.

regarding @schriftgestalt 's suggestion: I agree in general, but this would probably need a more orchestrated effort overall, no? it shouldn't just prioritise v3 handling for one attribute. I can rework this for the `alignmentZones` if that will otherwise prevent the PR from being merged. personally I just need to be able to work with v3 files, from releases preferably.

Lib/glyphsLib/classes.py
```diff
@@ -1444,6 +1444,11 @@ def __repr__(self):
     def __lt__(self, other):
         return (self.position, self.size) < (other.position, other.size)
 
+    def __eq__(self, other):
+        if not hasattr(other, "position") or not hasattr(other, "size"):
+            return False
```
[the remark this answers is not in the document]
anthrotype: ping @kontur

anthrotype:
does glyphs.app python API still provide getters like `alignmentZones` returning a list of [GSAlignmentZone](https://docu.glyphsapp.com/#GSAlignmentZone) objects for v3 as well as v2 fonts? If so, we could keep the old v2 API and file an issue to expose the new v3 API alongside it.

Lib/glyphsLib/classes.py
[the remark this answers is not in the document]
kontur: 👍 missed this one before

simoncozens:
I said the other day to Cosimo that we should just support the G3 model and have a separate bit of code to “upgrade” G2 files on load. Trying to manipulate both systems is making the code atrocious, and G3 is much cleaner. But this would be a major overhaul.

Lib/glyphsLib/classes.py
```diff
@@ -1574,30 +1581,36 @@ def _serialize_to_plist(self, writer):
         "ascender": 800,
         "descender": -200,
         "italic angle": 0,
+        "baseline": 0,
```
anthrotype, line 1584: what does this change have to do with this PR? what is it meant to fix?

Lib/glyphsLib/classes.py
```diff
@@ -1747,6 +1760,91 @@ def _set_metric(self, metricname, value):
...
+
+        if len(self.metrics) == 0:
+            return []
+
+        validZoneMetrics = (
+            "ascender",
+            "cap height",
+            "x-height",
+            "baseline",
+            "descender",
```
anthrotype, line 1778: where's this list of validZoneMetrics taken from? it seems constant, it should probably be defined at global scope, or in the constants module.

anthrotype: and why skip some of the font metric types as not "valid" when fetching a given master metric values? where is this logic encoded?

Lib/glyphsLib/classes.py
```diff
@@ -1747,6 +1760,91 @@ def _set_metric(self, metricname, value):
...
+        for index, fontMetric in enumerate(self.font.metrics):
+            if fontMetric.type not in validZoneMetrics:
+                continue
+            metric = self.metrics[index]
+            zone = GSAlignmentZone(pos=metric.position, size=metric.overshoot)
+            zones.append(zone)
+        return zones
+
+    @alignmentZones.setter
+    def alignmentZones(self, entries):
```
anthrotype, line 1791: looks like alignmentZones is a read-only (https://docu.glyphsapp.com/#GSFontMaster.alignmentZones) so maybe we should not add a setter.
I agree with @simoncozens when he wrote

> We should store things as v3 format internally to the object but provide accessors to get at information in a v2 way.

kontur: Some of the existing tests assign `alignmentZones` like this for constructing objects via code only (as opposed to actual conversion). Happy to drop the setter and tests.

Lib/glyphsLib/classes.py
kontur: Basically I went by what's in the UI (mind you I missed `bodyHeight`):

Screenshot 2022-12-16 at 21 09 41

Outputting the `GSMetric.type` of these objects of a file gives:

```
ascender
cap height
x-height
baseline
descender
italic angle
bodyHeight

```

The last empty line for type is what the UI has as "Custom" metric. This may or may not be an overshoot — presumably things with overshoot != 0 we want to treat as an alignmentZone as well?

I think when I did a "has no overshoot" exclusion instead of explicitly checking for the type _a lot_ of tests outright failed because the font metricValues does not match a given master's metricValues length (e.g. in all sorts of constructed test cases or programmatic access of those objects that don't adhere to the file format 3 referencing the font metricValues from the master by index). As far as I recall the `"baseline": 0` default above also addressed this.

Note sure what to do here, but I'll ponder it.

kontur:
Rebased and refactored the `alignmentZones` getter code filtering that determined what Metric is or is not an alignmentZone.

Lib/glyphsLib/classes.py
kontur: Updated in code below in forced push. Disregarding the "italic angle" type of metrics early was key to avoiding all those issues in tests that construct just masters. Gets rid of the explicit magic list and does not require the `baseline` default either. Thanks for poking holes @anthrotype :)

anthrotype: > ... because the font metricValues does not match a given master's metricValues length (e.g. in all sorts of constructed test cases or programmatic access of those objects that don't adhere to the file format 3 referencing the font metricValues from the master by index

if it's invalid that len(master.metricValues) != len(font.metrics) then it is those tests that would need to be fixed...

Lib/glyphsLib/classes.py
```diff
@@ -1747,6 +1759,87 @@ def _set_metric(self, metricname, value):
...
+
+        zones = []
+        for index, fontMetric in enumerate(self.font.metrics):
+            # Ignore the "italic angle" "metric", it is not an alignmentZone
+            if fontMetric.type == "italic angle":
+                continue
+            metric = self.metrics[index]
+            # Ignore metric without overshoot, it is not an alignmentZone
+            if metric.overshoot == 0:
+                continue
```
anthrotype, line 1780: can you confirm that Glyphs.app v3 also does the same (ignore 'italic angle' metric an those with overshoot=0) when you get a GSFont.alignmentZones?
What's an example of such a metric without overshoot?

Lib/glyphsLib/classes.py
anthrotype: If one wants to _set_ these alignment zones using Glyphs.app v3 scripting API (where alignmentZones is read-only), one would one do that exactly?
Does glyphsLib already support this? If not, what do we need to support it?

Lib/glyphsLib/classes.py
schriftgestalt: Yes. This is what Glyphs is doing.

Lib/glyphsLib/classes.py
kontur: Going by what I found in the code. The primary use case in the library seems to be [here](https://github.com/googlefonts/glyphsLib/blob/74aefb387a5a84659902b53fa805b74ff3d8ed0c/Lib/glyphsLib/builder/blue_values.py#L55) setting the alignmentZones when converting from UFO to glyphs. 

Also tests [here](https://github.com/googlefonts/glyphsLib/blob/74aefb387a5a84659902b53fa805b74ff3d8ed0c/tests/classes_test.py#L755), [here](https://github.com/googlefonts/glyphsLib/blob/74aefb387a5a84659902b53fa805b74ff3d8ed0c/tests/writer_test.py#L301), [here](https://github.com/googlefonts/glyphsLib/blob/74aefb387a5a84659902b53fa805b74ff3d8ed0c/tests/builder/builder_test.py#L495) and [here](https://github.com/googlefonts/glyphsLib/blob/74aefb387a5a84659902b53fa805b74ff3d8ed0c/tests/builder/builder_test.py#L504) — my assumption being that a new feature or refactor should not break existing tests and thus a setter needs to be supported.

Lib/glyphsLib/classes.py
kontur: The tests were correct, only when _not filtering_ things that are now returned by a master'a metrics which are not alignmentZones in format 3 those test broke because without knowledge of the master my previous iteration of the code could not filter those properly. Different approach in the updated code, essentially discarding the `italic angle` from being an alignmentZone outright, and further removing metrics which do not have an overshoot, e.g. do not form a bluezone because they would be a "line" only.

anthrotype: can a valid "italic angle" metric ever have overshoot != 0? I suppose not. And if not, then simply filtering out metrics with overshoot=0 (without special-casing italic angle itself) should accomplish the same thing.

Lib/glyphsLib/classes.py
anthrotype: i'm ok to keep a setter for alignmentZones, but I wonder maybe it should store them internally as v3 metricValues.
Say you go from UFO => glyphs and want to save as format_version=3, those zones will not be written out in the generated .glyphs file.
I think we should have a method to convert from v2 (alignmentZones and stems) to v3 metricValues and call that when saving a font and format_version=3.

anthrotype: but that setter only sets `self._alignmentZones`, which as far as I understand is _not_ written out to .glyphs source when format_version=3,

```python
        if writer.format_version == 2:
            writer.writeObjectKeyValue(self, "alignmentZones", "if_true")
```

anthrotype: so I presume that roundtripped glyphs file of yours has format version 2.. but when Glyphs.app v3 opens it up it automatically populate the metric values.
That's what we are missing here, a way to upgrade these metrics from v2 to v3. I am not saying that we should do within this same PR, just that this is something we would like to support

Lib/glyphsLib/classes.py
kontur: In
```
 for index, fontMetric in enumerate(self.font.metrics):
            # Ignore the "italic angle" "metric", it is not an alignmentZone
            if fontMetric.type == "italic angle":
                continue
            metric = self.metrics[index]
            # Ignore metric without overshoot, it is not an alignmentZone
            if metric.overshoot == 0:
                continue
            zone = GSAlignmentZone(pos=metric.position, size=metric.overshoot)
            zones.append(zone)
```

`self.font.metrics` is a [GSMetric](https://github.com/googlefonts/glyphsLib/blob/74aefb387a5a84659902b53fa805b74ff3d8ed0c/Lib/glyphsLib/classes.py#L1393) value which has no overshoot. One has to use the index of this  _font_ metric values to get the _master_ [GSMetricValue](https://github.com/googlefonts/glyphsLib/blob/74aefb387a5a84659902b53fa805b74ff3d8ed0c/Lib/glyphsLib/classes.py#L1408) value, and then discard things which have no overshoot. Filtering the _font_ metric values by type `italic angle` is necessary because just accessing `master.metrics[]` by index will not have a value for a _font_ metric that is an italic angle.

The format 3 handling of metric values referencing back up to the font metrics list is convoluted imo.

Lib/glyphsLib/classes.py
kontur: Ah, I see, didn't understand that's what's happening. Yes, correct. 

I'll have a look at that, either as part of this or separate. Follows the same obtuse logic of font metrics vs master metric 😄 Happy to adopt that if there is a more general v2 > v3 structure in place. That section of code you quoted, I suppose, is at the heart of what would need a more elegant and transparent file format version upgrade.

kontur: _How_ does one actually write something out in format 3? Tried to find tests that verify those other `self.format == 3` written things (e.g. [here onwards](https://github.com/googlefonts/glyphsLib/blob/24b4d340e4c82948ba121dcfe563c1450a8e69c9/Lib/glyphsLib/classes.py#L4234)), but could only find [this one](https://github.com/googlefonts/glyphsLib/blob/24b4d340e4c82948ba121dcfe563c1450a8e69c9/tests/writer_test.py#L1190). Am I completely off track or is there pretty little actually for outputting/writing/testing v3? Would testing the output of the writer in `writer_tests.py` the right spot to implement checking for a yet-to-be-done implementation of writing those metrics back to v3 files?

schriftgestalt: > but could only find [this one](https://github.com/googlefonts/glyphsLib/blob/24b4d340e4c82948ba121dcfe563c1450a8e69c9/tests/writer_test.py#L1190).

This looks like the way to do this. While writing out, check the format and what data you have and act accordingly.

kontur:
@anthrotype Ah sorry, must have missed the lint issues locally and didn't follow the updates of the test runner here. Fixed and rebased on `main` once more, gtg from my part.

kontur:
*patiently waiting*

anthrotype:
thanks for the patience

Issue #633, opened 2020-11-06
glyphsLib does not handle localized names in Glyphs


aaronbell:
I'm working on a font that has a localized name in Glyphs. While I see the key showing up in the UFO's lib file, it isn't being processed and put into the UFO fontinfo NameRecords. As such, the font that is generated out of fonttools lacks the proper localized name. 

Here is the relevant key from the lib file:
com.schriftgestaltung.customParameter.GSFont.localizedFamilyName
Japanese;しっぽりアンチック

Here is a sample Glyphs file that has a localized name:
[ShipporiAntiqueMKanaLantenOnly.glyphs.zip](https://github.com/googlefonts/glyphsLib/files/5503589/ShipporiAntiqueMKanaLantenOnly.glyphs.zip)

m4rc1e:
It seems GlyphsApp has two customParameters which allow users to create name records which are not (3, 1, 0x409) or (1, 0, 0); "Name Table Entry" and "localized*Name". Ideally, we should support both but perhaps a good starting point is to support "Name Table Entry" since it's more flexible?

Unfortunately both customParams are not so straight forward to implement. "Name Table Entry" is a string that we'll need to parse and for "localized*Name", we'll need to create a constant which maps the language names to langIDs, platformIDs etc. There's probably a lot of magic happening as well, for instance if you add a localizedFamilyName, it will create a localizedSubFamilyName even if it hasn't been specified.

- _localized*Name includes localizedFamilyName, localizedStyleName, localizedStyleMapFamilyName, localizedDesigner_
- _"Name Table Entry" string format, https://glyphsapp.com/downloads/handbook/Glyphs-Handbook-2.3.pdf p190_

aaronbell:
Thanks Marc! It sounds like I'll need to manually add those fields for now.

PR #827, opened 2022-11-07
[review] Add support for roundtripping Glyphs/UFO custom name table entries.

bramstein:
We have some Glyphs source files with custom name table entries that weren't preserved when converting to UFO. This PR parses the Glyphs `<nameid> <platform_id>? <encoding_id>? <language_id>?; <string>` format and creates UFO `openTypeNameRecords` (and vice versa).

I added a new test that covers all the different combinations the Glyphs name table entry format supports. In the case the Glyphs name table entry only specifies a name identifier and string default values are used for the other parameters. The default parameters are not removed when converting from UFO -> Glyphs, so roundtripping a Glyphs will not be literally identical but functionally identical.

This is a partial fix for #633.

### DIFF
--- a/tests/builder/custom_params_test.py
+++ b/tests/builder/custom_params_test.py
@@ -479,6 +479,75 @@ class SetCustomParamsTestBase(object):
         font = glyphsLib.to_glyphs([self.ufo])
         self.assertEqual(font.customParameters["meta Table"], glyphs_meta)
 
+    def test_name_table_entry(self):
+        self.font.customParameters.append(
+            GSCustomParameter("Name Table Entry", "1024; FOO; BAZ")
+        )
+        self.font.customParameters.append(
+            GSCustomParameter("Name Table Entry", "2048 1; FOO")
+        )
+        self.font.customParameters.append(
+            GSCustomParameter("Name Table Entry", "4096 1 2; FOO")
+        )
+        self.font.customParameters.append(
+            GSCustomParameter("Name Table Entry", "8192 1 2 3; FOO")
+        )
+        self.font.customParameters.append(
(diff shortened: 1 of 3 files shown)


anthrotype:
Thanks for the PR! Can you please rebase on main branch (or merge origin/main into yours), that should fix the setup error you're seeing in the "lint" job.

tests/builder/custom_params_test.py
```diff
@@ -479,6 +479,75 @@ def test_meta_table(self):
...
+        )
+        self.font.customParameters.append(
+            GSCustomParameter("Name Table Entry", "4096 1 2; FOO")
+        )
+        self.font.customParameters.append(
+            GSCustomParameter("Name Table Entry", "8192 1 2 3; FOO")
+        )
+        self.font.customParameters.append(
+            GSCustomParameter("Name Table Entry", "0x4000 074; BAZ")
+        )
```
anthrotype, line 497: can you also add a test using the 0-prefixed octal numbers?

anthrotype (approved):
LGTM, thanks!

tests/builder/custom_params_test.py
bramstein: The `074` is octal (for `60` decimal) :slightly_smiling_face:

anthrotype: oh! thanks 😅

anthrotype:
regression tests took ages, I'm not sure what's going on there. But if anything related to this PR has changed that made them fail, then it's for good and they _should_ automatically fix themselves after merging the PR to main. If not, we'll take it from there.
Thanks!

PR #828, opened 2022-11-07
Support Glyphs 3 Alternate Layers

### LINKED ISSUE #741: Intermediate/brace layers are not supported anymore

simoncozens:
This PR implements support for Glyphs 3 alternate layers:
* Alternates whose range is bounded both at the top and at the bottom.
* Alternates on axes other than the first.
* Alternates on multiple axes at the same time.

Supporting fully bounded ranges, with multiple axes, requires a change to the naming of the glyphs exported to UFO: instead of e.g. `A.BRACKET.300`, the glyphs are now named `A.BRACKET.wght_300_1000`. I feel like this is an improvement, as the range of the glyph is considerably easier to decode from the glyph name than before.

Additionally, the generation of designspace rules has changed; with this PR, glyphsLib emits a set of non-overlapping boxes. With an alternate layer `a` between wght=300-700, `b` between 300-700 and `x` between wght=200-700, the current behaviour of glyphsLib is to emit two rules, one for the `a` and `b` and one for the `x`. With this PR, two different rules will be emitted: one rule active been 200-300, substituting `x`, and one active between `300-700` substitution `a`, `b` and `x`. This is actually the same as what fontTools does with the substitutions when compiling the font anyway, and the resulting Python code is *much* simpler than the terrible mess we currently have which partitions the rules into forward and reverse "buckets".

Fixes #741.


Lib/glyphsLib/util.py
```diff
@@ -130,6 +130,14 @@ def pairs(list):
     return [list[i : i + 2] for i in range(0, len(list), 2)]
 
 
+def freezedict(dct):
+    return frozenset(dct.items())
+
+
+def unfreezedict(frozen):
+    return dict(list(frozen))
```
anthrotype, line 138: I don't think you need this `unfreezedict` helper at all, `dict(frozen)` already does what you want, no? `dict` takes an iterable already, don't need to make it a `list` first.

anthrotype:
> change to the naming of the glyphs exported to UFO: instead of e.g. A.BRACKET.300, the glyphs are now named A.BRACKET.wght_300_1000

should we support converting from the old names for backward compatibility with UFOs generated with previous versions of glyphs?

simoncozens:
> should we support converting from the old names for backward compatibility with UFOs generated with previous versions of glyphs?

I don't think we need to. It's not the Glyphs version that matters, because we support both v2 and v3, but the UFOs exported from pre-this-PR versions of glyphsLib will of course have the old layer names. However, they will also have a `_originalLayerName` lib key which is designed to get them round-tripping back into the GSFont correctly, so the fact that their UFO glyph name has changed won't actually be relevant.

simoncozens:
Regression tests will of course fail because all the glyph names are different. But you can inspect the diffs and convince yourself it's doing the same thing.

Lib/glyphsLib/builder/bracket_layers.py
```diff
@@ -4,117 +4,55 @@
...
-                axis_range_min,
-                axis_range_max,
-                reverse,
-            )
-            self._designspace.addRule(rule)
+    # overlayFeatureVariations does exactly what we need in terms of
+    # splitting rules into non-overlapping boxes and consolidating
+    # multiple substitutions into a single rule.
+    for rule in overlayFeatureVariations(rules):
+        box, mappings = rule
```
anthrotype, line 49:
```suggestion
    for (box, mappings) in overlayFeatureVariations(rules):
```

Lib/glyphsLib/builder/bracket_layers.py
```diff
@@ -4,117 +4,55 @@
...
+        # Reduce the mappings to a single mapping
+        mapping = {}
+        for this_mapping in mappings:
+            mapping.update(this_mapping)
```
anthrotype, line 53: you could also use `collections.ChainMap` if you like, I believe it does the same as iterating and updating
https://docs.python.org/3/library/collections.html#collections.ChainMap

```python
mapping = dict(ChainMap(*mappings))
```

tests/specal_layer_width_test.py
```diff
@@ -40,5 +40,5 @@ def test_substitution_layer_width():
     masters, instances = load_to_ufos(glyphs_file_path(), include_instances=True)
     assert masters[0]["B"].width == 500
     assert masters[1]["B"].width == 600
-    assert masters[0]["B.BRACKET.40"].width == 510
-    assert masters[1]["B.BRACKET.40"].width == 610
+    assert masters[0]["B.BRACKET.wght_40_100"].width == 510
+    assert masters[1]["B.BRACKET.wght_40_100"].width == 610
```
anthrotype, line 44: I know this file pre-existed this PR, but it would make sense to fix the typo in the test filename => `git mv tests/spec{,i}al_layer_width_test.py`

anthrotype (approved):
LGTM, thanks a lot Simon!!

Lib/glyphsLib/builder/bracket_layers.py
simoncozens: Nice idea. I looked into this, and found it applied the chaining the wrong way around (last to first). Then I tried `mapping = dict(ChainMap(*reversed(mappings)))` which works, but I felt it was Just Too Much Magic and what I originally wrote is considerably easier to follow.

PR #829, opened 2022-11-07
Smart component speedup

simoncozens:
Using `pickle` was indeed faster than using `deepcopy`, but it's still slower than shallow-cloning each node. This should speed up the regression tests significantly (and anything that uses smart components).

Also a slightly more helpful error message when a smart component is not interpolatable.

### DIFF
--- a/Lib/glyphsLib/builder/smart_components.py
+++ b/Lib/glyphsLib/builder/smart_components.py
@@ -22,10 +22,6 @@ OpenType interpolation model to adjust the node positions.
 
 from enum import IntEnum
 
-# We're going to use pickle/unpickle to copy the node objects because
-# it's considerably faster than copy.deepcopy
-import pickle
-
 from fontTools.varLib.models import VariationModel, normalizeValue
 from fontTools.ttLib.tables._g_l_y_f import GlyphCoordinates
 
--- a/Lib/glyphsLib/builder/smart_components.py
+++ b/Lib/glyphsLib/builder/smart_components.py
@@ -119,12 +119,14 @@ def to_ufo_smart_component(self, layer, component, pen):
     try:
         new_coords = model.interpolateFromMasters(normalized_location, coordinates)
     except Exception as e:
-        raise ValueError("Could not interpolate smart component %s used in %s" % (root.name, layer)) from e
+        raise ValueError(
+            "Could not interpolate smart component %s used in %s" % (root.name, layer)
+        ) from e
 
     # Decompose by creating a new layer, copying its shapes and applying
     # the new coordinates
     new_layer = GSLayer()
-    new_layer._shapes = [ shape.clone() for shape in masters[0]._shapes]
+    new_layer._shapes = [shape.clone() for shape in masters[0]._shapes]
     set_coordinates(new_layer, new_coords)
 
     # Don't forget that the GSComponent might also be transformed, so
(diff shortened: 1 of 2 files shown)


anthrotype:
regression tests still taking too long...

anthrotype:
in the other PR #828, the regression tests failed after 111 minutes with a suspicious error related to smart components, which might be relevant for here:

https://github.com/googlefonts/glyphsLib/actions/runs/3410863952/jobs/5674343376

```
  File "/home/runner/work/glyphsLib/glyphsLib/Lib/glyphsLib/builder/glyph.py", line 164, in to_ufo_glyph
    self.to_ufo_components(ufo_glyph, layer)  # .components
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/runner/work/glyphsLib/glyphsLib/Lib/glyphsLib/builder/components.py", line 70, in to_ufo_components
    to_ufo_smart_component(self, layer, component, pen)
  File "/home/runner/work/glyphsLib/glyphsLib/Lib/glyphsLib/builder/smart_components.py", line 123, in to_ufo_smart_component
    new_coords = model.interpolateFromMasters(normalized_location, coordinates)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/runner/work/glyphsLib/glyphsLib/generate/lib/python3.11/site-packages/fontTools/varLib/models.py", line 494, in interpolateFromMasters
    deltas = self.getDeltas(masterValues, round=round)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/runner/work/glyphsLib/glyphsLib/generate/lib/python3.11/site-packages/fontTools/varLib/models.py", line 457, in getDeltas
    delta -= out[j]
  File "/home/runner/work/glyphsLib/glyphsLib/generate/lib/python3.11/site-packages/fontTools/ttLib/tables/_g_l_y_f.py", line 1779, in __isub__
    assert len(a) == len(other)
AssertionError
```

anthrotype:
> the tests/tools/generate_regression_test_files.py script is run on main, not on this branch.

ok but what does that have to do with them being slow?

anthrotype:
oh I see, this PR _will_ speed them up but only after it gets merged.. but how faster? have you tried offline?

simoncozens:
Wrapping `generate_from_glyphs` in a try block so it completes even with bad inputs, on my M1 laptop it converts the files in 189s.

anthrotype:
i suppose the github runners are slower and support less parallelism than your laptop, but I'd be quite surprised if the runtime goes down from 1 hour and a half to a few minutes _just_ because of the changes in this PR..

simoncozens:
Prepare to be surprised. I'm still waiting for the version on main to finish.

simoncozens:
With this patch: `fontmake -o ufo -g DelaGothic.glyphs` finished (with an error) after 9s.
Without this patch: I gave up and hit ctrl-c after 8 minutes.

anthrotype:
that's impressive but I doubt it has to do with (avoiding the use of) pickle itself, more like the fact that you're now not deepcopying the unneeded attributes, right?

anthrotype (approved):

simoncozens:
The problem is that there is a cyclic link between a GSPath.node and GSNode._parent. Deepcopy doesn't do well copying cyclic dependencies.

schriftgestalt:
couldn’t that be avoided by implementing `__copy__` and `__deepcopy__`?

anthrotype:
yes maybe, but I think the purpose here was to make a more selective copy by excluding attributes that aren't needed for the task at hand, am I right @simoncozens? Perhaps the method name `.clone()` does not make that clear

PR #831, opened 2022-11-15
Implement corner components

simoncozens:
Smart components were easy! Corner components... not so much. I don't quite understand how they work. But once I *do*, here's a framework for getting them implemented:

* Export them as hints (which we do already) in the UFO, ensuring we pass along the hint's name (since this is the name of the corner component glyph).
* Add a custom filter to "apply" those hints, by modifying the glyphs' path.
  * Convert the hints into a sensible data structure
  * ???
  * Profit!

### DIFF
--- a/Lib/glyphsLib/builder/font.py
+++ b/Lib/glyphsLib/builder/font.py
@@ -40,6 +40,11 @@ def to_ufo_font_attributes(self, family_name):
         self.to_ufo_names(ufo, master, family_name)  # .names
         self.to_ufo_family_user_data(ufo)  # .user_data
 
+        if has_any_corner_components(font, master):
+            ufo.lib.setdefault(UFO2FT_FILTERS_KEY, []).append(
+                {"namespace": "glyphsLib.filters", "name": "cornerComponents", "pre": True}
+            )
+
         ufo.lib.setdefault(UFO2FT_FILTERS_KEY, []).append(
             {"namespace": "glyphsLib.filters", "name": "eraseOpenCorners", "pre": True}
         )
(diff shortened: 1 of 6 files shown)


simoncozens:
This may turn out to be useful information: https://help.fontlab.com/fontlab/8/whats-new/whats-new-05-build-assemble/#skin-filter

simoncozens:
It's done, ship it!

Screenshot 2022-11-18 at 06 42 23

simoncozens:
I think I have done as much on this as my sanity can manage for now. It seems to implement everything apart from the left/right anchors; there's a warning issued if these are present. The cubic splitting algorithm in fontTools is different to the one in Glyphs, so when the outstroke of a corner is a curve, this gives subtly different but visually comparable results.

simoncozens:
Write-up of algorithm: https://github.com/unified-font-object/ufo-spec/issues/213

bramstein:
I would love to see this merged. Anything I/The Type Founders can do to help ship this feature?

anthrotype:
@simoncozens is this in a mergeable state do you reckon? I can't go too deeply in the logic because I am not familiar enough with these.

anthrotype:
> mergeablew

(ofc, provided we fix the merge conflicts and lint CI job failing first)

simoncozens:
Yep, as mentioned I've really done as much as I can with this. And it's better than what we have already (i.e. nothing)

anthrotype:
thank you very much, Simon. Will you be able to fix the merge conflicts and lint job at some point this week so we can get this in the next release?

anthrotype:
the regression tests failing are expected, right? Feel free to merge when ready, thanks

Issue #735, opened 2021-10-21
Supports Rename Glyphs parameter


RosaWagner:
Some issues are talking about it #498 #126 and this one aims to list all the useful parameters that could be handled from source:

- [ ] Remove glyphs
- [x] Keep glyphs
- [x] Rename glyphs
- [ ] Remove features
- [x] Replace feature
- [ ] Remove classes

We are trying to unroll an update of [cormorant](https://github.com/CatharsisFonts/Cormorant/tree/master/4.%20Glyphs%20Source%20Files) which is creating different families by swapping stylistic sets, it looks like that in sources: 

<img width="759" alt="Capture d’écran 2021-10-21 à 14 43 38" src="https://user-images.githubusercontent.com/12222436/138281810-b3dc049e-47e0-4a23-be3f-153aa0762d39.png">

I see how I could reproduce the remove glyphs and remove feature with fonttools subsetter, but not the Rename Glyphs thing though (although making additional post-export script doesn't help the designer to use fontmake). We made a little test case for this particular parameter in case you want to implement it 😬

[test-swap.zip](https://github.com/googlefonts/glyphsLib/files/7389249/test-swap.zip) Exported result should be a circle.

cc @emmamarichal

anthrotype:
while it'd be nice to support all the myriad of custom parameters eventually, in the meantime I believe you should be able to achieve what you need by using the fonttools subsetter (to remove features and glyphs), and perhaps
also @twardoch's `pyftfeatfreeze` script from https://github.com/twardoch/fonttools-opentype-feature-freezer to get what the "Rename Glyphs" does (set .ss02 glyphs as the default ones).

anthrotype:
at least some of the custom params you mention should be supported already. IIRC, we do support Keep Glyphs and Replace feature. Doesn't Remove Glyphs achieve the same thing as Keep Glyphs (only reversed)? Same for Remove Feature, you can replace it with an empty one..
I'm trying to understand which one is really needed, and which one kind of isn't

MariannaPaszkowska:
➕1 to add this functionality. I tried `pyftfeatfreeze` script, and it seems to change a bit too much in the font. I have a lot of unrelated diffs.

anthrotype:
which one you mean? Rename Glyphs?

anthrotype:
How is Rename Glyphs supposed to work exactly?
[...]

PR #832, opened 2022-11-15
Implement Rename Glyphs custom parameter

simoncozens:
- Rename Glyphs custom parameter implementation
- Add tests for rename glyphs


anthrotype:
Thanks Simon! For the records, this is going to fix https://github.com/googlefonts/glyphsLib/issues/735

anthrotype:
Do we also need "Remove Glyphs" support to reach feature parity with the behaviour described in https://github.com/googlefonts/glyphsLib/issues/735#issuecomment-1270211329 ?

anthrotype (approved):

simoncozens:
Looks like we still need quite a few more. Will work on them.

PR #833, opened 2022-11-25
Ensure alternate glyph rules are floats

simoncozens:
Glyphs 3 sometimes stores the alternate glyph min/max coordinates as strings, for a bit of variety. See #741.

### DIFF
--- a/Lib/glyphsLib/classes.py
+++ b/Lib/glyphsLib/classes.py
@@ -3809,6 +3809,10 @@ class GSLayer(GSBase):
                 designspace_min, designspace_max = designspace_min_max(axis)
                 axis_min = rule.get("min", designspace_min)
                 axis_max = rule.get("max", designspace_max)
+                if isinstance(axis_min, str):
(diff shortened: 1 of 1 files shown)


anthrotype:
but why would sometimes be a str sometimes already a float?

simoncozens:
Oh, it's worse than that - sometimes the ints are actually strings:

```
attr = {
axisRules = (
{
max = "410";
min = "5";
},
{
max = "150";
min = "80";
},
{
}
);
};
```

anthrotype:
let's ask @schriftgestalt for clarification

schriftgestalt:
Can you point me to a .glyphs file location that shows this?

simoncozens:
Repo: https://github.com/emmamarichal/Playfair/blob/master/sources/Playfair-Italic.glyphs

schriftgestalt:
It can happen in older files. 
I’ll add code to validate this. But it might be a good idea to add a `float()` when accessing the values.

PR #834, opened 2022-11-25
Allow alternate layers to be unnamed

simoncozens:
Fixes the *other* issue with Playfair Italic in #741. Alternate layers may be unnamed because Glyphs does not rely on (or even display) the user-supplied name for them, and we make up our own name for them anyway.


simoncozens:
It's not covered. I could come up with a test.

PR #835, opened 2022-11-28
Support finding smart component info in G3 format

simoncozens:
Prior to Glyphs 3, smart component information was stored in a userData key; in Glyphs 3 format, it's stored in the `partSelection` slot of the layer itself. We need to look in both places.

### DIFF
--- a/Lib/glyphsLib/classes.py
+++ b/Lib/glyphsLib/classes.py
@@ -3666,11 +3666,14 @@ class GSLayer(GSBase):
     def smartComponentPoleMapping(self):
         if "PartSelection" not in self.userData:
             self.userData["PartSelection"] = {}
+        if self.partSelection:
+            return self.partSelection
         return self.userData["PartSelection"]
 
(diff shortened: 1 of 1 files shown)


Lib/glyphsLib/classes.py
```diff
@@ -3666,11 +3666,14 @@ def name(self, value):
     def smartComponentPoleMapping(self):
         if "PartSelection" not in self.userData:
             self.userData["PartSelection"] = {}
```
anthrotype, line 3668: do you still need to add "PartSelection" to userData dict in a Font loaded from G3 source where there's already a self.partSelection?

anthrotype:
> We need to look in _both_ places

so the data is duplicated?

simoncozens:
Well, it’s not duplicated but it’s in different places in different formats.

anthrotype:
> In my code check what format version and reads the data accordingly.

I try do the same in https://github.com/googlefonts/glyphsLib/pull/840

PR #838, opened 2022-12-08
Simple G3 RTL Kerning Support

yanone:
This is based on the latest discussion in https://github.com/googlefonts/glyphsLib/pull/778, namely @simoncozens’ [last comment](https://github.com/googlefonts/glyphsLib/pull/778#issuecomment-1335262059).

I needed to make this into a separate branch for my own sanity.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -154,7 +154,7 @@ regardless of writing direction.
 As of `PR #838 <https://github.com/googlefonts/glyphsLib/pull/838>`__ glyphsLib
 reverts this separate Glyphs 3-style RTL kerning back to Glyphs 2/UFO-style kerning
 upon conversion of a Glyphs object to a UFO object, *but it does not convert the kerning
-back to Glyphs 3-style when converting a UFO object to a Glyphs object.* 
+back to Glyphs 3-style when converting a UFO object to a Glyphs object!* 
 
 This means that if you convert a UFO to a Glyphs file and subsequently open that file
 in Glyphs 3, the RTL kerning will initially not be visible in the UI, but be hidden


yanone:
There are a lot of duplicate `.RTL` group now that ruin the tests. Let's focus on the core functionality first and then I'll think of ways to remove the unnecessary group.

Lib/glyphsLib/builder/kerning.py
```diff
@@ -20,31 +20,47 @@
...
             kerning_source = master
-        if kerning_source.id in self.font.kerning:
-            kerning = self.font.kerning[kerning_source.id]
-            _to_ufo_kerning(self, self._sources[master.id].font, kerning)
+        if kerning_source.id in self.font.kerningLTR:
+            kerning = self.font.kerningLTR[kerning_source.id]
+            _to_ufo_kerning(self, self._sources[master_id].font, kerning)
+        if kerning_source.id in self.font.kerningRTL:
+            kerning = self.font.kerningRTL[kerning_source.id]
+            _to_ufo_kerning(self, self._sources[master_id].font, kerning, "RTL")
```
anthrotype, line 32: since both LTR and RTL pairs end up in the single UFO kerning.plist dict, is a clash ever possible, and should be guard against that? If so, then they way this is coded means the last wins, so RTL duplicate sticks. I'm not 100% sure 1) it the clash is at all possible 2) what should we do about it

yanone: A "clash" as in: A kerning pair that has different values in both directions?

In that case I need to [quote Simon](https://github.com/googlefonts/glyphsLib/pull/778#issuecomment-1335262059): _That will work for all cases except for situations where the same glyph pair has both LTR and RTL kerning, with different values in each table. But we were never going to get that right anyway._

anthrotype: I mean a "clash" as in setting the same (first, second) key in the ufo.kerning dict as result of merging in both the LTR and RTL separate kerning dicts into the single dict. Maybe for the way we name the classes a situation where a class-class pair would clash is not possible, but a single-single pair might still clash.
At least a warning (right before overwriting a value) would help a user deal with it somehow

Lib/glyphsLib/builder/kerning.py
```diff
@@ -20,31 +20,47 @@
...
+                match = re.match(r"@MMK_R_(.+)", right)
+            elif direction == "RTL":
+                match = re.match(r"@MMK_L_(.+)", right)
             right_is_class = bool(match)
             if right_is_class:
-                right = "public.kern2.%s" % match.group(1)
+                if direction == "LTR":
+                    right = "public.kern2.%s" % match.group(1)
+                elif direction == "RTL":
+                    right = "public.kern2.%s.RTL" % match.group(1)
```
anthrotype, line 63: this repetition of if/elif blocks with minor differences feels a bit verbose, if we could factor some of this logic out and reuse would look nicer

Lib/glyphsLib/builder/kerning.py
yanone: Oh shite. Okay, I'll add a warning

Lib/glyphsLib/builder/kerning.py
anthrotype: also, watch out for if statements without an else branch, if `direction` is neither LTR nor RTL (I know, should be impossible) then we would incur in a NameError for the `match` local variable isn't initialized. Maybe do `else:` and then assert that it's the other value that you expect it to be so we enver fall off the if

anthrotype:
> ways to remove the unnecessary groups.

i guess, when you're done with the kerning, you can pass again on the groups and remove those that aren't actually used anywhere in the kerning dict

Lib/glyphsLib/builder/kerning.py
yanone: I can offer this:

```python
def _to_ufo_kerning(self, ufo, kerning_data, direction="LTR"):
    """Add .glyphs kerning to an UFO."""

    class_missing_msg = "Non-existent glyph class %s found in kerning rules."
    overwriting_kerning_msg = "Overwriting kerning value for %s."

    for left, pairs in kerning_data.items():
        match = re.match(
            r"@MMK_" + ("L" if direction == "LTR" else "R") + "_(.+)", left
        )
        left_is_class = bool(match)
        if left_is_class:
            left = "public.kern1.%s" % match.group(1)
            if direction == "RTL":
                left += ".RTL"
            if left not in ufo.groups:
                self.logger.warning(class_missing_msg % left)
        for right, kerning_val in pairs.items():
            match = re.match(
                r"@MMK_" + ("R" if direction == "LTR" else "L") + "_(.+)", right
            )
            right_is_class = bool(match)
            if right_is_class:
                right = "public.kern2.%s" % match.group(1)
                if direction == "RTL":
                    right += ".RTL"
                if right not in ufo.groups:
                    self.logger.warning(class_missing_msg % right)

            if left in ufo.kerning and right in ufo.kerning[left]:
                self.logger.warning(overwriting_kerning_msg % [left, right])
            ufo.kerning[left, right] = kerning_val
```

Lib/glyphsLib/builder/kerning.py
```diff
@@ -20,31 +20,47 @@
...
+            _to_ufo_kerning(self, self._sources[master_id].font, kerning, "RTL")
 
 
-def _to_ufo_kerning(self, ufo, kerning_data):
+def _to_ufo_kerning(self, ufo, kerning_data, direction="LTR"):
     """Add .glyphs kerning to an UFO."""
 
     warning_msg = "Non-existent glyph class %s found in kerning rules."
 
     for left, pairs in kerning_data.items():
```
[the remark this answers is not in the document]
yanone: Please comment on my proposal for the verbose LTR/RTL switches first. These changes would conflict the commits.

Lib/glyphsLib/builder/kerning.py
anthrotype: it is an improvement over the first iteration of the code, but I was thinking more in terms of reusing logic by parameterizing functions e.g.

```python
def _matches_mmk_class(name: str, direction: str, side: int) -> Optional[re.Match]:
    ...
```

or

```python
def _ufo_class_name(name, direction, side) -> str
```

you got the hang of it

yanone: How about this? It became more code overall, tho:

```python
def _to_ufo_kerning(self, ufo, kerning_data, direction="LTR"):
    """Add .glyphs kerning to an UFO."""

    class_missing_msg = "Non-existent glyph class %s found in kerning rules."
    overwriting_kerning_msg = "Overwriting kerning value for %s."

    for left, pairs in kerning_data.items():
        left, left_is_class = _ufo_class_name(left, direction, "left")
        if left_is_class and left not in ufo.groups:
            self.logger.warning(class_missing_msg % left)

        for right, kerning_val in pairs.items():
            right, right_is_class = _ufo_class_name(right, direction, "right")
            if right_is_class and right not in ufo.groups:
                self.logger.warning(class_missing_msg % right)

            # Emit kerning
            if left in ufo.kerning and right in ufo.kerning[left]:
                self.logger.warning(overwriting_kerning_msg % [left, right])
            ufo.kerning[left, right] = kerning_val


def _ufo_class_name(name, direction, side):
    """Return the UFO class name for a .glyphs class name."""
    if side == "left":
        side_key = "L" if direction == "LTR" else "R"
        side_number = "1"
    else:
        side_key = "R" if direction == "LTR" else "L"
        side_number = "2"

    match = re.match(r"@MMK_" + side_key + r"_(.+)", name)

    name_is_class = bool(match)
    if name_is_class:
        name = "public.kern%s.%s" % (side_number, match.group(1))
        if direction == "RTL":
            name += ".RTL"

    return name, name_is_class
```

anthrotype: it's looking better, but don;t use the terms left / right when you actually mean logical 1 or 2 position. Maybe "side" is not the right word (as it's tied to actual left/right), how about "order" (1 or 2)


(untested)

```python
import re


def _to_ufo_kerning(self, ufo, kerning_data, direction="LTR"):
    """Add .glyphs kerning to an UFO."""

    class_missing_msg = "Non-existent glyph class %s found in kerning rules."
    overwriting_kerning_msg = "Overwriting kerning value for %s."

    for first, pairs in kerning_data.items():
        first, is_class = _ufo_class_name(first, direction, 1)
        if is_class and first not in ufo.groups:
            self.logger.warning(class_missing_msg, first)

        for second, kerning_val in pairs.items():
            second, is_class = _ufo_class_name(second, direction, 2)
            if is_class and second not in ufo.groups:
                self.logger.warning(class_missing_msg, second)

            if (first, second) in ufo.kerning:
                self.logger.warning(overwriting_kerning_msg, first, second)
            ufo.kerning[first, second] = kerning_val


def _ufo_class_name(name, direction, order):
    """Return the UFO class name for a .glyphs class name."""
    if order == 1:
        side = "L" if direction == "LTR" else "R"
    else:
        assert order == 2
        side = "R" if direction == "LTR" else "L"

    match = re.match(fr"@MMK_{side}_(.+)", name)

    name_is_class = bool(match)
    if name_is_class:
        name = f"public.kerns{order}.{match.group(1)}"
        if direction == "RTL":
            name += ".RTL"

    return name, name_is_class
```

yanone: The words left and right were already there. I want to fix the logic first and then rename the variables.

anthrotype: they only make things more confusing, even more so now that we added the RTL logic, what they mean is actually logical first and second, since you're touching this code, I'd recommend you also do the renaming for clarity's sake

yanone: Anyway, there was one little mistake in your code, otherwise I took it as is.

Lib/glyphsLib/builder/kerning.py
```diff
@@ -20,34 +20,69 @@
...
-            kerning = self.font.kerning[kerning_source.id]
-            _to_ufo_kerning(self, self._sources[master.id].font, kerning)
+        if kerning_source.id in self.font.kerningLTR:
+            kerning = self.font.kerningLTR[kerning_source.id]
+            _to_ufo_kerning(self, ufo, kerning)
+        if kerning_source.id in self.font.kerningRTL:
+            kerning = self.font.kerningRTL[kerning_source.id]
+            _to_ufo_kerning(self, ufo, kerning, "RTL")
 
+        # Prune kerning groups that are not used in kerning rules
```
anthrotype, line 35: you could have `_to_ufo_kerning` method return the set of used groups, make a union of the LTR and RTL ones and use at the end to prune

anthrotype: I think the failing tests have to do with roundtripping, your code doesn't handle that.. not sure how to best unblock this

yanone: Well, now that we have explicitly named RTL classes, round-tripping should be implementable, shouldn't it?

yanone:
## About round-tripping

The way I see it, it's not easily possible, for the same reasons we were stuck at earlier: lack of directional data for single glyph kerning.

Theoretically it should happen here:

```python
def to_glyphs_kerning(self):
    """Add UFO kerning to GSFont."""
    for master_id, source in self._sources.items():
        for (left, right), value in source.font.kerning.items():
            if BRACKET_GLYPH_RE.match(left) or BRACKET_GLYPH_RE.match(right):
                # Skip all bracket glyph entries, as they are duplicates of their
                # parents'.
                continue
            left_match = UFO_KERN_GROUP_PATTERN.match(left)
            right_match = UFO_KERN_GROUP_PATTERN.match(right)
            if left_match:
                left = "@MMK_L_{}".format(left_match.group(2))
            if right_match:
                right = "@MMK_R_{}".format(right_match.group(2))
            print(left.endswith(".RTL"), right.endswith(".RTL"), left, right)
            self.font.setKerningForPair(master_id, left, right, value)
```

The inserted print statement shows the problem (`False False` in lowest line):
```
True False @MMK_L_leftAlef.RTL he-hb
True True @MMK_L_leftBet.RTL @MMK_R_rightAlef.RTL
True True @MMK_L_reh-ar.RTL @MMK_R_hah-ar.init.swsh.RTL
False True he-hb @MMK_R_rightAlef.RTL
False False he-hb he-hb
```

Any suggestions welcome.

## About the checks

I need help with the checks.

### `tests/builder/to_glyphs_test.py:167: AssertionError`

The pruning currently doesn't check for empty groups or whether they make any sense at all. It only checks for whether or not groups are part of kerning rules, so my assumption is that no kerning has been defined for these test-cases. Above from where the error is thrown, several groups are added that are nonsensical, like empty groups or groups named "notInFont". So these get thrown out. This check definitely clashes with kerning group pruning in several of those nonsensical groups. My added kerning group pruning is explicit in checking for groups to be pruned to start with `public.kern` so that arbitrary groups will be preserved.

Here are some examples of nonsensical groups in that check:
```python
    ufo.groups["public.kern1.notInFont"] = ["L"]
    ...
    # Empty groups as well (found in the wild)
    ufo.groups["public.kern1.empty"] = []
```

One suggestion to deal with this is to, instead of comparing dictionaries to be identical, to use a dictionary comparison to compare the difference of the two dictionaries to be identical to a known set of pruned groups.
I would use `deepdiff` (to my knowledge the only way to check for differences in dictionaries regardless of the dictionary order, as in the diff output will always be ordered in the same way). This adds an extra dependency package. Suggestions welcome.

This will output: `{'dictionary_item_removed': [root['public.kern1.T'], root['public.kern2.oe'], root['public.kern1.notInFont'], root['public.kern1.halfInFont'], root['public.kern1.empty'], root['public.kern1.hebrewLikeT'], root['public.kern2.hebrewLikeO']]}` which kind of confirms my idea and confirms that only kerning groups are pruned.


###`tests/builder/builder_test.py:2268: AssertionError`

Using the same technique, the dictionary difference is: `{'dictionary_item_removed': [root['public.kern2.foo'], root['public.kern1.foo']]}`

A few lines further down in that check, kerning is actually checked to be empty: `assert ds.sources[1].font.kerning == {}`, which confirms why the groups aren't there.

I propose to alter both checks to make sense under the idea of pruning unused kerning groups. I'm only insecure about using `deepdiff`. Maybe you know of a better way to compare without added dependencies. I know that sets can do stuff like that, but I'm not sure of the data order.

yanone:
Update:

I went ahead and tweaked the two existing checks that were previously failing as they were comparing two dictionaries with expected groups to be identical, but kerning groups without kerning rules now get pruned. Now it's checking that all other groups stay in place but only groups that are known to have gotten pruned are missing.

About round-tripping:
I've confirmed that round-tripping produces at least canonical results in `glyphs3_test.py` by actually round-tripping the UFO object back to Glyphs and then back to UFO once more, comparing that the kerning and groups are identical.

Saving a UFO back to a Glyphs now basically produces G2-style RTL kerning with *everything* being in `.kerningLTR`. Power users are familiar with the procedure of converting that to G3-style RTL kerning through various scripts out there that move the kerning to `.kerningRTL`, using script or direction data as then provided by Glyphs natively. Casual users will trip over this, but not even Glyphs itself does that conversion from G2 to G3 automatically when saving in a new source file version.

At least the kerning will technically stay intact. It won't be accessible through UI when in RTL mode without the use of the aforementioned 3rd-party scripts.

The remaining checks that are now failing have ostensibly nothing to do with my PR directly. Please have a look.
I'm done with this as far as I can tell.

yanone:
We *could* remove the `.RTL` ending for groups when converting from UFO back to Glyphs in order to at least clean up the naming, but then the round-tripping check would fail out of the box. I would then probably tweak that check to take that into consideration, as otherwise the groups wouldn't be identical anymore right away.

simoncozens:
This looks great when I am building master TTFs, however, when I build instances, I'm hitting things like `Generating fonts from Designspace failed: The glyph "period" occurs in too many kerning groups` when there are LTR and RTL kerns. Trying to dig further.

simoncozens:
... which may be related to

```
WARNING:fontmake.instantiator:The source ... contains different groups than the default source. The default source's groups will be used for the instances.
```

so maybe it's a problem with my font.

Lib/glyphsLib/builder/kerning.py
```diff
@@ -46,24 +47,28 @@ def _to_ufo_kerning(self, ufo, kerning_data, direction="LTR"):
     class_missing_msg = "Non-existent glyph class %s found in kerning rules."
     overwriting_kerning_msg = "Overwriting kerning value for %s."
 
+    used_groups = set()
+
     for first, pairs in kerning_data.items():
         first, is_class = _ufo_class_name(first, direction, 1)
         if is_class:
-            yield first
+            used_groups |= {first}
```
anthrotype, line 55: just do `used_groups.add(first)`

tests/builder/builder_test.py
```diff
@@ -2265,9 +2265,25 @@ def test_load_kerning_bracket(ufo_module):
...
+    d2 = bracketed_groups
+    assert set(d2) ^ set(d1) == {"public.kern2.foo", "public.kern1.foo"}
+
     assert ds.sources[2].font.groups == bracketed_groups
-    assert ds.sources[3].font.groups == bracketed_groups
+
+    # Prior to PR #838, we were checking for d1 and d2 to be equal.
+    # However, with PR #838, we are now pruning kerning groups that are not
+    # used in the kerning table.
+    # The groups in the assertion below got pruned.
```
anthrotype, line 2282: feels redundant to repeat exactly the same inline comment as a couple lines above

tests/builder/builder_test.py
```diff
@@ -2298,9 +2314,25 @@ def test_load_kerning_bracket(ufo_module):
...
+    d2 = bracketed_groups
+    assert set(d2) ^ set(d1) == {"public.kern2.foo", "public.kern1.foo"}
+
     assert ds2.sources[2].font.groups == bracketed_groups
-    assert ds2.sources[3].font.groups == bracketed_groups
+
+    # Prior to PR #838, we were checking for d1 and d2 to be equal.
+    # However, with PR #838, we are now pruning kerning groups that are not
+    # used in the kerning table.
+    # The groups in the assertion below got pruned.
```
anthrotype, line 2331: make this repeated comment less verbose at least, copy pasting repeatedly doesn't help

tests/glyphs3_test.py
```diff
@@ -45,18 +45,33 @@ def test_glyphs2_rtl_kerning(datadir, ufo_module):
...
-    assert ufos[0].kerning[("he-hb", "he-hb")] == -21
+    # Round-tripping back to Glyphs
+    round_tripped_glyphs_font = to_glyphs(first_derivative_ufos)
+
+    # Second conversion back to UFO
+    designspace = to_designspace(round_tripped_glyphs_font, ufo_module=ufo_module)
+    second_derivative_ufos = [source.font for source in designspace.sources]
+
+    # Comparing kerning between first and second derivative UFOs:
+    # Round-tripped RTL kernining ends up as LTR kerning, but at least it's lossess
```
anthrotype, line 74: did you mean 

```suggestion
    # Round-tripped RTL kernining ends up as LTR kerning, but at least it's lossless
```

anthrotype:
I suspect the warning @simoncozens is seeing may have to do directly with the fact that with this PR kerning groups that aren't used in kerning rules will now get pruned, so some masters may end up with differing groups.plist. The expectation in fontmake.instantiator was that all masters have the same groups definition as the default master.
Maybe you should make a union of all the groups used by all the masters' kerning dicts, and keep those everywhere..

simoncozens:
I can't share screenshots, but I have found situations where this PR produces correct RTL kerns and *Glyphs 3 (3109) doesn't*. Nice work, Yan.

anthrotype:
> We could remove the .RTL ending for groups when converting from UFO back to Glyphs

we could.. but I wonder if that .RTL suffix could be used as some sort of hint as to whether a kerning pair needs to be placed in kerningRTL? Hm sounds a bit brittle, I know - nevermind.

simoncozens:
> This looks great when I am building master TTFs, however, when I build instances, I'm hitting things like Generating fonts from Designspace failed: The glyph "period" occurs in too many kerning groups when there are LTR and RTL kerns. Trying to dig further.

This isn't a problem with the font, it's a problem with this approach. I forgot that there is a limitation in UFO in which it does not let you put the same glyph in two different kerning groups. The error that I'm getting, with a little added instrumentation, is:

```
 The glyph "ellipsis" occurs in too many kerning groups: public.kern2.period, public.kern2.period.RTL.
```

You and I know that it's absolutely fine, and the kerns involved will never overlap, but the UFO validator doesn't. So you can't have both an LTR kern involving a period/ellipsis group and also an RTL kern involving a period/ellipsis group.

I don't know how to fix this.

anthrotype:
>  The glyph "ellipsis" occurs in too many kerning groups: public.kern2.period, public.kern2.period.RTL.

who/what issues that warning?

simoncozens:
fontTools.ufoLib.validators.groupsValidator, on saving the UFO.

anthrotype:
ah, right...

https://unifiedfontobject.org/versions/ufo3/groups.plist/#4-glyphs-must-not-appear-in-more-than-one-kerning-group-per-side

> 4. Glyphs must not appear in more than one kerning group per side.

anthrotype:
ufo spec thinks that's ambiguos because it assumes that kerning.plist is one big dict that will end up in a single PairPos lookup, whereas in reality our kern feature writer will split them up into LTR/RTL (and soon in script-specific lookups) so that those specific groupings aren't going to be ambiguous.
The ufoLib validators can be disabled by the way..

yanone:
So shall I consolidate the kerning group pruning to be identical across all masters? I couldn't take away a clear call to action here.

yanone:
I reacted to most of your comments.

But I didn't do the kerning group pruning across all master UFOs. I tried, but it produced several errors along the lines of `glyph A is part of too many kerning groups` and hundreds of warnings, 7 failed checks in total.
I'm not exactly sure what the problem here is.

The way the PR is currently set up (altering the checks in only a few places to check for groups that are sensical to be pruned according to the nonsensical check setup) passes all checks.

yanone:
Windows checks are passing now. Only the regression check remains

anthrotype:
we want groups.plist be the same for all master UFOs, as fontmake.instantiator will just use the groups.plist of the default master when creating instance UFOs.
In Glyphs.app the kerning groups are at the glyph level and as such apply to all masters, so when converting to designspace all the UFO sources (1 for each GSMaster) are expected to have the same groups.plist.
Please take the union of all the used groups across all masters when deciding which groups to keep and ensure they are the same across all UFO masters, and fix up any test failures that may arise accordingly.
As regards the other ufoLib warning about glyph being in more than one group, I think for the time being the only solution, or rather, workaround is to add a logging filter to fontmake that specifically mutes that particular logging warning when building from .glyphs sources. Ignore that one in this PR.

Lib/glyphsLib/builder/groups.py
```diff
@@ -87,13 +88,41 @@ def to_ufo_groups(self):
             source.font.groups[name] = glyphs[:]
 
 
+def _to_ufo_groups(glyph, groups, glyph_group, side, other_side):
```
anthrotype, line 91: I suggest you call this method something like `_add_glyph_to_ufo_groups`; as it stands, it seems like it's making the whole ufo groups, whereas it's simply adding one glyph to some of them

Lib/glyphsLib/builder/groups.py
```diff
@@ -87,13 +88,41 @@ def to_ufo_groups(self):
...
+        # This will add lots of unused groups which we'll prune later
+        # but is necessary to implement G3 RTL kerning. See kerning.py
+        group = f"public.kern{other_side}.{glyph_group}.RTL"
+        if glyph.name not in groups[group]:
+            groups[group].append(glyph.name)
+    else:
+        # Additional RTL group
+        # This will add lots of unused groups which we'll prune later
+        # but is necessary to implement G3 RTL kerning. See kerning.py
+        if not glyph_group.endswith(".RTL"):
```
anthrotype, line 107: this if branch is never hit, because you have already ascertained that glyph_group.endswith(".RTL") within this `else:` block...

Lib/glyphsLib/builder/groups.py
```diff
@@ -87,13 +88,41 @@ def to_ufo_groups(self):
             source.font.groups[name] = glyphs[:]
 
 
+def _to_ufo_groups(glyph, groups, glyph_group, side, other_side):
+    if not glyph_group.endswith(".RTL"):
```
anthrotype, line 92: what if a user names their leftKerningGroup or rightKerningGroup with .RTL suffix...

Lib/glyphsLib/builder/groups.py
yanone: Right, that was a leftover of a previous version of this section. I removed the condition

Lib/glyphsLib/builder/groups.py
yanone: I don't have a definite answer to this question. But I took the opportunity to replace all remaining instances of `name.replace(".RTL", "")` with `name[:-4]` as it was already used in `groups.py`. This should preserve any such a manually named group. Is this important enough to have me check this specifically?

yanone: I see what you mean. If such a manually named group comes in here in `groups.py`, it would probably clash with this functionality.

yanone: We could decide for a more unusual ending such as `.__RTL__`

anthrotype: if the suffix is only a temporary placeholder and never gets ends up in the generated sources because we stirp it as soon as we have figured out RTL-ness of a kerning pair, then yes, we definitely need to use one that is as unique as possible and that must not clash with one a user might use.
Still a bit unclear to me why you need to check if the group ends with .RTL or not.. is it to avoid adding it twice? Why could that happen?

yanone: Two things were previously happening: I got groups that had double .RTL ending and I had groups that had all glyphs twice in it and I had all groups twice too, I believe, once with this side number and the other ending, once with the other side number and this ending. The code gets called twice somewhere, and even if only for the conversion back to Glyphs, don't ask me where. The current implementation prevents that in a reactionary manner.

Other topic: Now I know why I started purging the .RTL ending from all names: by doing that, we cleanly separate the data layer from the application layer, and revert to a clean G2 state. It's really Georg's responsibility to sort that on import which he never did but still should. I'll do that tomorrow

yanone: Before we're done, I would like to add a paragraph to the main README about this. People need to know exactly what's going on so they understand how to handle UFO->Glyphs kerning, even if that's not the main use case of glyphsLib.

yanone:
I updated the README about the kerning conversion. Please have a look. There's nothing more to change from my side.

yanone:
Happy New Year.
Can we please revisit this issue? We're slowly approaching one year now since I discovered the problem (and longer since Georg quietly changed his file data). That's one lost year for open-source, Glyphs-based Arabic typography.

yanone:
Can we please revisit this? 

I've randomly checked the differences of the regression tests and have found that these are all kerning groups for which no kerning is defined, so they are all legitimately missing from the UFOs as generated by this PR.

I therefore strongly believe that this PR is good and see no reason why this is stalled.

anthrotype:
I'm inclined to close this in favour of https://github.com/googlefonts/glyphsLib/pull/865

Issue #837, opened 2022-12-08
Alternate layers in components


emmamarichal:
Hello!

I have a font that has alternates layers in components (for example, the layers are in the bar used in the dollar, but not in the dollar itself).
The designer manages to make it work in his exports (maybe he exports from glyphs: https://github.com/clauseggers/Playfair/pull/38#issuecomment-1335039800), but when I export with gftools builder, it doesn't work. 

The solution would be to create layers in each affected glyph, but I wanted to see if it was possible to fix this directly here so it can work with fontmake?
It concerns for exemple: E-cy Ereversed-cy Tcedilla tcedilla tcedilla.smcp e-cy ereversed-cy cent dollar

You can find the sources here: https://github.com/emmamarichal/Playfair

Maybe @simoncozens you can take a look?
Thank you 🙏 
Cheers!

anthrotype:
I think this is the key part:

https://github.com/googlefonts/glyphsLib/blob/f6e9c4a29ce764d34c309caef5118c48c156be36/Lib/glyphsLib/builder/bracket_layers.py#L114-L118

I think the solution is to ensure that all composite glyphs which contain components with bracket (alternate) layers, also have the same alternate layers.

Also see this issue https://github.com/googlefonts/glyphsLib/issues/535

emmamarichal:
Oh yes I'm sorry I didn't saw these older issues! I'll take a look, thank you!

simoncozens:
This is not the same issue as the one that the test case fixes (since that test still passes!). In the test case, glyph B uses component A, and B has the appropriate alternate layers defined on it - the same layers as A. In the Playfair situation, the parent glyph does not have the same alternate layer set-up as the component that it uses.

I've got half way to a fix on this. The easy case is where the parent glyph has no alternate layers at all; then you just clone the master layer for each needed component use and either rename (G2) or set attributes (G3; oh, it's so fun having to support both).

The hard case is something like _f_ydieresis.ligature in Playfair, which has its own alternate layers defined, but the component glyph (y) has a *different* alternate layer setup. Then you have to do interesting box arithmetic to work out which layer to clone.

simoncozens:
Oh *hell*. Almost everything working except:

```
fontmake: Error: In 'sources/Playfair.glyphs' -> 'master_ufo/Playfair.designspace': Generating fonts from Designspace failed: <features>:2961:9: Glyph names must not be longer than 63 characters
```

```
[...]

PR #839, opened 2022-12-09
More alternate layer fixes

simoncozens:
This fixes as much of #837 as I'm prepared to fix. It deals with the situation of parent glyphs which have no alternate layers but use component glyphs which do have alternate layers.

It doesn't deal with the situation of parent glyphs which have alternate layer setups which differ from those of their component glyphs, but it has a decent stab at it and warns the user. (Those are pretty rare, anyway. In the situation of Playfair 3 which had 633 "problematic" layers, only 8 of them were in this category.)

It also deals with my obviously short-sighted approach to glyph naming for alternate layers. Instead of producing long names like `_f_ydieresis.ligature.BRACKET.opsz_5_410.wdth_50_75.wght_690_900`, it now just numbers them.

### DIFF
--- a/tests/builder/builder_test.py
+++ b/tests/builder/builder_test.py
@@ -2261,8 +2261,8 @@ def test_load_kerning_bracket(ufo_module):
 
     ds = glyphsLib.to_designspace(font, minimize_glyphs_diffs=True)
     bracketed_groups = {
-        "public.kern2.foo": ["a", "a.BRACKET.wght_300_1000"],
-        "public.kern1.foo": ["x", "x.BRACKET.wght_300_1000", "x.BRACKET.wght_600_1000"],
+        "public.kern2.foo": ["a", "a.BRACKET.varAlt01"],
+        "public.kern1.foo": ["x", "x.BRACKET.varAlt01", "x.BRACKET.varAlt02"],
     }
     assert ds.sources[0].font.groups == bracketed_groups
     assert ds.sources[1].font.groups == bracketed_groups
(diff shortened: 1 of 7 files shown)


tests/builder/builder_test.py
```diff
@@ -2261,8 +2261,8 @@ def test_load_kerning_bracket(ufo_module):
 
     ds = glyphsLib.to_designspace(font, minimize_glyphs_diffs=True)
     bracketed_groups = {
-        "public.kern2.foo": ["a", "a.BRACKET.wght_300_1000"],
-        "public.kern1.foo": ["x", "x.BRACKET.wght_300_1000", "x.BRACKET.wght_600_1000"],
+        "public.kern2.foo": ["a", "a.BRACKET.varAlt01"],
```
anthrotype, line 2264: do we need to keep both "BRACKET" and "varAlt" in the same name? Aren't the two labels supposed to convey the same meaning?
how about either just "BRACKET.01" or "varAlt01"?

anthrotype: We do lose a lot of the information that was previously baked in the (arguably quite long) glyph name before.

I wonder what order does this varAlt01, 02, 03, naming etc. follows? Is it deterministic at least?

simoncozens: We don't *need* to keep both, but I went this way because I think we want to keep something which is unmistakably bracket-related, for the various regex-based tests to work; and I went with `varAlt01` because that's what Glyphs adds.

anthrotype: if varAlt01 is what Glyphs.app uses, I'd say let's go for that one only, and fix related tests.

In what order are the digit suffix in varAlt01, 02, 03 assigned and incremented, in what order? Is it "don't ask too complicated doesn't actually matter" kind of order?

anthrotype: i see, it depends on stably ordered stuff like glyph order, layer order and order of components within each layer, so at least it's deterministic, it just as eloquent as the previous naming scheme

anthrotype (approved):
to be honest I just skimmed this one, I don't think I can do a thorough review. I trust Simon can fix any follow up issues that may arise, so FWIW here's my LGTM

@simoncozens if you want someone else to take a look, try ask the DaMa folks, otherwise feel free to merge, thanks

PR #840, opened 2022-12-16
Support finding smart component info in G3 format (2)

anthrotype:
Same as https://github.com/googlefonts/glyphsLib/pull/835 but checks the format version instead of looking in both places and avoids storing/writing it twice.

Prior to Glyphs 3, smart component information was stored in a `userData` key; in Glyphs 3 format, it's stored in the `partSelection` slot of the layer itself. We need to look in either places depending on the GSFont's format_version.


simoncozens:
LGTM, thanks!

Issue #841, opened 2022-12-16
regression tests fail on main after #839 was merged


anthrotype:
See https://github.com/googlefonts/glyphsLib/actions/runs/3658053080/jobs/6182355670

https://github.com/googlefonts/glyphsLib/pull/839

```
Traceback (most recent call last):
  File "/opt/hostedtoolcache/Python/3.11.0/x64/lib/python3.11/multiprocessing/pool.py", line 125, in worker
    result = (True, func(*args, **kwds))
                    ^^^^^^^^^^^^^^^^^^^
  File "/home/runner/work/glyphsLib/glyphsLib/tests/tools/generate_regression_test_files.py", line 16, in generate_from_glyphs
    glyphsLib.build_masters(gs_path, output_dir, None, designspace_path=ds_path)
  File "/home/runner/work/glyphsLib/glyphsLib/Lib/glyphsLib/__init__.py", line 120, in build_masters
    designspace = to_designspace(
                  ^^^^^^^^^^^^^^^
  File "/home/runner/work/glyphsLib/glyphsLib/Lib/glyphsLib/builder/__init__.py", line 123, in to_designspace
    return builder.designspace
           ^^^^^^^^^^^^^^^^^^^
  File "/home/runner/work/glyphsLib/glyphsLib/Lib/glyphsLib/builder/builders.py", line 335, in designspace
    self.to_designspace_bracket_layers()  # .bracket_layers
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/runner/work/glyphsLib/glyphsLib/Lib/glyphsLib/builder/bracket_layers.py", line 26, in to_designspace_bracket_layers
    find_component_use(self)
  File "/home/runner/work/glyphsLib/glyphsLib/Lib/glyphsLib/builder/bracket_layers.py", line 245, in find_component_use
    new_layer = synthesize_bracket_layer(
                ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/runner/work/glyphsLib/glyphsLib/Lib/glyphsLib/builder/bracket_layers.py", line 259, in synthesize_bracket_layer
    axis, (bottom, top) = next(iter(box.items()))
                          ^^^^^^^^^^^^^^^^^^^^^^^
StopIteration
```

simoncozens:
Taking a look now; it's Rubik.

anthrotype:
[...]

PR #842, opened 2022-12-16
Only add box to box list if there's actually a box

simoncozens:
So it turns out that #841 was caused by the fact that `[()]` is True, and so boxes were getting added to the list of layers to synthesise even if there wasn't actually any alternate-layer information in the layer.


anthrotype:
will that fix the regression tests after we merge it, right? If so, LGTM

simoncozens:
I don’t know. It will allow the test suite to generate a full set of “before” data, so I guess so.

anthrotype:
let's try

PR #848, opened 2023-01-17
Don't escape reserved names. Fixes #846

### LINKED ISSUE #846: .glyphspackage name rules are slightly different to UFO rules
We currently use `fontTools.misc.filenames.userNameToFileName` to convert a glyph name to a filename when reading/writing glyphspackage files.

As well as adding underscores to make a distinction between capitals and lower case letters on case-folding filesystems, `userNameToFileName` also adds an underscore to the start of certain "reserved" file names (such as the MSDOS console device "con"). Glyphs 3 does not apply this reserved file names check, so a glyph called `con` (U+A76F LATIN SMALL LETTER CON) is stored by Glyphs 3 as `glyphs/con.glyph`. We turn it into `glyphs/_con.glyph` and then we don't find it.

simoncozens:


anthrotype:
Have you tested on a real windows machine that `"{reservedFilename}.glyph"` still works? /cc @madig @belluzj

simoncozens:
I haven't. I'll add a test, and then the CI can tell us.

belluzj:
You can't create such a file by renaming from the file explorer, and if you create one otherwise, you can't open it e.g. with Notepad (and despite the .glyph extension, Notepad tries to open a con.txt...)
image
image

belluzj:
I managed to create the file from git-bash with `touch con.glyph` despite the UI bugs in explorer.exe, so a headless CI may seem fine despite issues in the Windows UI

belluzj:
I can't delete the file from explorer.exe (this file makes Windows go GONG and error dialog after error dialog)
image

simoncozens:
OK, having the file in the repo crashed git on windows. ;-)

Yeah, @schriftgestalt, you probably need to fix this on the Glyphs end...

schriftgestalt:
I fixed it. Thanks for checking this.

simoncozens:
So with this fixed in Glyphs, this PR is no longer needed. (For files generated with Glyphs 3.2 patch level 3169 and onwards. Of course, for files generated with older versions of Glyphs 3, it will be needed...)

PR #849, opened 2023-01-19
Don't use default feature writers for manual kern

### LINKED ISSUE #764: Default feature writers heuristic is baffling
If the Glyphs file does not have a manual `kern` feature, glyphsLib does nothing special, and ufo2ft uses a default list of feature writers, which is

```
        KernFeatureWriter,
        MarkFeatureWriter,
        GdefFeatureWriter,
        CursFeatureWriter,
```

But if the Glyphs file has a manual `kern` feature, glyphsLib writes a `com.github.googlei18n.ufo2ft.featureWriters` lib entry in the designspace file:

https://github.com/googlefonts/glyphsLib/blob/609b1765096b016e7382b1155c4034888a0de878/Lib/glyphsLib/builder/user_data.py#L47-L55

glyphsLib's version of `DEFAULT_FEATURE_WRITERS` contains two writers: `KernFeatureWriter` and `MarkFeatureWriter`.

So an Arabic which has a manual kern feature (even if that feature just says `# Automatic code start`) will lose its cursive positioning feature. That was surprising!

I could fix this by adding more feature writers to glyphsLib's default list, but I don't think that's the right fix. I don't understand why we are we making our own "default" list, which we now have to keep in sync with ufo2ft, when *not doing anything at all* would cause ufo2ft to do the right thing.

simoncozens:
Fixes #764; don't fiddle with the feature writers lib key, just add an Automatic Code marker instead.

- Use Automatic Code marker when there's a manual kern
- Tests for the above


anthrotype (approved):
LGTM, thank you Simon!

Issue #847, opened 2023-01-17
compute COLRv1 ClipBoxes to work around a bug in Chrome


anthrotype:
COLRv1 fonts built from .glyphs sources using fontmake do not explicitly define clipBoxes, in fact the ClipList is optional and should be computed on the fly when omitted says the COLRv1 spec. However, Chrome 108 has a bug whereby some gradients are not being rendered if the font doesn't contain some ClipBoxes.

Details are in https://bugs.chromium.org/p/chromium/issues/detail?id=1404678
This first surfaced in https://github.com/fonttools/fonttools/issues/2938
Also related: https://github.com/googlefonts/fontbakery/issues/4020

Here's an example of a font that is built that way: https://fonts.google.com/specimen/Blaka+Ink

COLRv1 fonts built using nanoemoji (from .svg inputs, e.g. Noto Color Emoji) won't have this issue because nanoemoji always precomputes explicit clip boxes, never omits them.

I think glyphsLib should also always compute them. They don't take that much space as things gets reused, it will make rendering a bit faster as renderer doesn't have to traverse the graph to compute on the fly, and will also work around this Chrome bug.

Current code lives in https://github.com/googlefonts/glyphsLib/blob/main/Lib/glyphsLib/builder/color_layers.py

drott:
> However, Chrome 108 has a bug whereby some gradients are not being rendered if the font doesn't contain some ClipBoxes.

I would describe the bug differently. The bug is not about ClipBoxes but about reading ColorLines as the end of the table. Placing ClipBoxes there happens to fix the bug because it has the side effect of making the bounds checks in reading the ColorLine pass. The bug itself is not more or less reason to add ClipBoxes. 

We previously discussed the difficulty of computing maximum ClipBoxes or suitable variable ClipBoxes with PaintVarRotate and PaintVarTransform.

anthrotype:
Thanks for the clarification Dominik, I didn't want to go into the details here, just linked to the upstream bug which explains that -- but you're right the bug is not really about clipboxes, and their presence just happens to workaround it.

But yeah, you make a valid point about [computing variable glyphs max bounds](https://github.com/googlefonts/nanoemoji/issues/239) being hard when there are complex rotation/transforms...

anthrotype:
[...]

PR #853, opened 2023-01-24
Compute clipBoxes for COLRv1 glyphs

anthrotype:
Fixes https://github.com/googlefonts/glyphsLib/issues/847

I want to add tests before I remove the "WIP" but I file the PR to confirm the logic is sound

### DIFF
--- a/Lib/glyphsLib/builder/color_layers.py
+++ b/Lib/glyphsLib/builder/color_layers.py
@@ -14,11 +14,17 @@
 
 import math
 
+from fontTools.misc.arrayTools import calcBounds, unionRect
 from fontTools.misc.transform import Identity, Transform
 from fontTools.ttLib.tables.otTables import PaintFormat
 
+from glyphsLib.affine import Affine, identity
 from .common import to_ufo_color
-from .constants import UFO2FT_COLOR_LAYERS_KEY, UFO2FT_COLOR_PALETTES_KEY
+from .constants import (
+    UFO2FT_COLOR_LAYERS_KEY,
+    UFO2FT_COLOR_PALETTES_KEY,
+    UFO2FT_COLR_CLIP_BOXES_KEY,
+)
 
 
 def _to_ufo_color_palette_layers(builder, master, layerMapping):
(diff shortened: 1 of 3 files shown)


anthrotype:
I had to also modify the `GSComponent.layer` property (which in the Glyphs Python Scripting API docs seems to be called differently as "[componentLayer](https://docu.glyphsapp.com/#GSComponent.componentLayer)" by the way..) and I am not 100% sure about this change, would love to get some confirmation as to the expected behavior.

Basically in order to compute the bounding box for a layer that contains components, I need to dereference these components to the corresponding layer in another glyph that these point to. The current code simply takes the component's base glyph and gets a layer with the same layerId as the component's parent layer, but I think this only works as intended for master layers. It surely did not work for me when I was using it with color layers.

For these, from my experiments within Glyphs.app, it appears that the layer that a component inside a color layer actually is pointing to is the _first_ color layer in the pointed-to glyph if any, otherwise the associated master layer.

Can anybody please confirm this?

/cc @schriftgestalt @khaledhosny @simoncozens

schriftgestalt:
`component.layer` is the layer the component is in. Could have been called "container" or "parent". 

`component.compnentLayer` is the matching layer of the glyphs the component points to. Mostly that will do: `component.compoent.layers[component.layer.layerId]`But it will interpolate a new layer when the comment is in a brace layer of when it is a smart component. 

> For these, from my experiments within Glyphs.app, it appears that the layer that a component inside a color layer actually is pointing to is the first color layer in the pointed-to glyph if any, otherwise the associated master layer.

Yes. It tries to find a matching layer and otherwise falls back to the default layer.

anthrotype:
> component.layer is the layer the component is in

oh, that's not how component.layer is implemented in glyphsLib currently.. There is already a component.parent that points to the GSLayer that component is in.

> Mostly

the devil is in the detail and I need to know _exactly_ rather than "mostly"

> a matching layer

matching in what sense?

anthrotype:
> There is already a component.parent that points to the GSLayer that component is in.

looks like GSComponent.layer and GSComponent.parent are aliases and mean the same thing:

```python
l = Glyphs.font.selectedLayers[0]
assert l.components[0].parent is l.components[0].layer
```

schriftgestalt:
The `.layer` and `.parent` is the same in most cases, only for GSNode it is different (`.parent` is a path).

schriftgestalt:
>> Mostly

> the devil is in the detail and I need to know exactly rather than "mostly"

As I said. doing the interpolations for brace layers and smart components.

>> a matching layer

>matching in what sense?

When the component is in a "Color 1" or "[300<wg]" layer and the glyph has a matching layer. It uses a `layer.layerKey` (that is basically what you see in the layer panel) and `glyph.layerForKey_masterId_()`

anthrotype:
hm what's a "layerKey"?

anthrotype:
> layerKey is what you see in the layer panel

I see.. looks like it's a string that depends from the layer.attributes that are set, which define what kind of layer it is. glyphsLib doesn't currently implement that.

anthrotype:
I added the functionality to fonttools directly (https://github.com/fonttools/fonttools/pull/3027), so I will no longer implement this in glyphsLib. I'll make ufo2ft call the fonttools' computeClipBoxes when building static COLR font and caller (in this case fontmake from glyphsLib-generated sources) does not provide pre-computed clipBoxes.

PR #854, opened 2023-01-27
remove glyphsLib.affine, use fontTools.misc.transform instead

anthrotype:
Fixes https://github.com/googlefonts/glyphsLib/issues/257

The glyphsLib.affine module has different copyright/license and is not actually needed because we already depend on fontTools which has an equivalent module for affine transforms. This removes that and replaces it with fontTools Transform.

The existing tests still pass. Let me know if you want me to add additional tests.


anthrotype:
ugh.. I really don't understand why the regression tests keep failing with spurious diffs where integers are turned to floats e.g. 575 => 575.0:

```diff
-      <point x="575" y="524" type="line"/>
+      <point x="575.0" y="524.0" type="line"/>
```

Any help appreciated. I give up for today...

anthrotype:
oh shoot, I misundestood the unified diff output. Confusingly enough, the minus-prefixed lines in the regression tests diffs are the newly generated files, whereas the plus-prefixed lines are the regression tests' reference files. I was expecting the opposite, i.e. reference or expected files on the left (`---`) and actual generated files on the right (`+++`).
So with this patch, nodes that previously would be turned to all floats after a transform was applied, are now kept as integers if the transform simply translates them by some integer amount without scaling them, which is ok. So the regression tests diffs look good.

khaledhosny (approved):

madig (approved):

anthrotype:
thanks guys

PR #856, opened 2023-02-05
Fix glyph order for custom parameter with partial coverage

khaledhosny:
When `glyphOrder` custom parameter is used with partial glyph coverage, Glyphs will use the .glyphs glyph order for the rest of the glyphs, while ufo2ft will sort the glyphs alphabetically. Fix this inconsistency by always making sure UFO `public.glyphOrder` has full coverage.


khaledhosny:
The regression tests failures are expected since now we always write full `public.glyphOrder`.

khaledhosny:
I changed how regression tests diffs are presents, so that new output appears as additions not subtractions as it always trips me (I’m not opening a new PR for this as here we can easily see the new diff change).

PR #858, opened 2023-02-07
Make defaults follow the handbook (custom parameters)

IvanUkhov:
According to the [handbook], when the platform ID is set to 1, the encoding ID and language ID assume zero values:

> If only platformID is specified as 1, then both encID and langID will be assumed as 0 (Mac Roman, and Mac English).

One slightly ambiguous case is what to do when not only the platform ID is set to 1 but also the encoding ID to some value. But it is arguably save to assume zero still.

[handbook]: https://glyphsapp.com/media/pages/learn/3ec528a11c-1634835554/glyphs-3-handbook.pdf

### DIFF
--- a/Lib/glyphsLib/builder/custom_params.py
+++ b/Lib/glyphsLib/builder/custom_params.py
@@ -653,9 +653,13 @@ class NameRecordParamHandler(AbstractParamHandler):
 
                 if len(identifiers) >= 3:
                     encoding_id = self.parse_decimal(identifiers[2])
+                elif platform_id == 1:
+                    encoding_id = 0
 
                 if len(identifiers) >= 4:
                     language_id = self.parse_decimal(identifiers[3])
+                elif platform_id == 1:
+                    language_id = 0
 
                 return {
                     "nameID": name_id,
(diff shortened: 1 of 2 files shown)


anthrotype:
thanks for this, make sense.

> One slightly ambiguous case is what to do when not only the platform ID is set to 1 but also the encoding ID to some value.

have you tried to see what Glyphs.app does in this case?

IvanUkhov:
Sorry. Missed a few lines in the test, but now it is green. Regarding your question, I’ll check tomorrow as soon as I get access to Glyphs.

IvanUkhov:
I tried with the following settings:

Skärmavbild 2023-02-10 kl  13 38 21

The result was as follows, which matches the assumption of the PR:

```
    ...
    <namerecord nameID="43" platformID="1" platEncID="0" langID="0x0" unicode="True">
      Second
    </namerecord>
    <namerecord nameID="44" platformID="1" platEncID="2" langID="0x0" unicode="True">
      Third
    </namerecord>
    <namerecord nameID="45" platformID="1" platEncID="2" langID="0x3" unicode="True">
      Fourth
    </namerecord>
    ...
    <namerecord nameID="42" platformID="3" platEncID="1" langID="0x409">
      First
    </namerecord>
    ...
```

bramstein:
Looks good to me. Thanks for catching this, I missed that statement in the handbook.

IvanUkhov:
@anthrotype, would you like me to do anything else in this PR?

PR #863, opened 2023-02-23
Reduce the ordering difference with respect to Glyphs

IvanUkhov:
When using `glyphsLib` for manipulating Glyphs files, it produces a lot of unrelated changes due to its serialization order that is different from the one used by Glyphs. This PR is an attempt to reduce the different.

### DIFF
--- a/Lib/glyphsLib/classes.py
+++ b/Lib/glyphsLib/classes.py
@@ -2567,10 +2567,10 @@ class GSComponent(GSBase):
         if self.smartComponentValues:
             writer.writeKeyValue("piece", self.smartComponentValues)
         if writer.format_version > 2:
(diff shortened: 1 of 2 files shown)


anthrotype (approved):

anthrotype:
LGTM (pending tests fixups)

IvanUkhov:
I have updated the tests.

It is important to emphasize that, since there is no documentation for such things, it was inferred manually based on a number of typefaces (Mark Simonson’s collection was used). There could be cases that have not been seen yet but would lead to further refinement of the ordering. To be continued, in other words.

(In general, there are other deviations that are not related to ordering. The most common ones are `infinity` and `"infinity"` as well as lists with and without spaces after commas.)

PR #862, opened 2023-02-22
Serialize instance properties

IvanUkhov:
Closes #861.

PR #864, opened 2023-02-23
Keep keys sorted alphabetically during serialization

IvanUkhov:
A continuation of #862.

### DIFF
--- a/Lib/glyphsLib/classes.py
+++ b/Lib/glyphsLib/classes.py
@@ -2559,15 +2559,17 @@ class segment(list):
 
 class GSComponent(GSBase):
     def _serialize_to_plist(self, writer):
+        # NOTE: The fields should come in alphabetical order.
         writer.writeObjectKeyValue(self, "alignment", "if_true")
         writer.writeObjectKeyValue(self, "anchor", "if_true")
+        if writer.format_version > 2:
+            writer.writeObjectKeyValue(self, "rotation", keyName="angle", default=0)
         writer.writeObjectKeyValue(self, "locked", "if_true")
         if writer.format_version == 2:
             writer.writeObjectKeyValue(self, "name")
         if self.smartComponentValues:
             writer.writeKeyValue("piece", self.smartComponentValues)
         if writer.format_version > 2:
-            writer.writeObjectKeyValue(self, "rotation", keyName="angle", default=0)
             writer.writeObjectKeyValue(
                 self, "position", keyName="pos", default=Point(0, 0)
             )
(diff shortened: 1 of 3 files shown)


Lib/glyphsLib/classes.py
```diff
@@ -2806,16 +2808,13 @@ def parent(self):
...
-        if writer.format_version > 2:
-            writer.writeObjectKeyValue(self, "stem", self.stem != -2)
-        for field in ["other1", "other2", "scale"]:
+        # NOTE: The fields should come in alphabetical order.
+        for field in ["horizontal", "name", "options"]:
+            writer.writeObjectKeyValue(self, field, "if_true")
+        for field in ["origin", "other1", "other2", "place", "scale"]:
             writer.writeObjectKeyValue(self, field)
+        writer.writeObjectKeyValue(self, "settings", "if_true")
+        writer.writeObjectKeyValue(self, "stem", self.stem != -2)
```
IvanUkhov, line 2817: This was a mistaken in the previous PR. It is added regardless of the version.

Lib/glyphsLib/classes.py
```diff
@@ -2825,11 +2824,8 @@ def _serialize_to_plist(self, writer):
             else:
                 writer.writeValue(self.target)
             writer.file.write(";\n")
-        for field in ["settings", "type"]:
-            writer.writeObjectKeyValue(self, field, condition="if_true")
-        if writer.format_version == 2:
-            writer.writeObjectKeyValue(self, "stem", self.stem != -2)
-            writer.writeObjectKeyValue(self, "name", condition="if_true")
+        for field in ["type"]:
```
anthrotype, line 2827: this single-item list feels redundant

Lib/glyphsLib/classes.py
```diff
@@ -2806,16 +2808,13 @@ def parent(self):
 
 class GSHint(GSBase):
     def _serialize_to_plist(self, writer):
-        if writer.format_version > 2:
-            writer.writeObjectKeyValue(self, "name", condition="if_true")
```
IvanUkhov, line 2810: Not clear why the name was treated as an exception in the original version.

Lib/glyphsLib/classes.py
```diff
@@ -2559,15 +2559,17 @@ def bezierMinMax(self, x0, y0, x1, y1, x2, y2, x3, y3):
 
 class GSComponent(GSBase):
     def _serialize_to_plist(self, writer):
+        # NOTE: The fields should come in alphabetical order.
```
anthrotype, line 2562: these comments can be ignored and quickly become obsolete.. Ideally we'd collect various parameters, sort alphabetically and emit them all in the desired order so that the ordering is enforced.

Lib/glyphsLib/classes.py
IvanUkhov: Fixed.

IvanUkhov:
One note is that I have not done this for `GSGlyph`. A lot of things are out of place then when looking from the order perspective.

IvanUkhov:
Added two tests for checking deltas with respect to the official reference, GlyphsFileFormatv2.glyphs and GlyphsFileFormatv3.glyphs. The original files do not capture everything but give a little bit of assurance. One could theoretically start to enrich them and capture more, but it is better done is the repository of the Glyphs SDK.

tests/writer_test.py
```diff
@@ -1227,6 +1237,69 @@ def test_write_background_image(self):
...
+        font = classes.GSFont(path)
+        with open(path, encoding="utf=8") as file:
+            expected = file.read()
+
+        path = os.path.join(self.temporary_path, name)
+        font.save(path)
+        with open(path, encoding="utf=8") as file:
+            actual = file.read()
+
+        exceptions = [
```
anthrotype, line 1253: I suppose we can't do anything about these 'exceptions'?

Lib/glyphsLib/classes.py
anthrotype: I see you +1, so are you going to do anything about this?

tests/writer_test.py
IvanUkhov: Well, we can. It is not so much about ordering. So one could tackle it separately to not mix topics. My objective was to capture the behavior.

Lib/glyphsLib/classes.py
IvanUkhov: I was thinking the same that one has to accumulate what is emitted and re-emit in the right order regardless of the order used in the serialize function. So I like the idea. But no, I was not planning to do anything within this PR. It can be addressed separately.

anthrotype: fair enough

tests/writer_test.py
schriftgestalt: I added several missing keys to the GlyphsFileFormatv3.glyphs. I also added some (hacked together) skripts that help getting all keys from a bunch of files.

PR #865, opened 2023-02-23
3rd attempt at fixing RTL kerning export

anthrotype:
This is adapted from @simoncozens' https://github.com/simoncozens/dumbifier/blob/main/Lib/dumbifier/rtl_kerning.py

see discussion in https://github.com/google/fonts/pull/5924#issuecomment-1441925247

~~The tests _don't_ work though yet...~~ (edit: they do now)

### DIFF
--- a/Lib/glyphsLib/builder/groups.py
+++ b/Lib/glyphsLib/builder/groups.py
@@ -25,6 +25,43 @@ from .constants import (
 )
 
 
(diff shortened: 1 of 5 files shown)


yanone:
@davelab6 asked me for feedback on this PR, and while I already commented somewhere else, I can say is that this works and I’d be super happy to see this merged into main soon. Thanks for all the work.

Issue #868, opened 2023-02-26
Support other alternatives of specifying hints in format 3


IvanUkhov:
As @schriftgestalt mentioned in #864, the official reference file, `GlyphsFileFormatv3.glyphs`, has been updated to capture more. Currently, the library fails to parse the updated file. One of the reasons is that not all alternatives of specifying `origin`, `other1`, `other2`, and `target` in hints seem to be supported. In format 3, they are tuples with not only two but occasionally one, three, and four entries:

https://github.com/schriftgestalt/GlyphsSDK/blob/Glyphs3/GlyphsFileFormat/GlyphsFileFormatv3.md

anthrotype:
possibly duplicate of https://github.com/googlefonts/glyphsLib/issues/817

IvanUkhov:
Close due to #869.

PR #869, opened 2023-02-26
Update the official reference for format 3

### LINKED ISSUE #867: Support cropped background images in format 3
As @schriftgestalt mentioned in #864, the official reference file, `GlyphsFileFormatv3.glyphs`, has been updated to capture more. Currently, the library fails to parse the updated file. One of the reasons is that `crop` in background images does not seem to be supported. Its representation in format 3 is different than the one used in format 2. It is now simply a tuple with four entries.

IvanUkhov:
Closes #867 and #868.

### DIFF
--- a/Lib/glyphsLib/classes.py
+++ b/Lib/glyphsLib/classes.py
@@ -38,6 +38,7 @@ from glyphsLib.parser import Parser
 from glyphsLib.pens import LayerPointPen
 from glyphsLib.types import (
     Point,
+    Pointer,
     Rect,
     Transform,
     UnicodesList,
@@ -262,15 +263,6 @@ class OnlyInGlyphsAppError(NotImplementedError):
         )
 
 
-def parse_hint_target(line=None):
-    if line is None:
-        return None
-    if line[0] == "{":
-        return Point(line)
-    else:
-        return line
-
-
 def transformStructToScaleAndRotation(transform):
     Det = transform[0] * transform[3] - transform[1] * transform[2]
     _sX = math.sqrt(math.pow(transform[0], 2) + math.pow(transform[1], 2))
@@ -369,7 +361,11 @@ class GSBase:
                 transformer=transformer,
             ):
                 if transformer:
-                    if isinstance(value, list) and transformer not in [Point, Rect]:
+                    if isinstance(value, list) and transformer not in [
+                        Point,
+                        Pointer,
+                        Rect,
+                    ]:
                         self[target] = [transformer(v) for v in value]
                     else:
                         self[target] = transformer(value)
@@ -2815,16 +2811,8 @@ class GSHint(GSBase):
             writer.writeObjectKeyValue(self, field)
         writer.writeObjectKeyValue(self, "settings", "if_true")
         writer.writeObjectKeyValue(self, "stem", self.stem != -2)
-        if writer.format_version == 2:
-            writer.writeObjectKeyValue(self, "target")
-        elif self.target:
-            writer.writeKey("target")
-            if isinstance(self.target, list):
-                writer.file.write("(%i,%i)" % (self.target[0], self.target[1]))
-            else:
-                writer.writeValue(self.target)
-            writer.file.write(";\n")
-        writer.writeObjectKeyValue(self, "type", "if_true")
+        for field in ["target", "type"]:
+            writer.writeObjectKeyValue(self, field, "if_true")
 
     _defaultsForName = {
         # TODO: (jany) check defaults in glyphs
@@ -2835,13 +2823,6 @@ class GSHint(GSBase):
         "scale": None,
         "stem": -2,
     }
-
-    def _parse_target_dict(self, parser, value):
-        # In glyphs 2 this is a string, in glyphs 3 it is a point or string
-        if isinstance(value, str):
-            self._target = parse_hint_target(value)
-        else:
-            self._target = Point([int(x) for x in value])
 
     def __init__(self):
         self.horizontal = False
@@ -2993,9 +2974,10 @@ class GSHint(GSBase):
 
 GSHint._add_parsers(
     [
-        {"plist_name": "origin", "object_name": "_origin", "converter": Point},
-        {"plist_name": "other1", "object_name": "_other1", "converter": Point},
-        {"plist_name": "other2", "object_name": "_other2", "converter": Point},
+        {"plist_name": "origin", "object_name": "_origin", "converter": Pointer},
+        {"plist_name": "other1", "object_name": "_other1", "converter": Pointer},
+        {"plist_name": "other2", "object_name": "_other2", "converter": Pointer},
+        {"plist_name": "target", "object_name": "_target", "converter": Pointer},
         {"plist_name": "place", "converter": Point},
         {"plist_name": "scale", "converter": Point},
         {"plist_name": "horizontal", "converter": bool},
--- a/Lib/glyphsLib/classes.py
+++ b/Lib/glyphsLib/classes.py
@@ -37,8 +37,8 @@ from fontTools.pens.pointPen import (
 from glyphsLib.parser import Parser
 from glyphsLib.pens import LayerPointPen
 from glyphsLib.types import (
+    IndexPath,
     Point,
-    Pointer,
     Rect,
     Transform,
     UnicodesList,
@@ -362,8 +362,8 @@ class GSBase:
             ):
                 if transformer:
                     if isinstance(value, list) and transformer not in [
+                        IndexPath,
                         Point,
-                        Pointer,
                         Rect,
                     ]:
                         self[target] = [transformer(v) for v in value]
@@ -2974,10 +2974,10 @@ class GSHint(GSBase):
 
 GSHint._add_parsers(
     [
-        {"plist_name": "origin", "object_name": "_origin", "converter": Pointer},
-        {"plist_name": "other1", "object_name": "_other1", "converter": Pointer},
-        {"plist_name": "other2", "object_name": "_other2", "converter": Pointer},
-        {"plist_name": "target", "object_name": "_target", "converter": Pointer},
+        {"plist_name": "origin", "object_name": "_origin", "converter": IndexPath},
+        {"plist_name": "other1", "object_name": "_other1", "converter": IndexPath},
+        {"plist_name": "other2", "object_name": "_other2", "converter": IndexPath},
+        {"plist_name": "target", "object_name": "_target", "converter": IndexPath},
         {"plist_name": "place", "converter": Point},
         {"plist_name": "scale", "converter": Point},
         {"plist_name": "horizontal", "converter": bool},
--- a/Lib/glyphsLib/classes.py
+++ b/Lib/glyphsLib/classes.py
@@ -369,7 +369,7 @@ class GSBase:
                 transformer=transformer,
             ):
                 if transformer:
-                    if isinstance(value, list) and transformer != Point:
+                    if isinstance(value, list) and transformer not in [Point, Rect]:
                         self[target] = [transformer(v) for v in value]
                     else:
                         self[target] = transformer(value)
--- a/Lib/glyphsLib/types.py
+++ b/Lib/glyphsLib/types.py
@@ -13,25 +13,29 @@
 # limitations under the License.
 
 
-import re
-import datetime
-import copy
 import binascii
+import copy
+import datetime
+import re
+from typing import List
+from typing import Optional
+from typing import Union
 
 __all__ = [
-    "Transform",
+    "BinaryData",
     "Point",
+    "Pointer",
     "Rect",
     "Size",
+    "Transform",
+    "UnicodesList",
     "ValueType",
-    "parse_datetime",
-    "parse_color",
     "floatToString3",
     "floatToString5",
-    "readIntlist",
-    "UnicodesList",
-    "BinaryData",
+    "parse_color",
+    "parse_datetime",
     "parse_float_or_int",
+    "readIntlist",
 ]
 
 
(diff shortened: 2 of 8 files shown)


tests/writer_test.py
```diff
@@ -1237,69 +1226,6 @@ def test_write_background_image(self):
             ),
         )
 
-    def test_write_official_reference_v2(self):
```
IvanUkhov, line 1240: Moved to `main_test.py`.

tests/writer_test.py
```diff
@@ -1237,69 +1226,6 @@ def test_write_background_image(self):
...
-        name = "GlyphsFileFormatv2.glyphs"
-
-        path = os.path.join(DATA, name)
-        font = classes.GSFont(path)
-        with open(path, encoding="utf=8") as file:
-            expected = file.read()
-
-        path = os.path.join(self.temporary_path, name)
-        font.save(path)
-        with open(path, encoding="utf=8") as file:
```
IvanUkhov, line 1250: Note a typo here and below.

tests/writer_test.py
```diff
@@ -1318,13 +1244,5 @@ def test_dumps(self):
         self.assertTrue(string)
 
 
-class WriterRoundtripTest(unittest.TestCase, test_helpers.AssertParseWriteRoundtrip):
```
IvanUkhov, line 1321: Redundant.

tests/writer_test.py
```diff
@@ -1237,69 +1226,6 @@ def test_write_background_image(self):
...
-        ]
-        actual = [
-            line
-            for line in actual.splitlines()
-            if not any(line == exception for exception in exceptions)
-        ]
-
-        assert actual == expected
-
-    def test_write_official_reference_v3(self):
```
IvanUkhov, line 1288: Exists in `main_test.py`.

tests/main_test.py
```diff
@@ -58,35 +61,65 @@ def test_parser_main(capsys):
     """This is both a test for the "main" functionality of glyphsLib.parser
     and for the round-trip of GlyphsUnitTestSans.glyphs.
     """
-    filename = os.path.join(os.path.dirname(__file__), "data/GlyphsUnitTestSans.glyphs")
+    filename = os.path.join(DATA, "GlyphsUnitTestSans.glyphs")
     with open(filename) as f:
         expected = f.read()
 
     glyphsLib.parser.main([filename])
-    out, _err = capsys.readouterr()
```
IvanUkhov, line 66: Unifying with the other tests.

Lib/glyphsLib/classes.py
```diff
@@ -2993,9 +2974,10 @@ def other2(self, other2):
 
 GSHint._add_parsers(
     [
-        {"plist_name": "origin", "object_name": "_origin", "converter": Point},
-        {"plist_name": "other1", "object_name": "_other1", "converter": Point},
-        {"plist_name": "other2", "object_name": "_other2", "converter": Point},
+        {"plist_name": "origin", "object_name": "_origin", "converter": Pointer},
```
IvanUkhov, line 2977: This is probably a breaking change if it is accessed from outside the library.

tests/data/GlyphsFileFormatv2.diff
```diff
@@ -0,0 +1,350 @@
+  {
```
IvanUkhov, line 1: There are no surprises here. It has not changed.

tests/data/GlyphsFileFormatv3.diff
```diff
@@ -0,0 +1,1679 @@
...
+  name = "Import Font";
+  value = {
+  Path = "files/LinkedFontv3.glyphs";
+  };
+  },
+  {
+  name = "Color Palettes";
+  value = (
+  (
++ (255,0,0,255),
```
IvanUkhov, line 46: How glyphsLib is doing it matches the Markdown description. Not sure why Glyphs produced it this way.

https://github.com/schriftgestalt/GlyphsSDK/blob/Glyphs3/GlyphsFileFormat/GlyphsFileFormatv3.md

tests/data/GlyphsFileFormatv3.diff
```diff
@@ -0,0 +1,1679 @@
...
+  name = somePrefix;
+  }
+  );
+  features = (
+  {
+  code = "sub C by D;
+  ";
++ notes = "a feature note.";
+  tag = test;
+- notes = "a feature note.";
```
IvanUkhov, line 86: A divergent ordering. Could be fixed.

tests/data/GlyphsFileFormatv3.diff
```diff
@@ -0,0 +1,1676 @@
...
+  name = "Import Font";
+  value = {
+  Path = "files/LinkedFontv3.glyphs";
+  };
+  },
+  {
+  name = "Color Palettes";
+  value = (
+  (
+- (255,0,0,255),
```
IvanUkhov, line 46: This contradicts the spec. Should be fixed.

tests/data/GlyphsFileFormatv3.diff
```diff
@@ -0,0 +1,1676 @@
...
+  code = "";
+  disabled = 1;
+  name = somePrefix;
+  }
+  );
+  features = (
+  {
+  code = "sub C by D;
+  ";
++ tag = test;
```
IvanUkhov, line 84: One reordering opportunity. Should be fixed.

tests/data/GlyphsFileFormatv3.diff
```diff
@@ -0,0 +1,1676 @@
...
+  {
+  horizontal = 1;
+  origin = (0,0);
+  target = (0,7);
+  type = Stem;
+  },
+  {
+  horizontal = 1;
+- place = (-46, 58);
++ place = (-46,58);
```
IvanUkhov, line 390: Note sure why Glyphs produces it without spaces. @schriftgestalt, would you be able to assist? Should be be with or without spaces when formatting points? Thank you!

tests/data/GlyphsFileFormatv3.diff
```diff
@@ -0,0 +1,1676 @@
...
+  {
+  attr = {
+  color = 1;
+  };
+  layerId = m01;
+  shapes = (
+  {
+  attr = {
+- fillColor = (255,211,55,255);
++ fillColor = (
```
IvanUkhov, line 1327: Another instance deviating from the spec.

tests/data/GlyphsFileFormatv3.diff
```diff
@@ -0,0 +1,1676 @@
...
+  {
+  associatedMasterId = m01;
+  attr = {
+  coordinates = (
+  450
+  );
+  };
+  layerId = "403102FA-569E-44B5-9F7D-836F0E7D2DAD";
+  name = "25. Feb. 23, 15:21";
++ visible = 1;
```
IvanUkhov, line 1423: Reordering opportunity. Should be fixed.

tests/data/GlyphsFileFormatv3.diff
```diff
@@ -0,0 +1,1676 @@
...
+  }
+  );
+  instances = (
+  {
++ axesValues = (
++ 100,
++ 100
++ );
+- name = Regular;
++ variablename = Regular;
```
IvanUkhov, line 1472: Probably a deviation from the spec.

tests/data/GlyphsFileFormatv3.diff
```diff
@@ -0,0 +1,1676 @@
...
+  },
+  {
+  disabled = 1;
+  name = "Remove Glyphs";
+  value = (
+  B
+  );
+  }
+  );
+- exports = 0;
```
IvanUkhov, line 1540: Another reordering.

IvanUkhov:
@anthrotype, please take a careful look at the two diff files. I do not see anything that should be fixed within this PR; however, there are many instances to fix in general.

Lib/glyphsLib/types.py
```diff
@@ -405,3 +409,46 @@ def plistValue(self, format_version=2):
         # TODO write hex bytes in chunks and split over multiple lines
         # for better readability, like the fonttools xmlWriter does
         return "<%s>" % binascii.hexlify(self).decode()
+
+
+class Pointer(ValueType):
```
anthrotype, line 418: Is this class named like that in Glyphs.app's python API as well? just curious

IvanUkhov: Mmm, no. I just came up with something that would capture the idea that we are pointing at where to attach. I do not think there is a dedicated `GSSomething` for this in the API.

anthrotype: ah. Well, "Pointer" has quite a baggage... Can we named it something else? It looks like this represents a variable-length sequence of int or str, am I right? Maybe call it `Sequence`. Or if it's only ever used for GSHints, maybe give it a  name that describes what it's used for..

anthrotype: what type do you get for an hint's `origin` or `target` if you print from within Glyphs.app macro window?

anthrotype:
> however, there are many instances to fix in general.

as long as we have specific issues tracking each of those, I suppose it should be fine

tests/data/GlyphsFileFormatv3.diff
anthrotype: > deviating from the spec

which "spec"?

and can you please clarify what the minus and the plus lines are supposed to represent in the diff?

tests/data/GlyphsFileFormatv3.diff
schriftgestalt: From the Spec:
> Some lists (enclosed in normal parenthesis) are put on one line (e.g., **(100,200)**). This is to make it easier to read and save space. We call it tuple for now.
> - this is used for colors, points and nodes

This shows it without spaces.

Lib/glyphsLib/types.py
schriftgestalt: Those are an objc class called "[NSIndexPath](https://developer.apple.com/documentation/foundation/nsindexpath?language=objc)". It is a list of indexes. They are used to identify things in a tree of objects (like selections in an outlineview.

tests/data/GlyphsFileFormatv3.diff
schriftgestalt: And it is a bit tricky to follow all those comments. Can we find a better way to talk about this? Maybe we can do a zoom (or whatever service you prefer)?

IvanUkhov: Thank you for the answer. This is exactly my point that, according to the spec, it should be without spaces, which is what glyphsLib produces. However, it is with spaces in the original file. (The red chunks are what removed by glyphsLib relative to `GlyphsFileFormatv3.glyphs`, and the green ones are what has been added by glyphsLib.)

Here it is:

https://github.com/schriftgestalt/GlyphsSDK/blob/8639eaf79645e664be06e001fe0417cc74689da7/GlyphsFileFormat/GlyphsFileFormatv3.glyphs#L376

So I am not sure what is happening here. Maybe `place` is of a different type?

tests/data/GlyphsFileFormatv3.diff
IvanUkhov: Exactly. glyphsLib does not follow the description about colors and tuples in the spec:

https://github.com/schriftgestalt/GlyphsSDK/blob/8639eaf79645e664be06e001fe0417cc74689da7/GlyphsFileFormat/GlyphsFileFormatv3.md

IvanUkhov: > and can you please clarify what the minus and the plus lines are supposed to represent in the diff?

Sure! The red chunks (the ones with minuses) are what removed by glyphsLib relative to GlyphsFileFormatv3.glyphs, and the green ones (the ones with pluses) are what has been added by glyphsLib.

Lib/glyphsLib/types.py
IvanUkhov: OK, I can rename it to IndexPath.

IvanUkhov: I am reading the documentation. It does not seem to match the description of `NSIndexPath`. According to

https://github.com/schriftgestalt/GlyphsSDK/blob/Glyphs3/GlyphsFileFormat/GlyphsFileFormatv3.md

the tuple can occasionally contain only one element in which case it is a string equal to `lsb` or `rsb` for `origin` or `up` and `down` for `target`. But conceptually, those for strings are kind of indices.

I have renamed to `Index` and added documentation. What do you think?

schriftgestalt: That it is implemented as an NSIndexPath is not important. 

And in the python (and objc) API, it is always an index path, with some special indexes denoting the "lsb", "rsb" … values. Only in the written file it is converted to those strings.

IvanUkhov: I agree that it is not important and thank you for clarifying. Then it is adequate, but let me rename to IndexPath to be crystal clear.

tests/data/GlyphsFileFormatv3.diff
IvanUkhov: > Can we find a better way to talk about this? Maybe we can do a zoom (or whatever service you prefer)?

Yes, I would be happy to have a call to discuss if you prefer.

schriftgestalt: I’m on CET. I’m generally available between 9am to 17pm.

IvanUkhov: I am in the same time zone. Please let me know what it would be convenient for you, and I will schedule a call.

Lib/glyphsLib/types.py
```diff
@@ -405,3 +413,58 @@ def plistValue(self, format_version=2):
...
+        elif value3 is not None:
+            self.value = [value, value2, value3]
+        elif value2 is not None:
+            self.value = [value, value2]
+        elif isinstance(value, (list, tuple)):
+            self.value = value
+        else:
+            self.value = self.fromString(value)
+
+    def fromString(self, string: str) -> List[Union[int, str]]:
```
IvanUkhov, line 449: @schriftgestalt, I have discovered one interesting case, which seems to contradict the Markdown file. It says that `origin` and `target` are `tuple`'s, even when they contain a single string. So it should be `target = (up)`, `target (down)`, etc. However, this is not what I see in Glyphs files. Unfortunately, this is not covered in `GlyphsFileFormatv3.glyphs`. Now I am not sure what is right. It seems that `origin` and `target` are not tuples when a string is the only element, in which case it would be `target = up`, `target = down`, etc. If this is the case, it would be nice to clarify in the Markdown document.

tests/data/GlyphsFileFormatv3.diff
IvanUkhov: Thank you for the call earlier today!

@anthrotype, to keep you updated, the above is due to a typo in Glyphs. The behavior exhibited by glyphsLib here is correct. So no action needed. The reference file will be updated accordingly.

PR #871, opened 2023-02-27
Propagate custom instance-level naming-table entries to UFO

### LINKED ISSUE #870: Support custom instance-level naming-table entries upon export to UFO
`Name Table Entry`'s on the file level (the Font tab) are properly propagated to the resulting binaries. However, such entries on the level of the individual instances (the Exports tab) are disregarded. The expected behavior is that the two lists should be concatenated upon compilation. This is how Glyphs itself exports.

IvanUkhov:
Closes #870.


anthrotype:
Thanks

PR #872, opened 2023-02-28
Fix a typo in the default behavior of naming-table entries

IvanUkhov:
There is arguably the following error in both the Glyphs [handbook](https://glyphsapp.com/media/pages/learn/3ec528a11c-1634835554/glyphs-3-handbook.pdf) and the UI:

> If not specified, platformID will be assumed as 3, and successively, encID as 1 (Unicode), and langID as 0×0049 (Windows English).

It is meant to say that the language is set to 0x409. This is what Glyphs does upon export. Moreover, 0x49 is not a valid language ID as far as the [specification](https://learn.microsoft.com/en-us/typography/opentype/spec/name#windows-language-ids) is concerned.

/cc @schriftgestalt


anthrotype:
no bother at all, please ping me in case I miss anything

PR #874, opened 2023-03-01
Update the reference for version 3

IvanUkhov:


tests/data/GlyphsFileFormatv3.diff
```diff
@@ -386,13 +385,11 @@
   },
   {
   horizontal = 1;
-- place = (-46, 58);
-+ place = (-46,58);
+  place = (-46,58);
```
IvanUkhov, line 388: Does not deviate anymore and the same below.

anthrotype:
thanks

Issue #866, opened 2023-02-24
Erroneous duplication of components in certain color layers of COLRv0 fonts


yanone:
Working on a Cairo Play COLRv0 update ([my fork here](https://github.com/yanone/Cairo)), I found that a handful of ligatures are painted red over an identical black outline. You can see a slight black bleed around the red shape:

<img width="161" alt="Bildschirm­foto 2023-02-24 um 11 46 31" src="https://user-images.githubusercontent.com/175386/221159910-d7dc7e8f-23b1-4ff2-93e4-48036530e2b5.png">

Most ligatures in question are based on the same shape (toothletter+noon), but one other shape is painted in purple, too. So by all accounts I need to describe the occurence as _seemingly random_. I couldn't see any pattern in the source that could explain why components in these glyphs are duplicated and in others not.

<img width="968" alt="Bildschirm­foto 2023-02-24 um 11 48 08" src="https://user-images.githubusercontent.com/175386/221160811-d03b888f-25f1-4d9a-b35d-9ebcc5c2adac.png">

Investigating the files I found that the intermediate UFOs show a total of two components in the `noon_noon-ar.fina.color1.glif` glyph which is supposed to only hold the red dot but now also holds the base component, too, a duplication of the component in the `.color0` glyph, or components being completely identical to the non-color glyph, whichever way you'd like to look at it.

```xml
<?xml version='1.0' encoding='UTF-8'?>
<glyph name="noon_noon-ar.fina.color1" format="2">
  <advance width="993"/>
  <unicode hex="FC8D"/>
  <outline>
    <component base="alefMaksura_noon-ar.fina"/>
    <component base="dotabove-ar" xOffset="787" yOffset="-20"/>
  </outline>
  <lib>
    <dict>
      <key>com.schriftgestaltung.Glyphs.ComponentInfo</key>
      <array>
        <dict>
          <key>anchor</key>
          <string>top_1</string>
          <key>index</key>
          <integer>1</integer>
          <key>name</key>
          <string>dotabove-ar</string>
        </dict>
      </array>
    </dict>
  </lib>
</glyph>
```

As always, I'm happy to assist any debugging.

anthrotype:
cc @khaledhosny who implemented exports of COLRv0 layers in glyphsLib

PR #875, opened 2023-03-06
Fix handling of COLRv0 layers

khaledhosny:
Glyphs handles components for color layers in a special way. If the component glyph has color layers of its own, the component use the first color layer with the same color index, otherwise it fallback to the default layer. We try to do that here as well.

Fixes https://github.com/googlefonts/glyphsLib/issues/866

### DIFF
--- a/Lib/glyphsLib/builder/components.py
+++ b/Lib/glyphsLib/builder/components.py
@@ -46,22 +46,28 @@ def to_ufo_components(self, ufo_glyph, layer):
         to_ufo_components_nonmaster_decompose(self, ufo_glyph, layer)
         return
 
-    # The same for color layers.
-    if (
-        not self.minimal
-        and layer.components
-        and layer.layerId != layer.associatedMasterId
-        and (layer._is_color_palette_layer())
-    ):
-        logger.warning(
-            f"Glyph '{ufo_glyph.name}': All components of the color layer "
-            f"'{layer.name}' will be decomposed."
-        )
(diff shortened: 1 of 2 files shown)


khaledhosny:
@yanone can you test this? I didn’t test with Cairo, but I tested with another font of mine that had a similar issue.

yanone:
@khaledhosny Works for Cairo Play 👍 Thank you

anthrotype:
Thanks Khaled!

Lib/glyphsLib/builder/components.py
```diff
@@ -46,30 +46,36 @@ def to_ufo_components(self, ufo_glyph, layer):
...
+            ]
+            for i, l in enumerate(color_layers):
+                if l._color_palette_index() == layer._color_palette_index():
+                    if l.layerId != l.associatedMasterId:
+                        # If it is not a master layer, we rename it in
+                        # _to_ufo_color_palette_layers(), so we reference the
+                        # same name here.
+                        component_name += f".color{i}"
+                    break
+            print(ufo_glyph.name, component_name)
```
anthrotype, line 70: I presune this `print` was a leftover? I'll delete it

anthrotype: ah you already did remove it, thank you!
https://github.com/googlefonts/glyphsLib/commit/896817d12798f9bddc374392913c0ecd1dd3dce7

PR #881, opened 2023-03-16
Account for preferred names in instance properties

IvanUkhov:
As we know, Glyphs 3 has switched from custom parameters to properties for some items. In this PR, we lay the foundation for fully-fledged processing of properties, in addition to custom parameters. Similarly to custom parameters, they now have a proxy for streamlining operations, such as lookup. In addition, properties are now exposed to `InstanceDescriptorAsGSInstance` for enabling the existing functionality to leverage them. Two items, which were our original motivation for these changes, are processed via this pipeline: preferred family and subfamily names.


IvanUkhov:
@anthrotype, it is ready for review, and I would be happy to address your feedback. Thank you.

Lib/glyphsLib/classes.py
```diff
@@ -3227,46 +3283,59 @@ def exports(self):
...
-        if value:
-            return value
-        return self.parent.familyName
 
     @familyName.setter
     def familyName(self, value):
         self.customParameters["familyName"] = value
 
+    @property
+    def postScriptFontName(self):
```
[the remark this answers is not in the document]
IvanUkhov: Properties are collection of alternative values for different languages. These getters are for fetching only one, which is considered the default. Before, it was called `defaultValue` in `GSFontInfoValue`.

Lib/glyphsLib/classes.py
```diff
@@ -3227,46 +3283,59 @@ def exports(self):
...
-        if value:
-            return value
-        return self.parent.familyName
 
     @familyName.setter
     def familyName(self, value):
         self.customParameters["familyName"] = value
 
+    @property
+    def postscriptFontName(self):
```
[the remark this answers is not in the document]
anthrotype: I don't know, I'd rather try to stick with whatever the Glyphs.app scripting api does, which this glyphsLib.classes is designed to replicate more or less.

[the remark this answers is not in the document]
IvanUkhov: I have removed that function.

anthrotype:
> Glyphs 3 has switched from custom parameters to properties for some items

can we find the full list of parameters that have been migrated to properties?

Lib/glyphsLib/classes.py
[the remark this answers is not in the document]
IvanUkhov: Please note I added also `preferredFamilyName` and forwarded `preferredFamily` to it. I thought it was a typo. Please let me know if you want me to remove `preferredFamilyName`.

IvanUkhov:
That would have been nice. I think if there is someone who knows, it must be @schriftgestalt.

Lib/glyphsLib/classes.py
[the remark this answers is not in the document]
anthrotype: looks like GSInstance has both of them for some reasons... Do they do the same? Might be worth trying from within Glyphs.app's Macro window

https://docu.glyphsapp.com/#GSInstance.preferredFamily
https://docu.glyphsapp.com/#GSInstance.preferredFamilyName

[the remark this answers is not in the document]
IvanUkhov: @schriftgestalt, would you be able to help us here? Thank you.

Lib/glyphsLib/classes.py
```diff
@@ -3227,46 +3283,54 @@ def exports(self):
...
     @preferredFamily.setter
     def preferredFamily(self, value):
+        self.preferredFamilyName = value
+
+    @property
+    def preferredFamilyName(self):
+        return (
+            self.properties.get("preferredFamilyNames")
+            or self.customParameters["preferredFamily"]
+            or self.parent.familyName
```
anthrotype, line 3313: @IvanUkhov I just tried to do the following from the Macro window in Glyphs.app for a font which doesn't define any localised "Typographic Family Names":

```python
print(Glyphs.font.instances[0].preferredFamily)
print(Glyphs.font.instances[0].preferredFamilyName)
```

I got:

```
FlippedSmartComponents
None
```

Which means that the `preferredFamilyName` getter should return None if there are no such properties; whereas the `preferredFamily` returs the default `familyName` even if the properites aren't set.

anthrotype (approved):
I think this is an improvement overall, so LGTM

Just note one little comment about the difference b/w preferredFamily/preferredFamilyName https://github.com/googlefonts/glyphsLib/pull/881#discussion_r1140453769

Lib/glyphsLib/classes.py
IvanUkhov: Thank you for looking into this. I have restored `preferredFamily` and reduced `preferredFamilyName` to not add uncertain functionality unnecessarily.

Lib/glyphsLib/classes.py
```diff
@@ -3227,46 +3283,47 @@ def exports(self):
...
+        # depending on the version.
         self.customParameters["familyName"] = value
 
     @property
     def preferredFamily(self):
-        value = self.customParameters["preferredFamily"]
-        if value:
-            return value
-        return self.parent.familyName
+        return self.customParameters["preferredFamily"] or self.parent.familyName
```
anthrotype, line 3302: from my experiments, I believe this should be like below, because it takes into account the properties if present, otherwise falls back to the default familyName:

```suggestion
        return self.preferredFamilyName or self.parent.familyName
```

Lib/glyphsLib/classes.py
```diff
@@ -3227,46 +3283,47 @@ def exports(self):
...
     @property
     def preferredSubfamilyName(self):
-        value = self.customParameters["preferredSubfamilyName"]
-        if value:
-            return value
-        return self.name
+        return (
+            self.properties.get("preferredSubfamilyNames")
+            or self.customParameters["preferredSubfamilyName"]
+            or self.name
```
anthrotype, line 3320: same here actually...

```python
print(Glyphs.font.instances[0].name)
print(Glyphs.font.instances[0].preferredSubfamilyName)
```

gives me

```
Regular
None
```

Lib/glyphsLib/classes.py
anthrotype: thanks, please see my other comments below

Lib/glyphsLib/classes.py
IvanUkhov: Changed.

Lib/glyphsLib/classes.py
IvanUkhov: Changed. Uncomfortable that no test had to be updated.

anthrotype: yeah.. let's say these accessors are provided at best-effort basis.. You're welcome to add tests of course

schriftgestalt:
```
familyNames,
designers,
designerURL,
manufacturers,
manufacturerURL,
copyrights,
versionString,
vendorID,
uniqueID,
licenses,
licenseURL,
trademarks,
descriptions,
sampleTexts,
postscriptFullNames,
postscriptFullName,
postscriptFontName,
compatibleFullNames,
styleNames,
styleMapFamilyNames,
styleMapStyleNames,
preferredFamilyNames,
preferredSubfamilyNames,
variableStyleNames,
WWSFamilyName,
WWSSubfamilyName,
variationsPostScriptNamePrefix,
```
The keys with and "s" suffix are supposed to be localized.

anthrotype:
thanks Georg!

Lib/glyphsLib/classes.py
IvanUkhov: Ah, already merged. But I will add tests for this in a separate PR still!

IvanUkhov:
Thank you, Georg! I will think about how to accommodate this.

Issue #882, opened 2023-03-17
should reverse contour when decomposing smart components with flipping transform


anthrotype:
When we decompose a smart component, and this has a transformation that would flip it either horizontally or vertically, the contour winding direction also get flipped and, unless we reverse that again while decomposing it, it may be rendered incorrectly. See screenshot below (top is exported from Glyphs.app which reverses the contour when smart component is flipped, bottom is fontmake via glyphsLib 6.2.0).

![image](https://user-images.githubusercontent.com/6939968/225899812-13acd9fa-fd0a-4e1c-bf9d-52b412b91d8c.png)

Attached test case:

[FlippedSmartComponents.zip](https://github.com/googlefonts/glyphsLib/files/11001045/FlippedSmartComponents.zip)

So we need to do like we do in ufo2ft decomposeComponents filter, where the determinant is negative, then reverse.

(Thanks to Paolo and Mew for the report and test case)

anthrotype:
FYI, I have a fix -- just need to come up with a way of testing this

PR #883, opened 2023-03-17
[smart_components] Fix flipped component path direction

anthrotype:
Fixes https://github.com/googlefonts/glyphsLib/issues/882


anthrotype:
as expected, the test fails because the path direction turns clockwise after decomposing the x-flipped smart component. It is expected to not change (was originally counter-clockwise). The fix is to reverse direction after applying the transformation to the decomposed smart component's paths. Fix coming in a few seconds.

anthrotype:
> and simple!

yeah. hopefully the existing GSPath.reverse() can be trusted to do the right thing, reversing contours can be tricky sometimes with all the various edge cases (doesn't look like heavily tested)

PR #884, opened 2023-03-18
Add tests for name accessors in GSInstance

IvanUkhov:
As discussed in #881, this PR adds more tests.

### DIFF
--- a/Lib/glyphsLib/classes.py
+++ b/Lib/glyphsLib/classes.py
@@ -3357,14 +3357,12 @@ class GSInstance(GSBase):
 
     @property
     def fontName(self):
-        value = (
+        return (
             self.properties.get("postscriptFontName")
             or self.customParameters["postscriptFontName"]
+            # TODO: Strip invalid characters.
+            or ("".join(self.familyName.split(" ")) + "-" + self.name)
         )
-        if value:
-            return value
-        # TODO: Strip invalid characters.
-        return "".join(self.familyName.split(" ")) + "-" + self.name
 
     @fontName.setter
     def fontName(self, value):
@@ -3374,10 +3372,9 @@ class GSInstance(GSBase):
 
     @property
     def fullName(self):
-        value = self.customParameters["postscriptFullName"]
-        if value:
-            return value
-        return self.familyName + " " + self.name
+        return self.customParameters["postscriptFullName"] or (
+            self.familyName + " " + self.name
+        )
 
     @fullName.setter
     def fullName(self, value):
--- a/tests/builder/instances_test.py
+++ b/tests/builder/instances_test.py
@@ -112,6 +112,74 @@ def test_reencode_glyphs(tmpdir):
     assert ufos[1]["C"].unicode is None
 
 
+def test_glyphs3_names():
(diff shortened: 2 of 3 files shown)


Lib/glyphsLib/classes.py
```diff
@@ -3357,14 +3357,12 @@ def windowsLinkedToStyle(self):
 
     @property
     def fontName(self):
-        value = (
+        return (
```
IvanUkhov, line 3360: Just unifying with the other accessors being tested.

Lib/glyphsLib/classes.py
```diff
@@ -3374,10 +3372,9 @@ def fontName(self, value):
 
     @property
     def fullName(self):
-        value = self.customParameters["postscriptFullName"]
-        if value:
-            return value
-        return self.familyName + " " + self.name
+        return self.customParameters["postscriptFullName"] or (
```
IvanUkhov, line 3375: The same here.

tests/builder/instances_test.py
```diff
@@ -148,19 +216,19 @@ def test_glyphs3_instance_filtering():
...
 def test_glyphs3_instance_properties(tmpdir):
-    expected_num_properties = [0, 0, 1, 1, 1, 1]
+    expected_num_properties = [0, 0, 1, 2, 1, 2]
 
     file = "InstanceFamilyName-G3.glyphs"
     font = glyphsLib.GSFont(os.path.join(DATA, file))
 
-    for instance, num in zip(font.instances, expected_num_properties):
-        assert len(instance.properties) == num
+    for expected, instance in zip(expected_num_properties, font.instances):
```
IvanUkhov, line 224: To have the expected value first for consistency.

tests/builder/instances_test.py
```diff
@@ -148,19 +216,19 @@ def test_glyphs3_instance_filtering():
 
 
 def test_glyphs3_instance_properties(tmpdir):
-    expected_num_properties = [0, 0, 1, 1, 1, 1]
+    expected_num_properties = [0, 0, 1, 2, 1, 2]
```
IvanUkhov, line 219: Enriched the test file for the other test.

Lib/glyphsLib/classes.py
```diff
@@ -3357,14 +3357,12 @@ def windowsLinkedToStyle(self):
 
     @property
     def fontName(self):
-        value = (
+        return (
             self.properties.get("postscriptFontName")
             or self.customParameters["postscriptFontName"]
```
schriftgestalt, line 3362: I would convert all custom parameters on loading the file. That makes all the accessor code much simpler.

IvanUkhov:
@anthrotype, here are the tests.

Lib/glyphsLib/classes.py
anthrotype: > I would convert all custom parameters on loading the file.

I don't disagree it'd be desirable, but upconverting glyphs v2 files to a v3 internal representation upon reading (and down-converting to v2 if requested upon writing) would involve some major refactoring.

PR #886, opened 2023-03-22
Load all .glyphspackage glyphs

simoncozens:
In https://github.com/googlefonts/glyphsLib/issues/846#issuecomment-1478023500, Georg said

> But was in really intended to load the .glif file based on the name? I thought that you should read all files and map them based off the name in the file?

This PR does that: loads all files in the `glyphspackage` directory and then reorders them based on the order in `order.plist`.

### DIFF
--- a/Lib/glyphsLib/parser.py
+++ b/Lib/glyphsLib/parser.py
@@ -104,11 +104,18 @@ def load_glyphspackage(package_dir):
     package = Path(package_dir)
     infofile = package / "fontinfo.plist"
     orderfile = package / "order.plist"
-    glyphorder = openstep_plist.load(open(orderfile, "r", encoding="utf-8"))
-    data = openstep_plist.load(open(infofile, "r", encoding="utf-8"), use_numbers=True)
+
+    glyphorder = []
+    if orderfile.exists():
+        with open(orderfile, "r", encoding="utf-8") as order_fh:
+            glyphorder = openstep_plist.load(order_fh)
+
+    with open(infofile, "r", encoding="utf-8") as info_fh:
+        data = openstep_plist.load(info_fh, use_numbers=True)
     data["glyphs"] = []
     for glyphfile in (package / "glyphs").glob("*.glyph"):
-        glyph = openstep_plist.load(open(glyphfile, "r"), use_numbers=True)
+        with open(glyphfile, "r") as fh:
+            glyph = openstep_plist.load(fh, use_numbers=True)
         data["glyphs"].append(glyph)
     # Sort according to glyphorder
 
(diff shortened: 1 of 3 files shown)


Lib/glyphsLib/parser.py
```diff
@@ -101,17 +101,25 @@ def _parse_dict_into_object(self, res, d):
...
-            openstep_plist.load(open(glyphfile, "r"), use_numbers=True)
+    for glyphfile in (package / "glyphs").glob("*.glyph"):
+        glyph = openstep_plist.load(open(glyphfile, "r"), use_numbers=True)
+        data["glyphs"].append(glyph)
+    # Sort according to glyphorder
+    data["glyphs"] = list(
+        sorted(
+            data["glyphs"],
+            key=lambda x: (
+                x["glyphname"] not in glyphorder,  # unknown glyphs at end
```
anthrotype, line 118: should the unknown glyphs at the end in turn be sorted alphabetically by their string value, to make things deterministic? I am not sure in what order they will be listed by glob.

Lib/glyphsLib/parser.py
```diff
@@ -101,17 +101,25 @@ def _parse_dict_into_object(self, res, d):
 
 
 def load_glyphspackage(package_dir):
-    infofile = os.path.join(package_dir, "fontinfo.plist")
-    orderfile = os.path.join(package_dir, "order.plist")
+    package = Path(package_dir)
+    infofile = package / "fontinfo.plist"
+    orderfile = package / "order.plist"
```
anthrotype, line 106: if in the end the order.plist is (unlike the UFO contents.plist) just way to sort glyphs but not an actual index that fully determines what's in the font, can it be omitted altogether?

simoncozens: Keeping the glyphs in their expected order is important though, isn't it?

anthrotype: well, not as important as keeping it at all. Just curious, does it actually load in Glyphs.app if you omit one?

anthrotype: and then, there also the `glyphOrder` custom parameter that one can set on the GSFont that basically accomplishes the same thing..

anthrotype: but I think our current code expects the order.plist to be present...
we unconditionally do:

```
glyphorder = openstep_plist.load(open(orderfile, "r", encoding="utf-8"))
```

Lib/glyphsLib/parser.py
simoncozens: I fixed this and simplified that horrible lambda.

Lib/glyphsLib/parser.py
anthrotype: by the way, it's good practice when opening a file to close it afterwards, here and below we just open. Ideally we do `with open(...):`

anthrotype: you said

> If there is no order.plist, then Glyphs.app will still load the file

so I'd expect that we match Glyphs.app behavior, whatever their spec says

anthrotype (approved):
LGTM, thanks!

PR #889, opened 2023-04-14
Fall back from properties to custom parameters

IvanUkhov:
Given that `properties` always exists in `GSInstance`s, the conversion to UFO never considers `customParameters` as a source for preferred family names among others. Custom parameters, however, are the primary mechanism in version 2, which means the current code does not work for version 2.

This pull request proposes to always fall back to custom parameters.

### DIFF
--- a/Lib/glyphsLib/builder/custom_params.py
+++ b/Lib/glyphsLib/builder/custom_params.py
@@ -153,13 +153,10 @@ class GlyphsObjectProxy:
         opposed to a master or instance."""
         return hasattr(self._owner, "glyphs")
 
-    def has_properties(self):
-        if self.is_font():
-            return self._owner.format_version > 2
-        return hasattr(self._owner, "properties")
-
     def get_property(self, key):
-        return self._owner.properties.get(key)
+        if key and hasattr(self._owner, "properties"):
+            return self._owner.properties.get(key)
+        return None
 
 
 class UFOProxy:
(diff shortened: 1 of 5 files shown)


tests/builder/custom_params_test.py
```diff
@@ -558,7 +558,7 @@ class SetCustomParamsTestDefcon(SetCustomParamsTestBase, unittest.TestCase):
     ufo_module = defcon
 
 
-def test_ufo_filename_custom_param(ufo_module):
+def test_ufo_filename(ufo_module):
```
IvanUkhov, line 561: Renamed a few tests to clarify what each one is testing.

tests/builder/custom_params_test.py
```diff
@@ -616,50 +616,79 @@ def test_ufo_filename_custom_param_instance_empty(ufo_module):
...
+def test_ufo_opentype_name_preferred_family_subfamily_name():
     from glyphsLib.interpolation import apply_instance_data_to_ufo
 
-    file = glyphsLib.GSFont(os.path.join(DATA, "UFOInstanceParametersTest.glyphs"))
-    space = glyphsLib.to_designspace(file, minimal=True)
+    filenames = [
+        "UFOInstanceParametersTestV2.glyphs",
+        # NOTE: In the format of version 3, the preferred family and subfamily
+        # names are not actually saved in custom paramters but properties.
+        "UFOInstanceParametersTestV3.glyphs",
```
IvanUkhov, line 626: Added it for symmetry but can remove if necessary.

schriftgestalt:
(I didn't read all changes yet.) Just want to make sure the general direction is right: the fallback to custom parameters should be handled when reading the .glyphs file and all available info should be put into the properties. Then all other code only needs to check there.

IvanUkhov:
Yes, this is primarily about reading. It was not taking custom parameters for sources in version 2, as it was stopping at the properties, which were introduced in version 3 but are technically always available in `GSInstance` in Python.

Lib/glyphsLib/builder/custom_params.py
```diff
@@ -251,21 +248,23 @@ def to_ufo(self, builder, glyphs, ufo):
...
-        if small_name_value is not None:
-            return small_name_value
+        value = getter(self.glyphs_name)
+        if value is not None:
+            return value
         if self.glyphs_long_name is not None:
-            return getter(self.glyphs_long_name)
-        return None
+            value = getter(self.glyphs_long_name)
+        return value
```
khaledhosny, line 267: I think what George is saying is that when reading a Glyphs 2 file, we should populate the properties from the corresponding custom parameters, and this code can be simplified to always check for the properties.

IvanUkhov: I see. Yes, this is a better interpretation. Thank you for clarifying 🙂

So the logic should be that, if there is a known property, it should be the only place checked in this code, and that, when parsing, all custom parameters with known properties should go to `properties`. Then when writing files in format 2, one would have to put some properties back to custom parameters instead of ignoring them all, as it is currently.

khaledhosny: I believe so. But it doesn’t have to happen in the PR, I think this change is fine as it is since it fixes an existing bug.

IvanUkhov: Yes, it is more of a refactoring task, which can be tackled separately.

schriftgestalt: Yes. That is correct. 

That simplifies the object API and keeps the conversation code in two place: the reading and the writing function for each object.

khaledhosny (approved):

PR #890, opened 2023-04-15
Merge font selection flags from the font and instance

IvanUkhov:
When font selection flags are specified on the level of both the font (the tab) and instance, the ones in the former get discarded. This pull request switches from resetting to merging (union).

### DIFF
--- a/Lib/glyphsLib/builder/custom_params.py
+++ b/Lib/glyphsLib/builder/custom_params.py
@@ -818,14 +818,14 @@ class OS2SelectionParamHandler(AbstractParamHandler):
         if not use_typo_metrics and not has_wws_name and unsupported_bits is None:
             return
 
-        selection_bits = []
+        selection_bits = ufo.get_info_value("openTypeOS2Selection") or []
         if use_typo_metrics:
             selection_bits.append(7)
         if has_wws_name:
             selection_bits.append(8)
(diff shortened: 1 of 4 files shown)


schriftgestalt:
Glyphs would not do this. It uses the first value it finds. It should use the info from the instances, then from the font, then the default.

IvanUkhov:
I based the implementation on what Glyphs does. Just to be clear, the scenario is that “Use Typo Metrics” is set on the font level, while “Has WWS Name” is set on the instance level. Glyphs has both flags set in the resulting binaries, while glyphsLib has only the last one. It starts to set `openTypeOS2Selection ` from scratch when it encounters either “Use Typo Metrics” or “Has WWS Name” in an instance.

IvanUkhov:
> It uses the first value it finds. It should use the info from the instances, then from the font, then the default.

This is exactly how it is in the pull request. It goes from the font to the instance. Also, since those two are flags represented by integers corresponding to bits set, it is not that one can have it explicitly unset. In other words, if one of the flags is set in the font, it cannot be unset in an instance. Or did I misunderstand it?

schriftgestalt:
Sorry, I didn’t get it that it is build from different parameters. You are right then.

khaledhosny (approved):

Issue #892, opened 2023-04-18
Extracting source with brace layers fails on str to int comparison


kontur:
In a Glyphs source file in format 3 this line [https://github.com/googlefonts/glyphsLib/blob/d93c234f03a6b7452931151797ec8ec50227078e/Lib/glyphsLib/builder/sources.py#L208] fails with a:

> File "/Users/johannes/.pyenv/versions/3.10.4/envs/mrhandy/lib/python3.10/site-packages/glyphsLib/builder/sources.py", line 210, in _to_designspace_source_layer
    brace_layers.sort(key=lambda x: tuple(x.location.values()))
TypeError: '<' not supported between instances of 'str' and 'int'

When I add a `print(brace_layers)` before I get:

```
...
SourceDescriptor(
    filename='Foobar-CondensedBold.ufo',
    path=None,
    name='Foobar Condensed Bold {72, 200}',
    layerName='{72, 200}',
    location=OrderedDict([('Weight', 72), ('Width', 200)]),
    copyLib=False,
    copyGroups=False,
    copyFeatures=False,
    muteKerning=False,
    muteInfo=False,
    mutedGlyphNames=[],
    familyName=None,
    styleName=None,
    localisedFamilyName={},
), SourceDescriptor(
    filename='Foobar-CondensedBold.ufo',
    path=None,
    name='Foobar Condensed Bold {22, 100}',
    layerName='{22, 100}',
    location=OrderedDict([('Weight', '22'), ('Width', '100')]),
    copyLib=False,
    copyGroups=False,
    copyFeatures=False,
    muteKerning=False,
    muteInfo=False,
    mutedGlyphNames=[],
    familyName=None,
    styleName=None,
    localisedFamilyName={},
),
...
```

Note how some locations are extracted as dicts with numeric values, and others as dicts with string values:

```
    layerName='{72, 200}',
    location=OrderedDict([('Weight', 72), ('Width', 200)]),
```
vs.
```
    layerName='{22, 100}',
    location=OrderedDict([('Weight', '22'), ('Width', '100')]),
```

Looking at the brace layer names in Glyphs I haven't been able to find what would cause this difference in interpreting the value as string vs int.

kontur:
Printing the return from [this line](https://github.com/googlefonts/glyphsLib/blob/b845089b1822a53cd81bd351767281ab1ff7b07e/Lib/glyphsLib/classes.py#L3983) gives me:

```
...
BRACE COORDS [152, 100]
[...]

PR #893, opened 2023-04-18
Fix glyphs3 brace coordinates being returned as non-int

kontur:
Possible fix for #892

### DIFF
--- a/Lib/glyphsLib/classes.py
+++ b/Lib/glyphsLib/classes.py
@@ -3980,7 +3980,7 @@ class GSLayer(GSBase):
             return None
 
         if self.parent.parent.format_version > 2:
-            return self.attributes["coordinates"]  # Glyphs 3
+            return (float(v) for v in self.attributes["coordinates"])  # Glyphs 3
 
         # Glyphs 2
         name = self.name


Lib/glyphsLib/classes.py
```diff
@@ -3980,7 +3980,7 @@ def _brace_coordinates(self):
             return None
 
         if self.parent.parent.format_version > 2:
-            return self.attributes["coordinates"]  # Glyphs 3
+            return (int(v) for v in self.attributes["coordinates"])  # Glyphs 3
```
anthrotype, line 3983: why int and not float?

kontur: Went with int instinctively since all the tutorials I've ever seen using the brace trick use int values. Turns out the GlyphsApp UI also coerces int's when you try to input commas or points.

anthrotype: ok.. that's odd, because in fvar axes coordinates are fractional Fixed16Dot16 so I don't see why the editor has to force them to be rounded to integer. Anyway, if you say glyphs wants them to be int, we could at least do `round(float(v))` so we don't simply truncate the float to int maybe. Or whatever Glyphs.app does.

anthrotype: > Turns out the GlyphsApp UI also coerces int's when you try to input commas or points

are you actually sure about that?
I just tried to create an intermediate (brace) layer in Glyphs 3.2 (3189) and it does let me type a float (e.g. 201.5) in the intermediate layer coordinates. I can see it saved back to disk as well. So it seems like floats are supported, as it should.

anthrotype: `float` is the right one, because in the if branch within that same method you're touching that handles the Glyphs 2 variant we do `return [float(c) for c in coordinates.split(",")]`

kontur: Hm, interesting. When I input a float with period I get "invalid value", when I input a "float" with a comma the layer name changes... I'm on latest GlyphsApp 3.2 (3189). Anyway, I concur, float is the right choice.

Screenshot 2023-04-24 at 17 02 11
Screenshot 2023-04-24 at 17 02 19

khaledhosny: Do you happen to be on a locale where comma is the decimal separator?

kontur: I am, but as shown above, the comma doesn't convert to a float either, it adds another supposed axis location, e.g. splitting the `{22, 200}` and it becomes `{22, 5, 200}` when inputting `22,5`, which is totally wrong (of GlyphsApp), too. Either way, I'm not to bothered with what GlyphsApp does as long as the glyphsLib conversion ends up consistent with consistent/usable data for what we see in the GlyphsApp interface. I've not spent more time, though, debugging what leads to (this particular file?) having some layer locations parsed as strings and some as int's.

schriftgestalt: Bildschirmfoto 2023-04-24 um 23 43 12
In Glyphs 3, the numbers use localized decimal separators. This looks a bit confusing in this case. But is should be treated as a float internally and written out as float when writing Glyphs 3 files. But not when writing v2 files. When setting it up in Glyphs 2 it will write whatever you have typed. I need to dig up the Glyphs 2 code to see if it actually used it.

kontur: Ah okay, on second very much closer look I see the space vs no-space. 

In any case, the PR now coerces float.

khaledhosny (approved):

PR #896, opened 2023-04-28
Support glyph property tokens

khaledhosny:
See Glyphs 3 handbook.

Currently only width and anchors properties work, if our GSLayer starts supporting LSB, RSB, TSB, and BSB, they should magically work here as well.


simoncozens (approved):
Nice and clean, LGTM.

PR #897, opened 2023-04-28
Accept NOT and ! in token not only lower case not

khaledhosny:


khaledhosny:

PR #899, opened 2023-04-28
Use fnmatch for LIKE comparison in tokens

khaledhosny:
Using re.match() is too greedy, but fnmatch supports Unix shell-style wildcards that is similar to LIKE. The only difference is that fnmatch supports also [seq] and [!seq], but we can’t have square brackets in predicates anyway.


khaledhosny:

Issue #851, opened 2023-01-20
layer.name causes intermediate sources with duplicate locations


schriftgestalt:
The `layer.name` should not be used for anything. Only when opening v2 files to convert to `layer.attributes`.

In Glyphs 3, `layer.name` is generated from the attributes but is only used for UI purposes. 

I just found a problem where two glyphs with a brace layer for the same position would generate duplicate `SourceDescriptor`s because of different names.

```
WARNING:glyphsLib.builder.sources:DesignSpace sources contain duplicate locations; varLib expects each master to define a unique location. Make sure that you used consistent 'brace layer' names in all the glyph layers that share the same location.
  ['{110, 0}', '25. May 22, 21:39'] => {'Weight': 110, 'Ascender': 0}
```

ktkm:
I have this issue now when building from v3 files. This means I can only use one set of coordinates for one glyph at a time. Is there a workaround?

kontur:
Encountered the same issue. For now I could only find that the Glyphs UI hides the "date name" of the layer when activating it as a e.g. "Intermediate" layer type, so finding what glyph actually causes this new layer in the UFO is a PITA. No idea as of yet where in glyphsLib the layer name is used to result in those in the extracted UFOs and then subsequently results in failed designspaces.

schriftgestalt:
The layer.name should not use at all anymore.

kontur:
The error message is almost word by word the above

> WARNING:glyphsLib.builder.sources:DesignSpace sources contain duplicate locations; varLib expects each master to define a unique location. Make sure that you used consistent 'brace layer' names in all the glyph layers that share the same location.
  ['{110, 0}', '25. May 22, 21:39'] => {'Weight': 110, 'Ascender': 0}

It seems somewhere the glyph layer name makes it into the master/source name. I'm condense my file into something testable.

kontur:
Printing the `duplicates` [here](https://github.com/googlefonts/glyphsLib/blob/6372da0c1f77cad3152364c2473b7a380ac6c301/Lib/glyphsLib/builder/sources.py#L50) I get:

```
{(('Weight', 22.0), ('Width', 100.0)): [SourceDescriptor(
    filename='Foobar-CondensedBold.ufo',
    path=None,
    name='Foobar Condensed Bold {22, 100}',
    layerName='{22, 100}',
    location=OrderedDict([('Weight', 22.0), ('Width', 100.0)]),
    copyLib=False,
    copyGroups=False,
[...]

PR #902, opened 2023-05-03
Normalize file format 3 brace layer names to legacy layer name format for writing to UFO

kontur:
Possible fix for #851

I'm happy to refactor this bit of layer name normalizing to somewhere more appropriate, if there is such a spot. As mentioned in the issue, I reckon the same might be needed for other secondary layer types, e.g. brace.


anthrotype:
closing as #928 got merged, thanks

PR #903, opened 2023-05-15
Fix multiple custom parameters with the same name on GSFont

khaledhosny:
GSFont with multiple custom parameters with the same name would lose all of them except the last one. It seems some of the code introduced in f7b1e07b63e5a51a59eedbfd31d1df4b8af7f64f handles customParameters as a dictionary, but it is this GlyphsApp hybrid list/dictionary and can have multiple entries with the same name.

This fix restores setting custom parameters using
CustomParametersProxy.__setter__() and introduces the special handling of axes custom parameter to this function.

### DIFF
--- a/Lib/glyphsLib/classes.py
+++ b/Lib/glyphsLib/classes.py
@@ -1163,6 +1163,14 @@ class CustomParametersProxy(ListDictionaryProxy):
                 return self._owner.axes
         return super().__contains__(item)
 
+    def __setter__(self, items):
+        axes_params = [i for i in items if i.name == "Axes"]
+        for param in axes_params:
+            self._owner._set_axes_from_custom_parameter(param.value)
+
+        params = [i for i in items if i.name != "Axes"]
+        super().__setter__(params)
+
     def __iter__(self):
         for item in super().__iter__():
             yield item
@@ -4345,7 +4353,6 @@ class GSFont(GSBase):
         "classes": [],
         "features": [],
         "featurePrefixes": [],
-        "customParameters": [],
         "disablesAutomaticAlignment": False,
         "disablesNiceNames": False,
         "gridLength": 1,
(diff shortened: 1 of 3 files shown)


schriftgestalt:
GSInstance.customParameter need to behave the same.

khaledhosny:
> GSInstance.customParameter need to behave the same.

It seems to work file already, but I’ve added a test case to make sure it is working.

simoncozens:
>  it is this GlyphsApp hybrid list/dictionary and can have multiple entries with the same name.

What are the intended semantics of that?!

khaledhosny:
> > it is this GlyphsApp hybrid list/dictionary and can have multiple entries with the same name.
> 
> What are the intended semantics of that?!

There are two issue here:
1. data loss since GlyphsApp allow multiple custom parameters with the same name, but when loaded with glyphsLib only the last one survives. This is what this PR is about.
2. How to interpret such duplicate custom parameters. It depends and this PR changes nothing here, it is already possible to have this in GSFontMaster and GSInstance.

(in my case, I have multiple `Virtual Master` custom parameters and I need to be able to access them all).

schriftgestalt:
> > it is this GlyphsApp hybrid list/dictionary and can have multiple entries with the same name.
> 
> What are the intended semantics of that?!

internally the custom parameters are a list of objects storing a key/value pair. To allow access like a dict is only for convenience.

anthrotype:
> To allow access like a dict is only for convenience.

ok but what does this pseudo-dict getitem actually gets? The value for the first matching key? Does the dict interface also support setting new key/value pairs? If so, what happens in that case (are all existing values under given key replaced, or only the first, or what)?

schriftgestalt:
> ok but what does this pseudo-dict getitem actually gets? The value for the first matching key? Does the dict interface also support setting new key/value pairs? If so, what happens in that case (are all existing values under given key replaced, or only the first, or what)?

The dict getter gives you the first value with a matching key. 

The setter will replace the value in the first custom parameter object if finds and if there is none, it adds a new CP.

khaledhosny:
Any objection to merging this?

anthrotype (approved):

anthrotype (approved):

PR #908, opened 2023-06-08
Store shape indices

### LINKED ISSUE #895: Corner components bug
Corner components are stored as a hint, referring to a path index and a point index:

```
{
name = _corner.large;
origin = (1,12);
type = Corner;
}
```

So when we come to find the relevant node and relevant path while fixing up the UFO, we do this:

```
target_node=glyph[path_idx][(node_idx + 1) % len(glyph[path_idx])],
```

The problem is that the first element of `origin` is not a *path* index. It's actually a *shape* index. I have a font which is failing to build because the glyph contains two shapes, a component and a path, and the corner component above is on the path, which is shape index 1. But ``glyph[1]`` fails, because the UFO glyph only has one contour.

The problem is that by the time we have got to UFO land, we have split the components from the paths, so it's no longer clear what "shape index" 1 refers to.

I suppose that when we write out corner hints to UFO, we have to convert them from the shape index to the path index, which is pretty gross and breaks round-tripping, but there we go.

simoncozens:
This fixes #895. We do this by:

1) Providing a `.shapes` accessor on the GSLayer object, similar to Glyphs 3 API.
2) Storing a "shapes signature" in the UFO lib for mixed path/component glyphs. For example, if the shape order is "path 1, component 1, path 2, path 3, component 2", a signature will be stored in the layer's lib dictionary with the value `"PCPPC"`. 
3) Using this signature to re-arrange the `.shapes` array when loading UFOs back into Glyphs objects.
4) When processing a corner component, mapping the "shape index" (stored in the hints array) to the "path index" using the signature.

### DIFF
--- a/Lib/glyphsLib/builder/constants.py
+++ b/Lib/glyphsLib/builder/constants.py
@@ -60,6 +60,7 @@ UFO_KERN_GROUP_PATTERN = re.compile("^public\\.kern([12])\\.(.*)$")
 LOCKED_GUIDE_NAME_SUFFIX = " [locked]"
 
 HINTS_LIB_KEY = GLYPHS_PREFIX + "hints"
+SHAPE_SIGNATURE_LIB_KEY = GLYPHLIB_PREFIX + "shapeSignature"
 
 SMART_COMPONENT_AXES_LIB_KEY = GLYPHS_PREFIX + "smartComponentAxes"
 
--- a/Lib/glyphsLib/builder/constants.py
+++ b/Lib/glyphsLib/builder/constants.py
@@ -60,7 +60,7 @@ UFO_KERN_GROUP_PATTERN = re.compile("^public\\.kern([12])\\.(.*)$")
 LOCKED_GUIDE_NAME_SUFFIX = " [locked]"
 
 HINTS_LIB_KEY = GLYPHS_PREFIX + "hints"
-SHAPE_SIGNATURE_LIB_KEY = GLYPHLIB_PREFIX + "shapeSignature"
+SHAPE_ORDER_LIB_KEY = GLYPHLIB_PREFIX + "shapeOrder"
 
 SMART_COMPONENT_AXES_LIB_KEY = GLYPHS_PREFIX + "smartComponentAxes"
 
--- a/Lib/glyphsLib/builder/glyph.py
+++ b/Lib/glyphsLib/builder/glyph.py
@@ -18,7 +18,7 @@ import logging
 
 import glyphsLib.glyphdata
 
-from .. import GSLayer
+from .. import GSLayer, GSPath
 from .common import from_loose_ufo_time, to_ufo_time
 from .constants import (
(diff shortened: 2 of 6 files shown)


simoncozens:
Regression tests are expected to fail as this changes the UFO output.

Lib/glyphsLib/builder/constants.py
```diff
@@ -60,6 +60,7 @@
 LOCKED_GUIDE_NAME_SUFFIX = " [locked]"
 
 HINTS_LIB_KEY = GLYPHS_PREFIX + "hints"
+SHAPE_SIGNATURE_LIB_KEY = GLYPHLIB_PREFIX + "shapeSignature"
```
schriftgestalt, line 63: Wouldn’t it be more future prove to put that into `public.objectLibs`?

schriftgestalt: Yes, proof. 

No, but using that key should be more stable when the file is edited in different tools?

Lib/glyphsLib/builder/glyph.py
```diff
@@ -162,6 +163,11 @@ def to_ufo_glyph(self, ufo_glyph, layer, glyph, do_color_layers=True):  # noqa:
     self.to_ufo_hints(ufo_glyph, layer)  # .hints
     self.to_ufo_paths(ufo_glyph, layer)  # .paths
     self.to_ufo_components(ufo_glyph, layer)  # .components
+    # Store shape order for mixed glyphs
+    if layer.paths and layer.components:
+        ufo_glyph.lib[SHAPE_SIGNATURE_LIB_KEY] = "".join(
+            [("P" if isinstance(x, GSPath) else "C") for x in layer.shapes]
```
anthrotype, line 169: maybe check that "C" is acutally a GSComponent? what if a new different of "shape" is introduced later that is neither GSPath nor GSComponent..

Lib/glyphsLib/builder/glyph.py
```diff
@@ -482,6 +488,21 @@ def to_glyphs_glyph(self, ufo_glyph, ufo_layer, master):  # noqa: C901
...
+    if SHAPE_SIGNATURE_LIB_KEY in ufo_glyph.lib:
+        # Reshuffle shapes array to match original shape order
+        new_shapes = []
+        path_counter = 0
+        comp_counter = 0
+        for sign in ufo_glyph.lib[SHAPE_SIGNATURE_LIB_KEY]:
+            if sign == "P":
+                new_shapes.append(layer.paths[path_counter])
+                path_counter += 1
+            else:
```
anthrotype, line 501: elif == "C"... else AssertionError?

Lib/glyphsLib/builder/constants.py
anthrotype, line 63: I don't find "shapeSignature" name particularly clear, how about "shapeOrder"?

anthrotype (approved):

Issue #910, opened 2023-06-12
Add mark filtering to context anchor


MariannaPaszkowska:

anthrotype:
are you going to provide a bit more context to understand what this is about? thanks :)

schriftgestalt:
The anchor context can now have a lookup flag prefix. 
When you could do `C *` before, now you can do: 
```
lookupflag UseMarkFilteringSet @CombiningTopAccents;C *
```
Everything before the semicolon is supposed to be at the beginning of the lookup. I do a dict lookup with the prefix to collect all rules with the same prefix.

simoncozens:
Even without the mark filtering, we don't currently support contextual anchors at all - but I'm pretty sure we will by the end of today...

schriftgestalt:
I thought @khaledhosny had worked on it?

khaledhosny:
No, I didn’t do any work related to contextual anchors.

simoncozens:
I've done 2 and chucked it into a branch (https://github.com/googlefonts/glyphsLib/tree/contextual-mark-writer) so I can use it. It's not ready for PR yet as it doesn't handle the mkmk case.

khaledhosny:
My vote for 2) as well, and it can be also extended to ignore anchors that Glyphs ignores (IIRC any anchors starting with non-letter, other than `*` and `_` of course).

anthrotype:
> set the UFO feature writer lib key to use it

note that setting that lib key overwrites the default list of writers, which means if you still want the default ones to be run you'll have to list them explicitly in the list. We added an `...` placeholder to mean the default writers, but that only works when you use the fontmake CLI option --feature-writers or when you call ufo2ft compile methods and pass the `featureWriters` parameter.

simoncozens:
We have a `DEFAULT_FEATURE_WRITERS` constant in glyphsLib constants, which we check when round-tripping. So I did this: https://github.com/googlefonts/glyphsLib/commit/7dacee459a3a099c8a4ce31327c9bc04111d4817

anthrotype:
yeah but we stopped using that DEFAULT_FEATURE_WRITERS at some point because it was outdated, ufo2ft gained another default writer I belive (gdefFeatureWriter?) and if we hard-coded the list we wouldn't get that

simoncozens:
Or we can grab it from `ufo2ft.featureCompiler.FeatureCompiler.defaultFeatureWriters`.

anthrotype:
[...]

PR #911, opened 2023-06-13
Send anchor userdata to objectlib

simoncozens:
This goes part of the way to fixing #910. It saves any userData on an anchor to the glyph's objectlib dictionary. This allows us to later get at the contextual user data attached to the anchor and generate contextual mark lookups for it.

Ideally we should (a) allow userData on anything which can take a userData on the Glyphs side and serialise it, and (b) convert the UFO objectlibs back to Glyphs objects. But I don't need that today, and I am in a hurry. :-/

### DIFF
--- a/Lib/glyphsLib/builder/anchors.py
+++ b/Lib/glyphsLib/builder/anchors.py
@@ -16,7 +16,7 @@
 from glyphsLib.types import Point
 import uuid
 
-from ufo2ft.constants import OBJECT_LIBS_KEY
+from glyphsLib.builder.constants import OBJECT_LIBS_KEY
 
 __all__ = [
     "to_ufo_glyph_anchors",
--- a/Lib/glyphsLib/builder/anchors.py
+++ b/Lib/glyphsLib/builder/anchors.py
@@ -16,6 +16,8 @@
 from glyphsLib.types import Point
 import uuid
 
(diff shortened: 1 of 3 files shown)


Lib/glyphsLib/builder/anchors.py
```diff
@@ -27,6 +28,10 @@ def to_ufo_glyph_anchors(self, glyph, anchors):
     for anchor in anchors:
         x, y = anchor.position
         anchor_dict = {"name": anchor.name, "x": x, "y": y}
+        if anchor.userData:
+            identifier = str(uuid.uuid4()).upper()
+            anchor_dict["identifier"] = identifier
+            glyph.lib.setdefault("public.objectLibs", {})[identifier] = dict(anchor.userData)
```
anthrotype, line 34: I think we have a module where we keep all the constants, maybe put public.objectLibs in there

anthrotype (approved):
LGTM, pls fix lint

anthrotype:
> Ideally we should (a) allow userData on anything which can take a userData on the Glyphs side and serialise it, and (b) convert the UFO objectlibs back to Glyphs objects. But I don't need that today, and I am in a hurry. :-/

it's ok to do later, we can file an issue so we don't forget

PR #921, opened 2023-06-28
Support contextual anchors

simoncozens:
See #910. This implements the `*top` (etc.) anchors with an embedded custom context, causing them to replace standard anchor positions in certain contexts. Tested fairly extensively on an internal font project, but:

* No test suite yet
* No mkmk handling yet

### DIFF
--- a/Lib/glyphsLib/builder/constants.py
+++ b/Lib/glyphsLib/builder/constants.py
@@ -210,7 +210,6 @@ DEFAULT_FEATURE_WRITERS = [
     {
         "module": "glyphsLib.featureWriters.markFeatureWriter",
         "class": "ContextualMarkFeatureWriter",
-        "options": {"mode": "skip"},
     },
     {"class": "GdefFeatureWriter"},
     {"class": "CursFeatureWriter"},
--- a/Lib/glyphsLib/builder/constants.py
+++ b/Lib/glyphsLib/builder/constants.py
@@ -204,19 +204,15 @@ UFO2FT_COLOR_LAYER_MAPPING_KEY = "com.github.googlei18n.ufo2ft.colorLayerMapping
 UFO2FT_COLOR_LAYERS_KEY = "com.github.googlei18n.ufo2ft.colorLayers"
 
 UFO2FT_META_TABLE_KEY = PUBLIC_PREFIX + "openTypeMeta"
-# ufo2ft KernFeatureWriter default to "skip" mode (i.e. do not write features
-# if they are already present), while Glyphs.app always adds its automatic
-# kerning to any user written kern lookups. So we need to pass custom "append"
-# mode for the ufo2ft KernFeatureWriter whenever the GSFont contain a non-automatic
-# 'kern' feature.
-# See https://glyphsapp.com/tutorials/contextual-kerning
+
 DEFAULT_FEATURE_WRITERS = [
-    {"class": "KernFeatureWriter", "options": {"mode": "append"}},
+    {"class": "KernFeatureWriter"},
     {
         "module": "glyphsLib.featureWriters.markFeatureWriter",
         "class": "ContextualMarkFeatureWriter",
         "options": {"mode": "skip"},
     },
+    {"class": "GdefFeatureWriter"},
     {"class": "CursFeatureWriter"},
 ]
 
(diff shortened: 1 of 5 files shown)


simoncozens:
I have to be honest here and say I've done what I need to make my font project work, and if anyone actually needs this in glyphsLib they're welcome to take it over.

khaledhosny:
I added a simple test and I think this is good enough to merge.

Lib/glyphsLib/builder/constants.py
```diff
@@ -210,14 +210,14 @@
...
-# would need to update this accordingly... :-/
 DEFAULT_FEATURE_WRITERS = [
     {"class": "KernFeatureWriter", "options": {"mode": "append"}},
-    {"class": "MarkFeatureWriter", "options": {"mode": "skip"}},
+    {
+        "module": "glyphsLib.featureWriters.markFeatureWriter",
+        "class": "ContextualMarkFeatureWriter",
+        "options": {"mode": "skip"},
+    },
+    {"class": "CursFeatureWriter"},
```
anthrotype, line 220: there is also a GdefFeatureWriter in the ufo2ft's list of default feature writers here:

https://github.com/googlefonts/ufo2ft/blob/9dacc83b6ec9de09bcc06c5e7402f1a179e58b7a/Lib/ufo2ft/featureCompiler.py#L169-L174

If you omit it from the list of filter overrides in the UFO lib, then it will not be run for .glyphs sources any more.
I think glyphsLib no longer writes out a GDEF table in the features.fea, but it sets the public.openTypeCategories in the ufo.lib with the intention of having ufo2ft GdefFeatureWriter use those to build the GDEF.

Lib/glyphsLib/builder/constants.py
```diff
@@ -210,14 +210,14 @@
 # mode for the ufo2ft KernFeatureWriter whenever the GSFont contain a non-automatic
 # 'kern' feature.
 # See https://glyphsapp.com/tutorials/contextual-kerning
-# NOTE: Even though we use the default "skip" mode for the MarkFeatureWriter,
-# we still must include it this custom featureWriters list, as this is used
-# instead of the default ufo2ft list of feature writers.
-# This also means that if ufo2ft adds new writers to that default list, we
-# would need to update this accordingly... :-/
 DEFAULT_FEATURE_WRITERS = [
     {"class": "KernFeatureWriter", "options": {"mode": "append"}},
```
anthrotype, line 214: also the mode="append" is no longer needed/used when building from .glyphs sources, because glyphsLib currently adds an insertion maker for ufo2ft to use in its default mode (called "skip" but actually checks for insertion markers). So you need to remove this from the KernFeatureWriter options, let it use the default KernFeatureWriter

anthrotype:
I haven't looked a the logic of this proposed ContextualMarkFeatureWriter because I don't know how this contextual anchors are supposed to work, but I am a bit worried that this might make it more complicated to maintain and debug the MarkFeatureWriter, as we would then have code scattered across multiple repositories. I don't mean to block this, I understand there is a need to implement this, but I wonder, would it be possible to activate this special writer only when the .glyphs source actually contains such contextual anchors, and otherwise use the vanilla markFeatureWriter in ufo2ft? I suspect the majority of projects don't make use of this specialized writer and thus do not need to, so to speak, incur this added "risk" of higher debugging/maintenance cost introduced by the default use of this glyhps-specific writer. Does it make sense?

khaledhosny:
> I haven't looked a the logic of this proposed ContextualMarkFeatureWriter because I don't know how this contextual anchors are supposed to work, but I am a bit worried that this might make it more complicated to maintain and debug the MarkFeatureWriter, as we would then have code scattered across multiple repositories. I don't mean to block this, I understand there is a need to implement this, but I wonder, would it be possible to activate this special writer only when the .glyphs source actually contains such contextual anchors, and otherwise use the vanilla markFeatureWriter in ufo2ft? I suspect the majority of projects don't make use of this specialized writer and thus do not need to, so to speak, incur this added "risk" of higher debugging/maintenance cost introduced by the default use of this glyhps-specific writer. Does it make sense?

I want to extend this mark feature writer to also filter out anchor that start with non-alphanumeric character (https://github.com/googlefonts/ufo2ft/issues/175 and https://github.com/googlefonts/ufo2ft/pull/510), and possibly more in the future since it is not ideal to impose Glyphs-specific anchor naming conventions on UFO files (where anchor naming conventions are lightly specified).

anthrotype (approved):

Issue #646, opened 2020-12-02
_ensureMasterLayers doesn't


simoncozens:
classes.py has this code:

https://github.com/googlefonts/glyphsLib/blob/6c409dd27cb46919c7fe10c5db2b77ce45751632/Lib/glyphsLib/classes.py#L855-L858

I believe the commented code is correct and the uncommented test is wrong.

With current code:

```
>>> import glyphsLib
>>> g = glyphsLib.GSFont("tests/data/MontserratStrippedDown.glyphs")
>>> newGlyph = glyphsLib.GSGlyph()
>>> newGlyph.name = "foobar"
>>> newGlyph.parent = g
>>> g.glyphs.append(newGlyph)
>>> g.glyphs["foobar"].layers._ensureMasterLayers()
>>> print(g.glyphs["foobar"].layers)
()
```

Swapping current code for commented code:

```
>>> import glyphsLib
>>> g = glyphsLib.GSFont("tests/data/MontserratStrippedDown.glyphs")
>>> newGlyph = glyphsLib.GSGlyph()
>>> newGlyph.name = "foobar"
>>> newGlyph.parent = g
>>> g.glyphs.append(newGlyph)
>>> g.glyphs["foobar"].layers._ensureMasterLayers()
>>> print(g.glyphs["foobar"].layers)
(<GSLayer "Light" (foobar)>, <GSLayer "SemiBold" (foobar)>, <GSLayer "Bold" (foobar)>)
```

simoncozens:
However! With my correction, this test fails:

```
    def test_layers_missing_master(self):
        """
        font.glyph['a'] has its layers in a different order
        than the font.masters and an extra layer.
        Adding a master but not adding it as a layer to the glyph should not
        affect glyph.layers unexpectedly.
        """
```

Note that the comment and the call to `_ensureMasterLayers` contradict each other.

anthrotype:
according to git blame Jany wrote that code, so I defer to him

anthrotype:
/cc @yanone who added that `_ensureMasterLayers` in the first place
https://github.com/googlefonts/glyphsLib/commit/5aa5ee7acbe4c3764c2dd551f4b518f6cd345042

belluzj:
My contribution here was to add the two commented lines... I can't remember what I saw then but was probably thinking the same as @simoncozens but didn't go all the way to replace the existing test with my version.

[...]

PR #922, opened 2023-07-01
Remove very slow code that does nothing

khaledhosny:
It is iterating over all masters in the font, then tries to access the master by its id and then checks if the master is None, but this can never happen since we are iterating over already existing masters.

This code is called in a very hot spot, and in a script I’m profiling, it takes more than 50% of the font build time.

Fixes https://github.com/googlefonts/glyphsLib/issues/646


schriftgestalt:
I think the code would make some sense when it would iterate over the masters and checks for layers but as discussed not be issue, that is not really needed.

anthrotype (approved):

PR #928, opened 2023-07-11
generate layer.name from intermediate layer coordinates

anthrotype:
Fixes https://github.com/googlefonts/glyphsLib/issues/851

currently .glyphs v3 sources containing intermediate ('brace') layers fail to build with fontmake because glyphsLib is generating designspace files with multiple sources at the same location.
glyphsLib uses the `layer.name` to define the name of the UFO layer containing the glyphs at intermediate locations, and for .glyphs v2 sources this was ok; but for .glyphs v3 the `layer.name` for intermediate layers is _not_ what the source file contains at the layer's `name` attribute (usually just the date-time when the layer was first created), but is supposed to be a string generated from the layer's coordinates (as displayed in the Glyphs UI).
This PR is an alternative to @kontur's https://github.com/googlefonts/glyphsLib/pull/902 PR, in that it changes the GSLayer.name property getter itself, instead of simply the calling site where the brace/intermediate layers get converted to UFO layers.
I believe this is better because:
1) printing the `layer.name` for a brace layer from within the Glyphs.app's Macro window will return the generated string, not the underlying original name string (e.g. timestamp), so we'd match that;
2) the original layer name can't be edited from the Glyphs.app UI and is not even shown to the user so we can treat it as non-existent and we should not care about roundtripping it back, it's just useless.


khaledhosny:
I see two issues here:
1. this goes against the direction of matching GlyphsApp API more closely (#923)
2. it breaks round tripping glyphs files (opening the file and saving it without changing anything will change many layer names).

I think we need to answer the question of what GlyphsApp API to use for layer names (https://github.com/googlefonts/glyphsLib/issues/851#issuecomment-1555960597) and change the code accordingly.

anthrotype:
this has been sitting for too long and it's blocking the build of several fonts. I don't know how to answer those valid questions you raised, I'm just interested in unblocking the glyphs2ufo (and thus build) for intermediate layers in .glyphs v3 sources
@khaledhosny would you rather do https://github.com/googlefonts/glyphsLib/pull/902 then?

anthrotype:
> I think https://github.com/googlefonts/glyphsLib/pull/902 is better.

alright, I'll incorporate that and finish it up.

I honestly don't feel like adding that nameUI you suggest, not only because it doesn't match what Glyphs.app currently does, but I have no idea how that's supposed to do with other special kinds of layers, here we are only interested in the intermediate (brace) sort.

anthrotype:
> color layers

are those also broken by the layer.name issue in #851?

khaledhosny:
> > color layers
> 
> are those also broken by the layer.name issue in #851?

Propably not, since we rename these layers already. https://github.com/googlefonts/glyphsLib/blob/dee4e92887425a056549f289986a2e167e29a457/Lib/glyphsLib/builder/layers.py#L36-L38

I was thinking if we are going to have a central solution, lets handle more cases than handling them all over the place. But feel free to focus on fixing the issue at hand.

anthrotype:
OK the glyphs->glyphs is not affected by this change, only the glyphs->ufo (and thus ufo->glyphs) but I think it is for good

schriftgestalt:
I think I have dealt with that in the Glyphs3 branch already. Or at least it will handle that the same way as Glyphs does when I’m done.

anthrotype:
Thanks Georg, good to know. However that branch doesn't seem to be in a mergeable state right now, ideally we would like to hot-fix this issue in `main` sooner than later.

schriftgestalt:
I’ll see if I can have a look at it in the next couple of days.

khaledhosny (approved):

PR #932, opened 2023-07-29
Accept `.glyphspackage` in GSFont constructor

khaledhosny:
Remove the file extension check, which should not be our concern, and use `parser.load()` which handles `.glyphspackage` for us.


simoncozens (approved):
LGTM!

PR #939, opened 2023-08-30
Do not set Unicode values for each layer in a color glyph.

### LINKED ISSUE #937: Duplicate codepoint mapping when generating color fonts
The ufo2ft library recently fixed an issue where an [error was accidentally swallowed when multiple glyphs were mapped to the same codepoint](https://github.com/googlefonts/ufo2ft/pull/739).

When this fix made it downstream to fontmake, our Glyphs-based color fonts stopped building with errors:

```
Compiling UFO failed: cannot map 'A.color1' to U+0041; already mapped to 'A'
```

It seems that glyphsLib generates codepoint mappings for each glyph in a layer instead of for the base glyph only. Before I open a PR to fix it, I wanted to checked if the desired result is that the base glyph should have the codepoint mapping and any glyphs in layers should not.

bramstein:
As described in #937, the current color layer implementation writes unicode values for each layer in a color glyph. Recent [fontmake fixes](https://github.com/googlefonts/ufo2ft/pull/739) now treat this as an error which prevents color fonts from building.

This changes `color_layers.py` to set the Unicode mapping on each layer-glyph in the color glyph to an empty list. It does this in `color_layers.py` rather than modifying the `builder.to_ufo_glyph` method because it seems highly specific to color layer-glyphs.

Fixes #937.

### DIFF
--- a/Lib/glyphsLib/builder/color_layers.py
+++ b/Lib/glyphsLib/builder/color_layers.py
@@ -37,6 +37,9 @@ def _to_ufo_color_palette_layers(builder, master, layerMapping):
                 ufo_layer = builder.to_ufo_layer(glyph, masterLayer)
                 ufo_glyph = ufo_layer.newGlyph(layerGlyphName)
                 builder.to_ufo_glyph(ufo_glyph, layer, glyph)
+                # Remove Unicode mapping from each color layer to avoid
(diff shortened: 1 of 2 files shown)


anthrotype (approved):
LGTM, thanks!

Issue #946, opened 2023-09-29
New problems with duplicate Unicode for color glyphs


yanone:
The duplicate Unicode issue still appears for [Blaka Ink](https://github.com/Gue3bara/Blaka/blob/master/sources/Blaka-Ink.glyphs), a COLRv1 font with gradients.

Observations:
* The tests that were [recently added](https://github.com/googlefonts/glyphsLib/pull/939/files) are clearly insufficient in catching this.
* I'm not 100% sure yet, but I believe that the unicodes must be muted in two places: [here](https://github.com/googlefonts/glyphsLib/blob/8c0157e4fba7a97fc9fbd4e4b532bcf92796c06f/Lib/glyphsLib/builder/color_layers.py#L197) in addition to [here](https://github.com/googlefonts/glyphsLib/blob/8c0157e4fba7a97fc9fbd4e4b532bcf92796c06f/Lib/glyphsLib/builder/color_layers.py#L42).
* [This condition](https://github.com/googlefonts/glyphsLib/blob/8c0157e4fba7a97fc9fbd4e4b532bcf92796c06f/Lib/glyphsLib/builder/glyph.py#L272) as triggered from [here](https://github.com/googlefonts/glyphsLib/blob/8c0157e4fba7a97fc9fbd4e4b532bcf92796c06f/Lib/glyphsLib/builder/builders.py#L34) currently prevents the processing of color layers for my font (`minimal=False`).

I can't easily fix this because I don't understand the _minimal_ architecture. But I'm happy to help with a PR if I'm told what to do.

yanone:
I would start by adding

```python
    color0.attributes["color"] = 1
    color1.attributes["color"] = 1
```

to the new test [here](https://github.com/googlefonts/glyphsLib/blob/8c0157e4fba7a97fc9fbd4e4b532bcf92796c06f/tests/builder/builder_test.py#L1200).

It already starts throwing other errors before even checking for the unicodes.

anthrotype:
> _clearly_ insufficient

well, not clearly enough to me by just reading this issue. Could you provide steps to reproduce the issue?

anthrotype:
> I don't understand the minimal architecture. 

fontmake sets minimal=True because it's only interested in the minimal pieces of data required for compilation, not caring about other metadata desirable when roundtripping Glyphs<=>UFO

yanone:
Take Blaka Ink (linked at the top), and try to generate a font from it.

[...]

PR #947, opened 2023-09-29
Ensure no unicodes for color layers

yanone:
Similar to https://github.com/googlefonts/glyphsLib/pull/939 (color palettes unicodes), this PR ensures that color layers are written without unicodes, which otherwise prevents font generation.

Fixes https://github.com/googlefonts/glyphsLib/issues/946


anthrotype (approved):
LGTM, thank you!

PR #955, opened 2023-11-08
Add support for 'Virtual Master' custom parameters

### LINKED ISSUE #859: support for virtual masters
We gonna have more and more fonts with parametric axes. Those axes often affect a part of a glyphset's outlines (ascender, descender, x-heigth, etc). Therefore to be able to do that better while using Glyphs, we are going to need virtual masters which is a feature allowing a better management of design sources.

Doc about virtual masters: https://glyphsapp.com/learn/creating-a-variable-font at section "Optional: add a virtual master".

I made a test case with Rubik. 
[virtual-master.zip](https://github.com/googlefonts/glyphsLib/files/10757339/virtual-master.zip)

The set up is: there is one component that is a line. And it is implemented in each glyphs to create guidelines. 

Capture d’écran 2023-02-16 à 16 10 19

Then I have an axis called "Guideline thickness". This axis only affect that one component. So it would be annoying to duplicate masters in sources for that, we may as well use virtual masters.

Capture d’écran 2023-02-16 à 16 25 06

These intermediate layers are only in the component glyph.

This is the result when exporting with Glyphs:
https://user-images.githubusercontent.com/12222436/219409640-430e33f2-24fc-4dd7-9e7d-1634777de531.mov

Great stuff.

cc @simoncozens

anthrotype:
Fixes #859

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,7 +15,7 @@ jobs:
     - name: Set up Python
       uses: actions/setup-python@v4
       with:
-        python-version: "3.x"
+        python-version: "3.11"
     - name: Install dependencies
       run: pip install tox
     - name: Run style and typing checks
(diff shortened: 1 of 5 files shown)


anthrotype:
I couldn't find real world font projects using virtual master custom parameters, but @MariannaPaszkowska shared with me a couple of test fonts of her and they seem to work with the current PR, so I think this is is good to go.
Do let me know if you find any issues and we can fix later.

Issue #925, opened 2023-07-04
DesignSpace sources contain duplicate locations


clauseggers:
This was using `glyphsLib 6.2.3.dev13+gdee4e928` in `gftools`. Is it related to https://github.com/googlefonts/glyphsLib/issues/851 ? Or perhaps another issue entirely?

```
INFO:GFBuilder:Creating variable fonts from Playfair-2_1-Roman.glyphs
INFO:fontmake.font_project:Building master UFOs and designspace from Glyphs source
INFO:glyphsLib.parser:Parsing .glyphs file
WARNING:glyphsLib.builder.sources:DesignSpace sources contain duplicate locations; varLib expects each master to define a unique location. Make sure that you used consistent 'brace layer' names in all the glyph layers that share the same location.
  ['7 Jun 23 at 10:23', '7 Jun 23 at 10:12', '7 Jun 23 at 11:44', '7 Jun 23 at 11:57', 'Playfair Needlepoint SemiCondensed ', '7 Jun 23 at 11:04', '7 Jun 23 at 12:03', '7 Jun 23 at 10:42'] => {'Optical size': 1200.0, 'Width': 95.0, 'Weight': 600.0}
  ['7 Jun 23 at 10:52', '7 Jun 23 at 11:50'] => {'Optical size': 1200.0, 'Width': 95.0, 'Weight': 640.0}
  ['7 Jun 23 at 10:45', '7 Jun 23 at 11:50'] => {'Optical size': 1200.0, 'Width': 100.0, 'Weight': 600.0}
  ['7 Jun 23 at 10:23', '7 Jun 23 at 10:18', '7 Jun 23 at 11:45', '7 Jun 23 at 11:58'] => {'Optical size': 1200.0, 'Width': 114.0, 'Weight': 500.0}
  ['Playfair Needlepoint SemiCondensed ', '7 Jun 23 at 11:06', '7 Jun 23 at 12:04'] => {'Optical size': 1200.0, 'Width': 114.0, 'Weight': 600.0}
  ['16 Jun 23 at 15:24', '7 Jun 23 at 10:55'] => {'Optical size': 1200.0, 'Width': 114.0, 'Weight': 640.0}
INFO:fontmake.font_project:Building variable fonts gftools_ttf/variable/Playfair-2_1-Roman-VF.ttf
Traceback (most recent call last):
[...]

PR #956, opened 2023-11-17
Allow brace layers with the same coordinates to have different associatedMasterId

belluzj:
Glyphs.app allows brace layers with the same coordinates to have different associatedMasterId. The same situation is not allowed in Designspace + UFO (because you can't have 2 sources with the same location), so the idea for a fix is: for each brace layer location, select one UFO = designspace source that will receive all the brace layers for that location.

Related issue: https://github.com/googlefonts/glyphsLib/issues/925

NOTE: the code is very ugly, I was just exploring ideas of how to fix this. I marked as Draft for now.

Example that works without this patch:
* dollar
  * Regular
    * {401, 100}
  * Bold
* cent
  * Regular
    * {401, 100} -- good because {401, 100} is always associated to Regular
  * Bold

Example that needs this patch to work:
* dollar
  * Regular
    * {401, 100}
  * Bold
* cent
  * Regular
  * Bold
    * {401, 100} -- bad because the brace layers for /dollar and /cent have the same coordinates {401, 100} but are associated to different masters

### DIFF
--- a/Lib/glyphsLib/builder/sources.py
+++ b/Lib/glyphsLib/builder/sources.py
@@ -166,40 +166,49 @@ def _to_designspace_source_layer(self):
     for key, master_ids in layer_to_master_ids.items():
         brace_coordinates = list(key[1])
         layer_name = key[0]
-        for master_id in master_ids:
-            # ... as they may need to be filled up with the values of the associated
-            # master.
-            master = self._sources[master_id]
-            master_coordinates = brace_coordinates
-            if len(master_coordinates) < len(designspace.axes):
-                master_locations = [master.location[a.name] for a in designspace.axes]
-                master_coordinates = (
-                    brace_coordinates + master_locations[len(brace_coordinates) :]
-                )
-            elif len(master_coordinates) > len(designspace.axes):
-                logger.warning(
-                    "Glyph(s) %s, brace layer '%s' defines more locations than "
-                    "there are design axes.",
-                    layer_to_glyph_names[key],
-                    layer_name,
-                )
-
-            # If we have more locations than axes, ignore the extra locations.
-            layer_coordinates_mapping = collections.OrderedDict(
-                (axis.name, location)
-                for axis, location in zip(designspace.axes, master_coordinates)
+        # In a designspace, we can't create the brace layers with the same
+        # location under different UFOs because that would lead to several
+        # <source> elements with the same location. Instead we'll pick the first
+        # (after sorting) master and stick all the brace layers with the same
+        # key = location there.
+        # Redirect other masters who have brace layers at the same location to
+        # the "chosen" one.
+        master_id = sorted(master_ids)[0]
+        for other_master_id in master_ids:
+            self._where_to_put_brace_layers[key, other_master_id] = master_id
+        # ... as they may need to be filled up with the values of the
+        # associated master.
+        master = self._sources[master_id]
+        master_coordinates = brace_coordinates
+        if len(master_coordinates) < len(designspace.axes):
+            master_locations = [master.location[a.name] for a in designspace.axes]
+            master_coordinates = (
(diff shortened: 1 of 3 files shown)


Lib/glyphsLib/builder/sources.py
```diff
@@ -166,40 +166,49 @@ def _to_designspace_source_layer(self):
...
+        # In a designspace, we can't create the brace layers with the same
+        # location under different UFOs because that would lead to several
+        # <source> elements with the same location. Instead we'll pick the first
+        # (after sorting) master and stick all the brace layers with the same
+        # key = location there.
+        # Redirect other masters who have brace layers at the same location to
+        # the "chosen" one.
+        master_id = sorted(master_ids)[0]
+        for other_master_id in master_ids:
+            self._where_to_put_brace_layers[key, other_master_id] = master_id
```
anthrotype, line 178: looks like you are doing this twice (also in builder/builders.py:150-152)? why?

Lib/glyphsLib/builder/sources.py
```diff
@@ -166,40 +166,49 @@ def _to_designspace_source_layer(self):
...
+        # ... as they may need to be filled up with the values of the
+        # associated master.
+        master = self._sources[master_id]
```
anthrotype, line 181: are you sure you're filling up the values from the actually associated master and not from the first one that you picked when multiple ones are available?

anthrotype:
> for each brace layer location, select one UFO = designspace source that will receive all the brace layers for that location.

the proposed solution sounds ok to me, but maybe the PR needs some clean up work.
I'm also tempted to say this can be fixed in the sources relatively easily and I don't see a strong need to support the use case of defining intermediate layers with same location but associated to different masters in different glyphs, I understand that Glyphs.app UI doesn't prevent that and the designer can easily create those unknowingly but I doubt that they want to deliberately do that.

belluzj:
Discussing this with @schriftgestalt , he suggests that the default behaviour of glyphsLib could be cleaner on the UFO side (as there's not a problem anymore of "under which full UFO to put the sparse layer". However this is harder to round-trip because it loses the association to a specific master on the Glyphs.app side

belluzj:
Georg adds that merging his Glyphs3 branch should probably be a reason to bump the major version number, and that could also be an opportunity to change the default output for brace layers from layers to standalone UFOs.

One question with standalone is how to choose the filename of the UFOs, and it could be some combo of axis tags + values joined by underscores.

anthrotype:
I disagree that using standalone UFOs instead of UFO layers referenced from DS source via layer attribute is a solution here.
Standalone UFOs are currently treated like full masters that have their own fontinfo and kerning dictionaries, unlike brace/intermediate layers which should only contribute to variate glyph outlines and metrics. We added DS source.layerName exactly to support glyphs brace layers!

schriftgestalt:
But all projects (that I have seen recently) that are build as designspace/ufo (not in Glyphs) have the sparse masters in separate .ufos (with or without extra font info/kerning). So it would be good to being able to read/write those designspace files without massively changing the file structure. Extra ufos makes it easier to edit/audit them.

And the vertical metrics defined in the full/"parent" source are most likely not going to match the outlines in the brace layers, because those need to be interpolated from the surrounding sources.

anthrotype:
> Extra ufos makes it easier to edit/audit them.

why?

> And the vertical metrics defined in the full/"parent" source are most likely not going to match the outlines in the brace layers

the sparse UFO layers don't have define any global metrics of their own, their parent UFO does, so they don't contribute to the interpolation of vertical metrics, similarly to the Glyphs.app's brace/intermediate layers themselves, which only contribute glyph outlines and glyph advances.

schriftgestalt:
But does it show what glyphs have multiple layers or lets filter for glyphs that have?

PR #961, opened 2023-11-19
Gitignore

schriftgestalt:


.gitignore
```diff
@@ -125,6 +125,9 @@ dmypy.json
 
 # Autosaved files
 *~
+*(Autosaved).*
+*(Automatisch gesichert).*
+*(guardado automáticamente).*
```
anthrotype, line 130: i've never noticed these, what app creates them?

.gitignore
```diff
@@ -14,7 +14,6 @@ dist/
 downloads/
 eggs/
 .eggs/
-lib/
```
anthrotype, line 17: have you actually had an issue with this `lib/` directory being git-ignored on a case-insensitive filesystem?

I think it comes from this example .gitignore file for python projects from Github itself: 
https://github.com/github/gitignore/blob/main/Python.gitignore#L17

/cc @madig who authored the original change

schriftgestalt: There is no "lib" folder, only a "Lib" folder. And we like to keep that. 
A new file inside that Lib folder will not show up in the git listing.

anthrotype: ah ok yes, I see now, thanks
let's merge

anthrotype (approved):

.gitignore
simoncozens: Glyphs!

Issue #963, opened 2021-03-02
Error generating NotoSerifDevanagari


JelleBosmaMT:
Trying to generate fonts from Noto Serif Devanagari I get this error:

`fontmake: Error: In 'NotoSerifDevanagari-MM.glyphs' -> 'master_ufo/NotoSerifDevanagari.designspace' -> 'master_ufo/NotoSerifDevanagari-Light.ufo': Compiling UFO failed: <features>:6089:9: Expected '{'`

Version fontmake 2.3.1 with Python 3.8 on Mac.
Used command: python3 fontmake -g "NotoSerifDevanagari-MM.glyphs"

The error message was not very meaningful to me. If there is is a missing { in the embedded feature code, I do not think Glyphs.app would manage to generate fonts. That made me suspect the GPOS code. I could generate fonts after removing all the anchors. I haven't tried to narrow it down any further.

Attached a ZIP file with the latest Monotype version of the Glyphs file.

[NotoSerifDevanagari-MM.glyphs.zip](https://github.com/googlefonts/fontmake/files/6068300/NotoSerifDevanagari-MM.glyphs.zip)

PR #964, opened 2023-11-22
markFeatureWriter: Skip ignorable anchors

khaledhosny:
Glyphs allows anchors to start with non-letter characters and these are used internally to align components and should not be exported to the feature file.

Fixes https://github.com/googlefonts/glyphsLib/issues/963

### DIFF
--- a/Lib/glyphsLib/featureWriters/markFeatureWriter.py
+++ b/Lib/glyphsLib/featureWriters/markFeatureWriter.py
@@ -24,6 +24,7 @@ class ContextuallyAwareNamedAnchor(NamedAnchor):
         "number",
         "markClass",
         "isContextual",
+        "isIgnorable",
         "libData",
     )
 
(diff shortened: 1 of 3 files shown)


anthrotype (approved):

anthrotype:
@khaledhosny thanks. Could you by any chance try to see if this glyphsLib-specific markFeatureWriter will continue to work with the changes in https://github.com/googlefonts/ufo2ft/pull/796 ?

simoncozens:
`*` is not ignorable, though.

khaledhosny:
> `*` is not ignorable, though.

It is already handled and dropped from the name before the new check is called.

Lib/glyphsLib/featureWriters/markFeatureWriter.py
```diff
@@ -83,13 +84,15 @@ def parseAnchorName(
         else:
             isMark = False
 
-        return isMark, key, number, isContextual
+        isIgnorable = not key[0].isalpha()
```
khaledhosny, line 87: By this point `key` is the anchor name without `*` or `_`.

PR #970, opened 2024-01-09
features: Do not include empty labels

m4rc1e:
In Glyphsapp, users can define names for Stylistic Sets.

Screenshot 2024-01-09 at 11 23 56

Currently, if a user doesn't add a name, GlyphsLib will export the stylistic set with an empty name.

This PR won't add empty names which mimics the behaviour of GlyphsApp's exporter.

I discovered this issue while checking https://github.com/google/fonts/pull/7104. Fontbakery is reporting that the font is missing nameIDs 256, 257 etc which stem from this issue, https://github.com/fonttools/fontbakery/issues/3671#issuecomment-1882830352

### DIFF
--- a/Lib/glyphsLib/builder/features.py
+++ b/Lib/glyphsLib/builder/features.py
@@ -78,7 +78,7 @@ def _is_manual_kern_feature(feature):
     return feature.name == "kern" and not feature.automatic
 
 
-def _to_ufo_features(
+def _to_ufo_features(  # noqa: C901
     font: GSFont,
     ufo: Font | None = None,
     generate_GDEF: bool = False,
@@ -143,9 +143,9 @@ def _to_ufo_features(
             for label in feature.labels:
                 langID = _to_name_langID(label["language"])
                 name = label["value"]
-                name = name.replace("\\", r"\005c").replace('"', r"\0022")
                 if name == "":
                     continue
+                name = name.replace("\\", r"\005c").replace('"', r"\0022")
                 if langID is None:
                     feature_names.append(f'  name "{name}";')
                 else:
--- a/Lib/glyphsLib/builder/features.py
+++ b/Lib/glyphsLib/builder/features.py
@@ -140,16 +140,19 @@ def _to_ufo_features(
                     feature_names = ["featureNames {", f'  name "{name}";', "};"]
         elif font.format_version == 3 and feature.labels:
             feature_names = []
-            feature_names.append("featureNames {")
             for label in feature.labels:
                 langID = _to_name_langID(label["language"])
                 name = label["value"]
                 name = name.replace("\\", r"\005c").replace('"', r"\0022")
+                if name == "":
+                    continue
                 if langID is None:
                     feature_names.append(f'  name "{name}";')
                 else:
                     feature_names.append(f'  name 3 1 0x{langID:X} "{name}";')
-            feature_names.append("};")
+            if feature_names:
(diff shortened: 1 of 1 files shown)


anthrotype:
but why are there feature labels with empty string value? how are those created to begin with? Can the user just clean up their source files to make sure these empty labels are removed?

Lib/glyphsLib/builder/features.py
```diff
@@ -138,7 +138,11 @@ def _to_ufo_features(
                     # Replace special chars backslash and doublequote for AFDKO syntax
                     name = name.replace("\\", r"\005c").replace('"', r"\0022")
                     feature_names = ["featureNames {", f'  name "{name}";', "};"]
-        elif font.format_version == 3 and feature.labels:
+        elif (
+            font.format_version == 3
```
anthrotype, line 142: do we need to check the format_version? if there are any feature.labels we just use them..

Lib/glyphsLib/builder/features.py
```diff
@@ -138,7 +138,11 @@ def _to_ufo_features(
...
+            and feature.labels
+            and all(l["value"] != "" for l in feature.labels)
```
anthrotype, line 144: maybe you meant `any(l["value"] != "" for l in feature.labels)`?
I think we want to skip the whole `featureNames` section if all the labels are empty (or add one if any is non empty)

Also below in the `for label in feature.labels:` loop, I think you want to do `continue` if the name == ""

m4rc1e:
> but why are there feature labels with empty string value? how are those created to begin with?

When the user adds a stylistic set feature, you get the UI in the above pic which includes an empty text box. 

>  Can the user just clean up their source files to make sure these empty labels are removed?

It doesn't look like it. If you want a stylistic set, the text box is there and you cannot remove it.

Lib/glyphsLib/builder/features.py
m4rc1e: I've decided to drop `any` or `all`, far easier to just include them if they get appended to the `feature_names` list.

Lib/glyphsLib/builder/features.py
```diff
@@ -140,16 +140,19 @@ def _to_ufo_features(
                     feature_names = ["featureNames {", f'  name "{name}";', "};"]
         elif font.format_version == 3 and feature.labels:
             feature_names = []
-            feature_names.append("featureNames {")
             for label in feature.labels:
                 langID = _to_name_langID(label["language"])
                 name = label["value"]
                 name = name.replace("\\", r"\005c").replace('"', r"\0022")
+                if name == "":
```
anthrotype, line 147: How about you move this one line up? The replacements don't affect the emptiness of the string and can be skipped

anthrotype (approved):

Lib/glyphsLib/builder/features.py
m4rc1e: 🤣 your wish is my command

schriftgestalt:
it would be good to have a test case for this. That can ensure that the Glyphs3 branch will not regress on this.

emmamarichal:
I tested it on Pacifico:
[Test_stylisticsets_nameID.zip](https://github.com/googlefonts/glyphsLib/files/14138629/Test_stylisticsets_nameID.zip)

The fail disappeared, with only have `⚠ WARN: Ensure Stylistic Sets have description.` now.
Thanks @m4rc1e!

Issue #954, opened 2023-11-08
Issue with components from intermediate layers


anthrotype:
If a font contains a mixed glyph (with outlines and components) that should get decomposed at build time, and any of the components in turn reference a glyph that has intermediate ("brace") layers, currently the mixed glyph has to have the same set of intermediate layers as its components' base glyphs, otherwise the latter are ignored as the mixed glyph is decomposed and thus the originally mixed now decomposed glyph will not display as expected (i.e. with the expected intermediate layers) in the generated font.
Glyphs.app itself does not impose this, and the mixed-composite glyph needs not replicate the intermediate layers of the component glyphs.

The following test file, [IntermediateLayer.glyphs.zip](https://github.com/googlefonts/glyphsLib/files/13297181/IntermediateLayer.glyphs.zip), contains glyph "i" which is composed of "idotless" and an outline for the i's dot. The "idotless" glyph has an additional intermediate layer (drawn as a triangle to make it clear if the intermediate is working or not when rendered). 
If the "{700}" intermediate layer is removed from the glyph "i", the font compiled with fontmake (as well as fontc) will render the glyph "i" incorrectly at wght=700, the stem of the i rendered as a rectangle instead of a triangle (like in idotless at wght=700). 

<img width="600" alt="Screenshot 2023-11-08 at 13 35 01" src="https://github.com/googlefonts/glyphsLib/assets/6939968/be93fd60-0a03-4833-9295-f397b96e154c">

<img width="600" alt="Screenshot 2023-11-08 at 13 35 06" src="https://github.com/googlefonts/glyphsLib/assets/6939968/2fe055aa-bdb4-445e-be43-f9c2199f8d17">

anthrotype:
I noticed this issue while working on #859 where @RosaWagner's test font similarly uses a `_part.line` component in a bunch of ABCDE etc. glyphs which also contain regular outlines thus should be decomposed. The workaround there was to add all the intermediate layers to the composite glyphs (manually one by one..)

schriftgestalt:
Adding the extra intermediate layer to the composites what needed in Glyphs 2 but in Glyphs 3 it is not any more.

anthrotype:
Could you explain how this is supposed to work?
For each glyph that has a mixture of outlines and components, for each of the components, if the component's base glyph has any intermediate layers that the parent glyph has not, then a new intermediate layer is added to the parent glyph containing decomposed outlines interpolated at that intermediate layer's location?

[...]

PR #971, opened 2024-01-10
Build intermediate layers with non-intermediate components

simoncozens:
This is kind of the inverse of #954. If an ~alternate~ intermediate layer references a component, and that component does not have a matching intermediate layer, glyphsLib currently generates a "too sparse" master. In the test file in this PR, `Aslash` has an intermediate layer but it uses a component glyph `A` which does not have that intermediate layer. Sometimes the resulting UFO file for the sparse master can't be compiled, because the referenced component `A` is missing; sometimes it can be compiled but there is no variation:

```
INFO:ufo2ft.filters:Running DecomposeComponentsFilter on ComponentsWithIntermediates-Regular-{151}
WARNING:ufo2ft.util:dropping non-existent component 'A' in glyph 'Astroke'
...
WARNING:fontTools.varLib:glyph Astroke has incompatible masters; skipping
```

There are two ways this could be addressed - we could interpolate and decompose the component, or we could make the sparse master a bit less sparse by interpolating the referenced layer. I've taken the second approach because (a) it's simpler, and (b) keeping components where possible can lead to smaller file sizes.

This also introduces a new way of preparing Glyphs files for Glyphs->UFO translation; we now have a `preflight_glyphs` function which runs a list of translations, progressively "dumbifying" all the smart stuff in the `GSFont` object. Currently this is only used for the current job (de-sparsifying overly sparse masters), but in the future, we could use this to simplify some of the more gnarly translations we do in compilation, especially those requiring Glyphs-specific ufo2ft plugins. (i.e. EraseOpenCorners, CornerComponents, etc.)

### DIFF
--- a/Lib/glyphsLib/builder/builders.py
+++ b/Lib/glyphsLib/builder/builders.py
@@ -28,6 +28,7 @@ from .constants import (
     FONT_CUSTOM_PARAM_PREFIX,
 )
 from .axes import WEIGHT_AXIS_DEF, WIDTH_AXIS_DEF, find_base_style, class_to_value
+from .transformations import TRANSFORMATIONS
 from glyphsLib.util import LoggerMixin
 
 
@@ -86,7 +87,7 @@ class UFOBuilder(LoggerMixin):
                    production, and unnecessary steps will be skipped.
         glyph_data -- A list of GlyphData.
         """
-        self.font = font
+        self.font = preflight_glyphs(font)
 
         if ufo_module is None:
             import ufoLib2 as ufo_module
@@ -409,6 +410,18 @@ class UFOBuilder(LoggerMixin):
         to_ufo_layer_user_data,
         to_ufo_node_user_data,
     )
+
+
+def preflight_glyphs(font):
+    # Run a set of transformations over a GSFont object to make
+    # it easier to convert to UFO; resolve all the "smart stuff".
+
+    # We could "font = copy.deepcopy(font)" here, since we'll be
+    # modifying the original. But that's a pain and we won't do it
+    # until someone complains.
+    for transform in TRANSFORMATIONS:
+        transform(font)
+    return font
 
 
 def filter_instances_by_family(instances, family_name=None):
--- a/Lib/glyphsLib/builder/builders.py
+++ b/Lib/glyphsLib/builder/builders.py
@@ -28,7 +28,6 @@ from .constants import (
     FONT_CUSTOM_PARAM_PREFIX,
 )
 from .axes import WEIGHT_AXIS_DEF, WIDTH_AXIS_DEF, find_base_style, class_to_value
-from .transformations import TRANSFORMATIONS
 from glyphsLib.util import LoggerMixin
 
 
@@ -55,7 +54,9 @@ class UFOBuilder(LoggerMixin):
         """Create a builder that goes from Glyphs to UFO + designspace.
 
         Keyword arguments:
-        font -- The GSFont object to transform into UFOs
+        font -- The GSFont object to transform into UFOs. We expect this GSFont
+                object to have been pre-processed with
+                ``glyphsLib.builder.preflight_glyphs``.
         ufo_module -- A Python module to use to build UFO objects (you can pass
                       a custom module that has the same classes as ufoLib2 or
                       defcon to get instances of your own classes). Default: ufoLib2
@@ -87,7 +88,7 @@ class UFOBuilder(LoggerMixin):
                    production, and unnecessary steps will be skipped.
         glyph_data -- A list of GlyphData.
         """
-        self.font = preflight_glyphs(font)
+        self.font = font
 
         if ufo_module is None:
             import ufoLib2 as ufo_module
@@ -410,18 +411,6 @@ class UFOBuilder(LoggerMixin):
         to_ufo_layer_user_data,
         to_ufo_node_user_data,
     )
-
-
-def preflight_glyphs(font):
-    # Run a set of transformations over a GSFont object to make
-    # it easier to convert to UFO; resolve all the "smart stuff".
-
-    # We could "font = copy.deepcopy(font)" here, since we'll be
-    # modifying the original. But that's a pain and we won't do it
-    # until someone complains.
-    for transform in TRANSFORMATIONS:
-        transform(font)
-    return font
 
 
 def filter_instances_by_family(instances, family_name=None):
--- a/Lib/glyphsLib/builder/transformations/intermediate_layers.py
+++ b/Lib/glyphsLib/builder/transformations/intermediate_layers.py
@@ -4,6 +4,7 @@ import uuid
 from fontTools.varLib.models import VariationModel, normalizeValue
 
 from glyphsLib.classes import GSLayer, GSNode, GSPath
+from glyphsLib.builder.axes import get_regular_master
 
 
 logger = logging.getLogger(__name__)
@@ -30,11 +31,14 @@ def simple_variation_model(font):
     locations = [x.axes for x in font.masters]
     limits = {tag: (min(x), max(x)) for tag, x in zip(tags, (zip(*locations)))}
     master_locations = []
+    default_location = get_regular_master(font).axes
     for loc in locations:
         this_loc = {}
         for ix, axisTag in enumerate(tags):
             axismin, axismax = limits[axisTag]
-            this_loc[axisTag] = normalizeValue(loc[ix], (axismin, axismin, axismax))
+            this_loc[axisTag] = normalizeValue(
+                loc[ix], (axismin, default_location[ix], axismax)
+            )
         master_locations.append(this_loc)
     return VariationModel(master_locations, axisOrder=tags), limits
 
--- a/Lib/glyphsLib/builder/transformations/intermediate_layers.py
+++ b/Lib/glyphsLib/builder/transformations/intermediate_layers.py
@@ -24,28 +24,6 @@ def resolve_intermediate_components(font):
                 components_with_intermediate_layers.add(glyph.name)
                 all_intermediate_locations.add(tuple(layer.attributes["coordinates"]))
 
-    # It's not clear to me that this is needed... (See #954)
-
-    # for glyph in font.glyphs:
-    #     # Just peek at first layer, assume compatibility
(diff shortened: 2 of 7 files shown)


Lib/glyphsLib/builder/builders.py
```diff
@@ -411,6 +412,18 @@ def instance_data(self):
     )
 
 
+def preflight_glyphs(font):
+    # Run a set of transformations over a GSFont object to make
+    # it easier to convert to UFO; resolve all the "smart stuff".
+
+    # We could "font = copy.deepcopy(font)" here, since we'll be
+    # modifying the original. But that's a pain and we won't do it
+    # until someone complains.
```
anthrotype, line 421: I wonder if we were modifying the input GSFont previously, before this PR; I'd expect we should not, though I can see how this might be convenient for you here..
How about you document that this preflight works in place (so if one cares, they can pass in a copy) and instead of doing it automatically inside the UFOBuilder constructor, you run it immediately before the latter is run?
Feels like the two should be kept distinct from one another instead of conflated like this.

anthrotype:
> If an alternate layer references a component, and that component does not have a matching intermediate layer

just a terminology clarification: IIRC, an "alternate layer" in Glyphs.app has a specific meaning akin to so-called 'bracket' layers (that are swapped within given axis ranges) which is distinct from "intermediate layer" proper (aka 'brace' layers), best not to use the two interchangeably

Lib/glyphsLib/builder/transformations/intermediate_layers.py
```diff
@@ -0,0 +1,113 @@
...
+    for loc in locations:
+        this_loc = {}
+        for ix, axisTag in enumerate(tags):
+            axismin, axismax = limits[axisTag]
+            this_loc[axisTag] = normalizeValue(loc[ix], (axismin, axismin, axismax))
+        master_locations.append(this_loc)
+    return VariationModel(master_locations, axisOrder=tags), limits
+
+
+def check_component_has_sparse_layer(font, component, parent_layer):
```
anthrotype, line 42: it's not "check" but more like "ensure", as this is not just warning about something is missing, but modifying the thing to add what's expected

Lib/glyphsLib/builder/transformations/intermediate_layers.py
```diff
@@ -0,0 +1,113 @@
...
+def simple_variation_model(font):
+    tags = [axis.axisTag for axis in font.axes]
+    locations = [x.axes for x in font.masters]
+    limits = {tag: (min(x), max(x)) for tag, x in zip(tags, (zip(*locations)))}
+    master_locations = []
+    for loc in locations:
+        this_loc = {}
+        for ix, axisTag in enumerate(tags):
+            axismin, axismax = limits[axisTag]
+            this_loc[axisTag] = normalizeValue(loc[ix], (axismin, axismin, axismax))
```
anthrotype, line 37: I'm not sure this "simple_variation_model" is correct in the general: you're discarding the actual _default_ master axis coordinates (which can be set by the user via 'Variable Font Origin' and glyphsLib should know since it uses it when creating the DS axes) and instead assuming that default == min all the time; the choice of a default is not neutral, it may have consequences throughout the designspace because in the variation model that's always the first thing that doesn't move and is used as reference point for computing the deltas for the rest of the masters thus will influence how the VF behaves everywhere.

Lib/glyphsLib/builder/builders.py
anthrotype: I undestand, but it still feels strange to me that we now suddenly choose to modify the input GSFont in place just because it's handy, we can't be 100% sure that one (ourselves even in the unit tests) aren't calling the existing to_ufos or to_designspace or UFOBuilder methods on the _same_ GSFont object with different parameters maybe or after having changed something in it (with the assumption that these methods weren't actually supposed to have side effects but just return something new each time and keep the input unchanged).

Lib/glyphsLib/builder/transformations/intermediate_layers.py
simoncozens: Right, but I don't care about OT deltas here. I'm *only* using this model for interpolating the outlines between masters. I don't need it to be correct in the general case, which is why it's "simple".

simoncozens: You're right, and I was only dragging my feet because I thought it was going to be a pain to fix, but it turned out not to be.

Lib/glyphsLib/builder/transformations/intermediate_layers.py
```diff
@@ -0,0 +1,118 @@
...
+    return VariationModel(master_locations, axisOrder=tags), limits
+
+
+def ensure_component_has_sparse_layer(font, component, parent_layer):
+    tags = [axis.axisTag for axis in font.axes]
+    model, limits = simple_variation_model(font)
+    location = parent_layer.attributes["coordinates"]
+    normalized_location = {
+        axisTag: normalizeValue(
+            location[ix], (limits[axisTag][0], limits[axisTag][0], limits[axisTag][1])
```
anthrotype, line 52: since you're now using the real default and not assume min==default, I think you should adapt this part as well so that 0.0 default actually means the same thing as in the VarModel

Lib/glyphsLib/builder/transformations/intermediate_layers.py
```diff
@@ -0,0 +1,118 @@
...
+            this_loc[axisTag] = normalizeValue(
+                loc[ix], (axismin, default_location[ix], axismax)
+            )
+        master_locations.append(this_loc)
+    return VariationModel(master_locations, axisOrder=tags), limits
+
+
+def ensure_component_has_sparse_layer(font, component, parent_layer):
+    tags = [axis.axisTag for axis in font.axes]
+    model, limits = simple_variation_model(font)
```
anthrotype, line 48: the simple model is only taking into account the global master locations.. should it not also consider the locations of the are defined for intermediate layers to be able to interpolate the ones that are missing?

Lib/glyphsLib/builder/transformations/intermediate_layers.py
```diff
@@ -0,0 +1,118 @@
...
+from glyphsLib.classes import GSLayer, GSNode, GSPath
+from glyphsLib.builder.axes import get_regular_master
+
+
+logger = logging.getLogger(__name__)
+
+
+def resolve_intermediate_components(font):
+    components_with_intermediate_layers = set()
+    all_intermediate_locations = set()
```
anthrotype, line 15: you don't seem to be currently using these two sets. In the comment below you say "Later we _will_ check if...", do you mean these sets that you're populating will be useful in the future when we actually write that logic (like a TODO or self reminder)?

Lib/glyphsLib/builder/transformations/intermediate_layers.py
anthrotype: IIUC, the problem here is handling a situation whereby there's an intermediate layer that contains a component, and the component's glyph does not have a corresponding intermediate layer explicitly defined at the same location. But this is not necessarily all or nothing, i.e. the component glyph either defining the same set of intermediate layers as the parent layer, or having no intermediate layer at all (only proper master layers). It is possible that the component glyph's layers contain _some_ intermediate while missing others; and when you go to interpolate the latter, should you not also take into account the intermediate layers that _are_ present? I believe so, however right now you only use the master layers from the font.

Lib/glyphsLib/builder/transformations/intermediate_layers.py
simoncozens: I wrote some code to try to fix #954, but because I couldn't replicate that issue I removed it again. Or most of it, at least. These will be leftovers.

Lib/glyphsLib/builder/transformations/intermediate_layers.py
anthrotype: > What we're checking is it if the component defines the same alternate layer as the thing using the component.

yes, that's what I meant with "the component's glyph does not have a corresponding intermediate layer explicitly defined at the same location [as the intermediate layer that contains that component]", slightly more verbosely

> you're right, we should use any intermediate layers that it does have in our interpolation model.

👍

anthrotype:
@simoncozens is this ready to merge? or you plan to do more work to tackle https://github.com/googlefonts/glyphsLib/issues/954 in here?

anthrotype:
IIUC this and the other https://github.com/googlefonts/glyphsLib/issues/954 are basically two faces of the same problem, we want to make sure that composite glyphs and their components all have the same set of intermediate layers; in here, you fixed the issue when the composite has intermediates that the component has not; in #954, the component has more intermediate layers than the composite glyph that uses it, so we need to add the missing ones there. It's probably even possible (though unlikely) that the composite and the component have _different_ intermediate layers for whatever reason, which means you need to add the respectively missing ones to both.

anthrotype:
let's merge this as is for now so I can make a new release

anthrotype:
>  The composite glyph doesn't have any intermediate layers, which I thought this PR addresses.

anthrotype:
>  The composite glyph doesn't have any intermediate layers, which I thought this PR addresses.

no, this PR only addresses the issue when a composite glyph defines more intermediate layers than the component glyph that it references. You are hitting https://github.com/googlefonts/glyphsLib/issues/954 -- glyph used as component has more intermediate layers than the composite glyph that uses it

yanone:
I had a feeling I got it mixed up. Thank you, I'll try out the other PR.

simoncozens:
I think I screwed up something here:

```
DesignSpace sources contain duplicate locations; varLib expects each master to define a unique location. Make sure that you used consistent 'brace layer' names in all the glyph layers that share the same location.
  ['{151, 100}', '{151.0, 100.0}'] => {'Weight': 151.0, 'Width': 100.0}
```

anthrotype:
I presume this started happening since v6.6.2 (which includes this PR)? 
Also is this a Glyphs 3 souce? if so the error message is misleading because it talks about "consistent 'brace layer' names" which only makes sense for Glyphs 2.

anthrotype:
> How can this layer._brace_coordinates() == location condition ever be true if the first operand is of type generator expression and the second is of type tuple?

I made it return a list so that comparison should work now (https://github.com/googlefonts/glyphsLib/commit/a944fd3afac8b6e4de6ef5699c89c104874b981f)

Issue #973, opened 2024-01-29
Problems with color components


simoncozens:
This glyphs file: [BadColor.glyphs.gz](https://github.com/googlefonts/glyphsLib/files/14082212/BadColor.glyphs.gz) has two masters, each with two color layers. I would expect these layers to be exported as ".color0" and ".color1". However, the names of some (not all!) component glyphs used in these masters are incorrect:

```
$ glyphs2ufo BadColor.glyphs
$ cat BadColor-Thin.ufo/glyphs.color.1/A_acute.glif
<?xml version='1.0' encoding='UTF-8'?>
<glyph name="Aacute" format="2">
  <advance width="500"/>
  <unicode hex="00C1"/>
  <anchor x="685" y="10" name="ogonek"/>
  <anchor x="355" y="708" name="top"/>
  <outline>
    <component base="A.color1"/>
    <component base="acutecomb.color2" xOffset="370" yOffset="208"/>
  </outline>
```

There is no color2, so the font does not compile.

anthrotype:
the issue must be somewhere here:

https://github.com/googlefonts/glyphsLib/blob/ed02363c5e2d53294fc7a49992df2da5006f5f2e/Lib/glyphsLib/builder/components.py#L50-L69

simoncozens:
Yep, that looks suspicious. I will try to fix.

anthrotype:
thanks. I tried to understand what was going on but failed so far, happy to pass it to you

PR #974, opened 2024-01-29
Fix color component layers

simoncozens:
This is a tiny fix for #973. In `components.py` we gather a list of color layers and take the index to get the new component name, like so:

https://github.com/googlefonts/glyphsLib/blob/ed02363c5e2d53294fc7a49992df2da5006f5f2e/Lib/glyphsLib/builder/components.py#L59-L68

We do this because we do the same operation in `_to_ufo_color_palette_layers` to rename the glyphs:

https://github.com/googlefonts/glyphsLib/blob/ed02363c5e2d53294fc7a49992df2da5006f5f2e/Lib/glyphsLib/builder/color_layers.py#L184-L194

However, the list which we take the index from *there* is actually built here:

https://github.com/googlefonts/glyphsLib/blob/ed02363c5e2d53294fc7a49992df2da5006f5f2e/Lib/glyphsLib/builder/glyph.py#L257-L262

So in `components.py` we need to make sure we are building the same list by also testing for `l.associatedMasterId == layer.associatedMasterId`.

### DIFF
--- a/Lib/glyphsLib/builder/components.py
+++ b/Lib/glyphsLib/builder/components.py
@@ -58,6 +58,7 @@ def to_ufo_components(self, ufo_glyph, layer):
             component_glyph = font.glyphs[component_name]
             color_layers = [
                 l for l in component_glyph.layers if l._is_color_palette_layer()
+                and l.associatedMasterId == layer.associatedMasterId
             ]
             for i, l in enumerate(color_layers):
                 if l._color_palette_index() == layer._color_palette_index():


anthrotype (approved):
AH! thanks!!

anthrotype:
would be nice if we could also add that test file of yours to the test suite. Maybe just parsing it and converting to_designspace and asserting that the component names are as expected with the correct `.color{i}` suffix, perhaps inside tests/builder/components_test.py. Could you please add that?

PR #975, opened 2024-01-30
[anchor_propagation] fix issue adjusting mark component anchor

anthrotype:
In anchor_propagation.py, we have code (added by @yanone in https://github.com/googlefonts/glyphsLib/pull/617) to adjust propagated anchor position when a mark component is anchored to a specific named anchor as specified in the ComponentInfo lib key (this can be used to specify which of the seveal ligature anchors the given component is supposed to be attached to e.g. top_1, top_2 etc). 

The code indiscriminately applies the adjustment inside a for loop, without checking the anchor names match the expected component anchor.

I modified one of the AnchorPropagation.glyphs test file by changing the order of anchors in a mark glyph used as component in order to trigger a failure (anchor order is itself meaningless, it only happened to work before because the last anchor was the right one).
I will follow up with a fix

### DIFF
--- a/Lib/glyphsLib/builder/anchor_propagation.py
+++ b/Lib/glyphsLib/builder/anchor_propagation.py
@@ -110,10 +110,17 @@ def _adjust_anchors(anchor_data, ufo, parent, component):
     t = Transform(*component.transformation)
     _componentAnchor = _componentAnchorFromLib(parent, component)
     for anchor in glyph.anchors:
+        if anchor.name.startswith("_"):
+            # only base anchors should get propagated
+            continue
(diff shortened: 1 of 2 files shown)


anthrotype:
I'm not sure about the regression tests failing, if it's a good thing or not in this particular case.

anthrotype:
I've inspected the regresson test and I've convinced myself that these were bugs and this fixes them

anthrotype:
@simoncozens does this change look sane to you? I'm inclined to merge

anthrotype:
I think this is correct so I'm gonna merge, but feel free to file any issues that may arise

PR #977, opened 2024-01-31
Export variable instances as DSv5 variable-fonts

khaledhosny:
Export Glyphs 3 as DesignSpace v5 variable-fonts, setting their custom parameters is the variable-font lib, mainly using the "public.fontInfo" key.

### DIFF
--- a/Lib/glyphsLib/builder/instances.py
+++ b/Lib/glyphsLib/builder/instances.py
@@ -56,13 +56,41 @@ logger = logging.getLogger(__name__)
 def to_designspace_instances(self):
     """Write instance data from self.font to self.designspace."""
     for instance in self.font.instances:
-        if instance.type == InstanceType.VARIABLE:
-            continue
         if self.minimize_glyphs_diffs or (
             is_instance_active(instance)
             and _is_instance_included_in_family(self, instance)
         ):
-            _to_designspace_instance(self, instance)
+            if instance.type == InstanceType.VARIABLE:
+                _to_designspace_varfont(self, instance)
+            else:
+                _to_designspace_instance(self, instance)
+
+
+def _to_designspace_varfont(self, instance):
+    from fontTools.designspaceLib import RangeAxisSubsetDescriptor
+    from fontTools.ufoLib import fontInfoAttributesVersion3
+
+    ds = self.designspace
+    ufo_varfont = self.designspace.addVariableFontDescriptor(
+        name=instance.name,
+        filename=instance.customParameters["fileName"],
+        axisSubsets=[RangeAxisSubsetDescriptor(name=axis.name) for axis in ds.axes],
+    )
+
+    # to_ufo_custom_params() wants a UFO, create a dummy one
+    ufo = self.ufo_module.Font()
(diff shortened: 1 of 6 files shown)


Lib/glyphsLib/builder/instances.py
```diff
@@ -56,13 +56,41 @@
...
+        axisSubsets=[RangeAxisSubsetDescriptor(name=axis.name) for axis in ds.axes],
+    )
+
+    # to_ufo_custom_params() wants a UFO, create a dummy one
+    ufo = self.ufo_module.Font()
+    to_ufo_custom_params(self, ufo, instance, set_default_params=False)
+
+    info = {}
+    for attr in fontInfoAttributesVersion3:
+        if value := getattr(ufo.info, attr, None):
```
anthrotype, line 86: wouldn't this `if value := getattr(ufo.info, attr, None)` branch not trigger if `value == 0`?

Lib/glyphsLib/builder/instances.py
```diff
@@ -56,13 +56,41 @@
...
+
+def _to_designspace_varfont(self, instance):
+    from fontTools.designspaceLib import RangeAxisSubsetDescriptor
+    from fontTools.ufoLib import fontInfoAttributesVersion3
+
+    ds = self.designspace
+    ufo_varfont = self.designspace.addVariableFontDescriptor(
+        name=instance.name,
+        filename=instance.customParameters["fileName"],
+        axisSubsets=[RangeAxisSubsetDescriptor(name=axis.name) for axis in ds.axes],
```
anthrotype, line 77: this feels redundant, I wish one coudl leave axisSubsets empty to mean "this VF uses all the defined axes", but I suppose that's not how the DS spec wants this.

Lib/glyphsLib/builder/instances.py
```diff
@@ -56,13 +56,41 @@
...
+    info = {}
+    for attr in fontInfoAttributesVersion3:
+        if (value := getattr(ufo.info, attr, None)) is not None:
+            info[attr] = value
+
+    if info:
+        ufo_varfont.lib["public.fontInfo"] = info
+
+    for key in ufo.lib:
+        ufo_varfont.lib[key] = ufo.lib[key]
```
anthrotype, line 93: I wonder, what could end up here? Any params for which there isn't a corresponding fontinfo key basically? I guess it's fine to store them here, even though they won't be used right know by anybody.

anthrotype (approved):
Thanks Khaled!

Lib/glyphsLib/builder/instances.py
khaledhosny: Yes. They might be useful for round-tripping, though this PR does not handle it right now.

anthrotype:
This does not round-trip back to .glyphs, if I understand correctly, right? How hard do you reckon will that be? Maybe we could file an issue as a TODO for later.

khaledhosny:
I don’t use ufo2glyphs myself, but I might look into it later. I’ll file an issue for now.

Issue #906, opened 2023-06-01
Propagate anchors in glyphs2ufo


schriftgestalt:
It is very difficult to not propagate anchors. As far as I can see, the --propagate-anchors command line option doesn't make a difference as it is on by default anyway.

(some comments from Cosimo:
https://github.com/googlefonts/glyphsLib/commit/a86ff9363f8ec059d08a84840f04556da9a07710#commitcomment-115893027 )
I’ll add an `--no-propagate-anchor` option.

But is it really needed to be on by default for the CLI tools? I understand it is useful to have for the fontmake pipeline, but when converting the file manually, I would aim for a cleaner/roundtripable conversion? And we have the CML argument to enable it already.

And isn’t the anchor decomposition needed for a purely ufo based workflow, too? So the decomposition could be moved into fontmake/ufo2ft?

PR #979, opened 2024-02-07
[glyphs2ufo] add --no-propagate-anchors CLI option

anthrotype:
Fixes https://github.com/googlefonts/glyphsLib/issues/906

Currently glyhps2ufo does the anchor propagation by default, unless the .glyphs file contains "Propagate Anchors" custom parameter set to false. There is a `--propagate-anchors` option but it doesn't do much since we do it even when the custom parameter is missing.

This PR adds the mutually exclusive `--no-propagate-anchors` which will disable this without having to modify the .glyphs file to add the custom parameter.

### DIFF
--- a/Lib/glyphsLib/cli.py
+++ b/Lib/glyphsLib/cli.py
@@ -85,15 +85,25 @@ def main(args=None):
             "roundtripping between Glyphs and UFOs."
         ),
     )
-    group.add_argument(
+    anchor_group = group.add_mutually_exclusive_group()
+    anchor_group.add_argument(
         "--propagate-anchors",
         default=None,
         action="store_true",
         help=(
-            "Copy anchors from underlying components to actual "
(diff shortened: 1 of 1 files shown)


anthrotype:
I also consider changing the default so that glyphs2ufo script does not do the anchor propagation unless requested (via either custom param explicitly set to true or via CLI option) but that would change the current behavior and might get unnoticed producing other issues down the road, so I simply added the negative option without changing the default behavior.

schriftgestalt:
I remember some mention of a cli option to control if the ufo is intended to be used to build binary files or to work on the source (and potentially round trip back). That would make the "minimal" option obsolete, too. 

That option would trigger a lot preprocessing, decomposing of anchors, corner components, built kern/mark features …

So by default when you run glyphs2ufo, it would keep the data as roundtripable as possible. And if you run `fontmake -g …`, it would always preprocessing everything.

PR #980, opened 2024-02-12
Convert MATH plugin data into more usable form in UFO

khaledhosny:
The Glyphs MATH plugin stores its data in master, glyph, and layer userData.  Master and layer userData convert seamlessly to UFO font and layer lib. Glyph userData, however, have no equivalent in UFO so we store them in the font’s lib under a per-glyph key prefixed by glyphsLib prefix, but this is a bit ugly. So this change groups them under two common keys using the same prefix as the other keys.

So instead of:

    <key>com.schriftgestaltung.Glyphs.glyphUserData.braceright</key>
    <dict>
            <key>com.nagwa.MATHPlugin.extendedShape</key>
            <integer>1</integer>
            <key>com.nagwa.MATHPlugin.variants</key>
            <dict>
                    <key>vVariants</key>
                    <array>
                    ...
                    </array>
            </dict>
    </dict>
    <key>com.schriftgestaltung.Glyphs.glyphUserData.braceleft</key>
    <dict>
            <key>com.nagwa.MATHPlugin.extendedShape</key>
            <integer>1</integer>
            <key>com.nagwa.MATHPlugin.variants</key>
            <dict>
                    <key>vVariants</key>
                    <array>
                    ...
                    </array>
            </dict>
    </dict>

We now save:

    <key>com.nagwa.MATHPlugin.extendedShape</key>
    <array>
            <string>parenleft</string>
            <string>parenright</string>
            ....
    </array>
    <key>com.nagwa.MATHPlugin.variants</key>
    <dict>
            <key>parenleft</key>
            <dict>
                 <key>vVariants</key>
                 <array>
                 ...
                 <array>
            </dict>
            <key>parenleft</key>
            <dict>
                 <key>vVariants</key>
                 <array>
                 ...
                 <array>
            </dict>
            ...
    </dict>


Lib/glyphsLib/builder/user_data.py
```diff
@@ -59,7 +61,18 @@ def to_ufo_master_user_data(self, ufo, master):
 def to_ufo_glyph_user_data(self, ufo, glyph):
     key = GLYPH_USER_DATA_KEY + "." + glyph.name
     if glyph.userData:
-        ufo.lib[key] = dict(glyph.userData)
+        # Convert MATH userData to top-level keys and group them under the same
+        # key so that they are in a more usable/compact form.
+        if GLYPHS_MATH_EXTENDED_SHAPE_KEY in glyph.userData:
+            ufo.lib.setdefault(GLYPHS_MATH_EXTENDED_SHAPE_KEY, []).append(glyph.name)
+            del glyph.userData[GLYPHS_MATH_EXTENDED_SHAPE_KEY]
```
anthrotype, line 68: why do you need to modify the Glyphs glyph object's userData when converting this to UFO lib?

anthrotype (commented):
makes sense to group per-glyph userdata under same font lib key. I wonder if we could make this more general instead of ad-hoc for the MATH plugin. I imagine we might want to give similar treatment to other stuff that may live in userDatas

Lib/glyphsLib/builder/user_data.py
khaledhosny: So that it does not get written again by the code below. I’ll figure a way that does not modify the input font.

khaledhosny:
> makes sense to group per-glyph userdata under same font lib key. I wonder if we could make this more general instead of ad-hoc for the MATH plugin. I imagine we might want to give similar treatment to other stuff that may live in userDatas

Makes sense. It might not allow for efficient representation of boolean data, e.g. `isExtendedShape` is saved as a simple list of glyph names because I know it is a boolean, but in general I don’t think Glyphs file format distinguishes between boolean and integers, so if this is done generically it will be a dictionary.

khaledhosny:
> > makes sense to group per-glyph userdata under same font lib key. I wonder if we could make this more general instead of ad-hoc for the MATH plugin. I imagine we might want to give similar treatment to other stuff that may live in userDatas
> 
> Makes sense. It might not allow for efficient representation of boolean data, e.g. `isExtendedShape` is saved as a simple list of glyph names because I know it is a boolean, but in general I don’t think Glyphs file format distinguishes between boolean and integers, so if this is done generically it will be a dictionary.

Thinking more about it, I think also handling this generically will be hard to round-trip. Right now I know that certain font lib keys are actually glyph-level userData, but I have no way of knowing this in the generic case.

My use case here is to get the data to ufo2ft (https://github.com/googlefonts/ufo2ft/pull/819) and I’m trying to avoid making ufo2ft side too GlyphsApp-specific, hence trying to make the data here usable for a UFO-only workflow as well.

Lib/glyphsLib/builder/user_data.py
```diff
@@ -124,8 +144,9 @@ def to_glyphs_family_user_data_from_ufo(self, ufo):
 def to_glyphs_master_user_data(self, ufo, master):
     """Set the GSFontMaster userData from the UFO master-specific lib data."""
     target_user_data = master.userData
+    special_math_keys = {GLYPHS_MATH_VARIANTS_KEY, GLYPHS_MATH_EXTENDED_SHAPE_KEY}
     for key, value in ufo.lib.items():
-        if _user_data_has_no_special_meaning(key):
+        if _user_data_has_no_special_meaning(key) and key not in special_math_keys:
```
anthrotype, line 149: maybe you should add the math keys to `_user_data_has_no_special_meaning`?

Lib/glyphsLib/builder/user_data.py
```diff
@@ -57,9 +60,24 @@ def to_ufo_master_user_data(self, ufo, master):
...
+        # Convert MATH userData to top-level keys and group them under the same
+        # key so that they are in a more usable/compact form.
+        if GLYPHS_MATH_EXTENDED_SHAPE_KEY in mathData:
+            ufo.lib.setdefault(GLYPHS_MATH_EXTENDED_SHAPE_KEY, []).append(glyph.name)
+        if GLYPHS_MATH_VARIANTS_KEY in mathData:
+            ufo.lib.setdefault(GLYPHS_MATH_VARIANTS_KEY, {})[glyph.name] = dict(
+                mathData[GLYPHS_MATH_VARIANTS_KEY]
+            )
+    if self.minimal:
+        return
```
anthrotype, line 76: i'm ok if you want to export the math data regardless of the `minimal` flag, but why exclude the rest of the userData if self.minimal = False? What does this change have to do with the conversion of MATH data?

Lib/glyphsLib/builder/user_data.py
```diff
@@ -85,6 +103,8 @@ def to_ufo_layer_lib(self, master, ufo, ufo_layer):
 def to_ufo_layer_user_data(self, ufo_glyph, layer):
     user_data = layer.userData
     for key in user_data.keys():
+        if self.minimal and not key.startswith(GLYPHS_MATH_PREFIX):
+            continue
```
anthrotype, line 107: same here, I don't understand why you skip exporting layer.userData if self.minimal is False

Lib/glyphsLib/builder/user_data.py
khaledhosny: Previously, this function was not called at all if in mininmal mode (see the change in [`builder/glyph.py`](https://github.com/googlefonts/glyphsLib/pull/980/commits/2195c7046b32caaa7974ef5481cf0e41e5ebb317#diff-d082d50d6170f5a514b1c741f44911a11531f1d9badf10fc91064eda9ecc9e72)), so this is keeping the existing behavior for other userData keys.

anthrotype: oh I see, didn't notice that. ok

Lib/glyphsLib/builder/user_data.py
khaledhosny: `_user_data_has_no_special_meaning` is used in multiple places and in some of them we want to skip MATH key and in others we don’t.

anthrotype (approved):

Issue #981, opened 2024-02-12
DSv5 variable-font's "Name" used as VF filename but only contains the style name


anthrotype:
Since #977, glyphsLib creates `<variable-font>` elements in a DesignSpace v5 document, for each Variable Font Settings (special "instance" of type variable). It sets the `name` attribute to whatever the instance.name was. This normally only contains the _style_ name of an exported (static or variable) font and is used by Glyphs.app for naming the exported fonts by concatenating to the font's family name, and also appending a "VF" suffix at the end of the file name (see https://docu.glyphsapp.com/#GSInstance.name).

fontmake uses the the `<variable-font>` name attribute as is for naming the output path (unless the `"filename"` attribute is also set which takes priority, this corresponds to the `fileName` custom parameter in Glyphs.app, which we do support already).
Thus fontmake won't prepend the family name to it, but will treat it as a full descriptive name of that variable font and simply add the file extension: e.g. you get things like "Regular.ttf" for a Variable Font Setting where the default "Regular" name that it gets when first created has not been modified.

Previously when these Variable Font Settings were ignored and not exported into the designspace, fontmake would make up the VF filename by using the designspace's filename (which glyphsLib sets to the font family name) plus the file extension suffix: so for the test font in glyphsLib/tests/data/VariableInstance.glyph, fontmake would name the output path of the VF as "Cairo-VF.ttf" which is arguably more specific and correct than "Regular.ttf" or "Variable.ttf".

I think here it is glyphsLib which has to adapt to the existing conventions whereby a DSv5 variable-font name contains not just the style name but a full name, also stripped of whitespaces in most examples that I have seen e.g. in the designspaceLib test suite; so instead of merely passing on the Variable Font Setting's Name field and set it as the variable-font's name attribute as is, it should concatenate it to font's family name and strip spaces: e.g. in the test example above, instead of `<variable-font name="Variable"/>` it will set it to `<variable-font name="CairoVariable"/>`, and then fontmake will use it for the output filename "CairoVariable.ttf".

khaledhosny:
May be we can use here whatever algorithm GlyphsApp uses to construct VF file name and set the `fileName` to it if it is not set, and leave the `<variable-font>` name as it is now.

PR #982, opened 2024-02-12
compose DSv5 variable-font filename like Glyphs.app would name exported VFs

anthrotype:
Fixes https://github.com/googlefonts/glyphsLib/issues/981

### DIFF
--- a/Lib/glyphsLib/builder/instances.py
+++ b/Lib/glyphsLib/builder/instances.py
@@ -71,9 +71,21 @@ def _to_designspace_varfont(self, instance):
     from fontTools.ufoLib import fontInfoAttributesVersion3
 
     ds = self.designspace
+
+    # unless the `fileName` custom parameter was explicitly set, do like Glyphs.app
+    # and concatenate the family, instance (style) names and "VF" to form a filename;
+    # the default 'Regular' is omitted by Glyphs.app.
+    # https://github.com/googlefonts/glyphsLib/issues/981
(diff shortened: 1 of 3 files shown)


Lib/glyphsLib/builder/instances.py
```diff
@@ -71,9 +71,21 @@ def _to_designspace_varfont(self, instance):
...
+
+    # unless the `fileName` custom parameter was explicitly set, do like Glyphs.app
+    # and concatenate the family, instance (style) names and "VF" to form a filename;
+    # the default 'Regular' is omitted by Glyphs.app.
+    # https://github.com/googlefonts/glyphsLib/issues/981
+    filename = instance.customParameters["fileName"]
+    if not filename:
+        filename = self.font.familyName
+        if instance.name != "Regular":
+            filename += "-" + instance.name
```
khaledhosny, line 83: Does GlyphsApp use instance name at all here?

khaledhosny (approved):

Lib/glyphsLib/builder/instances.py
anthrotype: yes, it does, just tried again to confirm. It concatenates to the font familyName using a hyphen.

anthrotype: Oh! I also just noticed that in addition to that, Glyphs.app also strips all the spaces from the generated filename. I guess we should do the same.

khaledhosny (approved):

PR #985, opened 2024-02-12
No width linking of brace layers

schriftgestalt:
it was getting it form the master they are attached too and that is supposed to have a different width

### DIFF
--- a/tests/special_layer_width_test.py
+++ b/tests/special_layer_width_test.py
@@ -81,7 +81,7 @@ def test_intermediate_layer_width_with_metrics_source_on_parent():
         if intermediate_round.getFullDesignLocation(doc) == {"Weight": 500, "ROUND": 1}:
             break
     else:
-        assert False, "Can't find intermediate layer in the desigspace"
+        raise AssertionError("Can't find intermediate layer in the desigspace")
     assert (
         intermediate_round.font.layers[intermediate_round.layerName]["a"].width == 200
     )


anthrotype:
it's not clear to me what this is doing (or no longer doing), can you please clarify?

schriftgestalt:
This is mostly a hot fix for a specific project. Brace layer almost always have a different width than the master layers they are attached too. Linking to that width will mess things up. There might be some extra layer that might benefit from linking, so this needs more testing. Until then, this is a draft.

schriftgestalt:
That is correct.

belluzj:
@anthrotype @simoncozens This is ready for review. I added a test, and I checked locally that the test was failing before the patch (the glyph had width == 100) and is green after (width == 200 as specified in the file).

Lib/glyphsLib/builder/glyph.py
```diff
@@ -250,22 +250,20 @@ def effective_width(layer, glyph):
...
-            width = metric_layer.width
-            if layer.width != width:
-                logger.debug(
-                    f"{layer.parent.name}: Applying width from master "
-                    f"'{metrics_source.id}': {layer.width} -> {width}"
-                )
-        else:
-            width = None
-    return width
+    master = font.masters[layer.layerId]
```
belluzj, line 253: This line is the important change of the patch, it says to only consider the master of this layer (if any) for the metrics source, and never the "associated master". For intermediate layers it means that there's no metrics source, whereas previously they used the metrics source of their parent layer, which was wrong.

PR #986, opened 2024-02-17
Fix: "DeprecationWarning: open_binary is deprecated."

khaledhosny:


khaledhosny:
This requires Python 3.9, though, so lets keep open until we drop 3.8 support.

anthrotype:
the importlib_resources backport should work for python < 3.9 if you really want to fix this now, you'd have to add it to the install_requires but only for python < 3.9

Issue #681, opened 2021-04-13
Handle Unicode Variation Sequences


moyogo:
Since [GlyphsApp 2.6.1](https://glyphsapp.com/news/glyphs-2-6-1-released) Glyphs files can have some UVS support:
> Preliminary support for Unicode Variation Sequences: add a .uvXXX suffix to the alternate glyph, with XXX ranging from 001 to 256.

While this is incomplete, as it doesn't support default UVS or duplicates glyphs when it doesn't need to (see https://forum.glyphsapp.com/t/unicode-variation-sequences/11813/7), glyphs2ufo should probably do something with that data and fill in or update `public.unicodeVariationSequences` in the font.lib.

schriftgestalt:
Thanks for reminding me about the missing default UV mappings. I just fixed this. This is fully handled on export so glyphsLib can implement this properly already (using the manually set mappings from the .glyphs file and the default mappings from the Unicode data).

PR #988, opened 2024-02-19
Handle Unicode Variation Sequences

khaledhosny:
If a glyph name ends with ".uvNNN" with NNN ranging from 001 to 256, then it is a variation sequence with uv001 being U+0xFE00 and uv256 being U+0xE01EF.

The only documentation for this is the "More Improvements" section in Glyphs 2.6.1 announcement:

  https://glyphsapp.com/news/glyphs-2-6-1-released

And this forum post:

  https://forum.glyphsapp.com/t/unicode-variation-selector-u-fe00/21701

Fixes https://github.com/googlefonts/glyphsLib/issues/681

### DIFF
--- a/tests/builder/builder_test.py
+++ b/tests/builder/builder_test.py
@@ -2389,6 +2389,22 @@ def test_load_kerning_bracket(ufo_module):
     assert ds2.sources[3].font.kerning == {}
 
 
+def test_unicode_variation_sequences(ufo_module):
+    font = generate_minimal_font()
+    add_glyph(font, "zero")["unicode"] = f"{ord('0'):04x}"
+    add_glyph(font, "zero.uv001")
+    add_glyph(font, "zero.uv255")
+    add_glyph(font, "u1F170")["unicode"] = "1F170"
+    add_glyph(font, "u1F170.uv015")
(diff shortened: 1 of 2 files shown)


tests/builder/builder_test.py
```diff
@@ -2389,6 +2389,22 @@ def test_load_kerning_bracket(ufo_module):
     assert ds2.sources[3].font.kerning == {}
 
 
+def test_unicode_variation_sequences(ufo_module):
+    font = generate_minimal_font()
+    add_glyph(font, "zero")["unicode"] = f"{ord('0'):04x}"
+    add_glyph(font, "zero.uv001")
+    add_glyph(font, "zero.uv255")
```
anthrotype, line 2396: Could we use some standard Unicode variation sequences instead of made up ones for this test?

anthrotype (approved):
Lgtm

tests/builder/builder_test.py
khaledhosny: They are all stabdard variation sequences or IVD sequences, except the `zero.uv255` one, which was
supposed to test the last UVS since I couldn’t find any standard one that use
it. There was a typo in the code and the test expectation was wring as well (yay
for TDD), so I’ve added another test and fixed the typo.

PR #989, opened 2024-02-27
[glyphdata] Custom data should fallback to default for missing glyphs

khaledhosny:
When using custom GlyphData, we should fallback to default GlyphData for missing glyphs. This matches what GlyphsApp does, and allow overriding only a subset of the glyphs in the font.


Lib/glyphsLib/glyphdata.py
```diff
@@ -100,21 +100,25 @@ def get_glyph(glyph_name, data=None, unicodes=None):
...
 
-        assert len(data.names) > 30000
+    if data is None:
+        data = GLYPHDATA
 
     # Look up data by full glyph name first.
     attributes = _lookup_attributes(glyph_name, data)
 
+    # If we are using custom GlyphData, fallback to default GlyphData
+    if attributes == {} and data != GLYPHDATA:
```
anthrotype, line 119: should you use `is not` instead of `!=` here?

anthrotype (approved):

PR #1000, opened 2024-04-19
[glyphdata] Add test to hit _split_glyph_name

cmyr:
This function has very particular behaviour that was not covered by existing tests. (I'm porting this to Rust, and found this logic tricky; might as well backport the test cases)


khaledhosny (approved):

khaledhosny:
Thanks!

Issue #368, opened 2018-05-31
issues with composites' anchor propagation


anthrotype:
while working on the ufo2ft MarkFeatureWriter (I'm adding support for abvm and blwm), I noticed a few issues with glyphsLib's current implementation of the propagate anchor feature in Glyphs.app, whereby composite glyphs inherit anchors from their components.

I started the discussion here https://gitter.im/fonttools-dev/glyphsLib?at=5b0eef9793dc78791c9a5a64

I'll try to summarize my comments again here.

The way glyphsLib's anchor propagation currently works is more or less the following: anchors are automatically propagated from the component's base glyphs to the composite glyphs that reference them; if the target composite glyph already has an anchor with the same name, the anchor from the component is not propagated. In addition, if the composite references more than one components (either the same component more than once, or different components having anchors with same name), then the anchors with clashing names are duplexed with a `_1`, `_2`, suffix.
The latter is a convention used in Glyphs.app to denote anchors on ligature components (the index is the ligature component number, starting from 1). They are translated in `pos ligature ... ligComponent ...;` or mark2liga (LookupType 5) lookups.

One problem I noticed is that, neither glyphsLib nor Glyphs.app (from what I can tell) handle the case when in a ligature glyph, one of the ligature components has no marks attached.
The Feature File spec defines a special `<anchor NULL>` for this purpose:

> If there are no anchor points on a component, it must still specify at least one anchor, which should be the NULL anchor.

https://www.adobe.com/devnet/opentype/afdko/topic_feature_file_syntax.html#6.d

It's not clear what should happen when there is a gap in the sequence of `_1`, `_2`, etc. numbers, if one or more ligature components contains no anchors. How is one supposed to denote such a NULL anchor (i.e. absence of anchors in a ligature component) within this anchor naming scheme? FWIW, Glyphs.app completely ignores such anchors on ligatures unless they start from `_1`, and stops as soon as there is a gap. :thinking: 

I also noticed a crucial difference between our anchor propagation algorithm and Glyphs.app's. We do the propagation to any composite glyph, and do the numbered suffix trick whenever there is any name collision in the propagated anchors.
Glyphs.app, on the other hand, does _not_ seem to do any automatic anchor propagation when the composite glyph is marked as `subCategory` "Ligature" (either in the built-in GlyphData.xml database, or in the CMD-ALT-I panel for overriding these properties on a glyph-by-glyph basis). Marking the glyph to something else like 'Letter" re-enables the anchor propagation.
 
[...]

PR #1007, opened 2024-06-03
[anchor_propagation] the last component's anchor wins for non-liga composites (e.g. DZ)

anthrotype:
https://github.com/googlefonts/glyphsLib/issues/368#issuecomment-2103376997

> For non-ligature glyphs, you should get the anchors from the last component they show up. I just tried the DZ in Glyphs. The D and Z have a "top" anchor so you get the one from the Z. The D has a center anchor (for the bar to make a "Eth"). As the Z doesn’t have a center have it, you get the one from the D.


anthrotype:
superseded by #1011

PR #1010, opened 2024-07-25
markFeatureWriter: Optimize contextual anchor lookups

khaledhosny:
Instead of making one `pos` statement per individual base glyph (which translates to one subtable, at least in feaLib), group all base glyphs in one statement. When there is a large number of base glyphs with the same contextual anchor, this can save quite a bit (for a font I’m working on, this save ~4kb for an anchor with 93 base glyphs).

### DIFF
--- a/Lib/glyphsLib/featureWriters/markFeatureWriter.py
+++ b/Lib/glyphsLib/featureWriters/markFeatureWriter.py
@@ -230,18 +230,27 @@ class ContextualMarkFeatureWriter(MarkFeatureWriter):
                 lookup.statements.append(MarkToBasePos(glyph, [anchor]).asAST())
             lookups.append(lookup)
 
+            # Insert mark glyph names after base glyph names if not specified otherwise.
+            if "&" not in after:
+                after = after.replace("*", "* &")
+
+            # Group base glyphs by anchor
+            glyphs = {}
             for glyph, anchor in glyph_anchor_pair:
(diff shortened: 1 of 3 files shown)


Lib/glyphsLib/featureWriters/markFeatureWriter.py
```diff
@@ -230,18 +230,27 @@ def _makeFeatures(self):
...
                 marks = ast.GlyphClass(
                     self.context.markClasses[anchor.key].glyphs.keys()
                 ).asFea()
-                if "&" not in after:
-                    after = after.replace("*", "* &")
-                # Replace & with mark name if present
-                contextual = after.replace("*", f"{glyph}")
+
+                # Replace * with base glyph names
+                contextual = after.replace("*", f"[{bases}]")
```
anthrotype, line 249: should we avoid the enclosing `[]` if `bases` happens to contain only one glyph maybe? or it doesn't actually make any difference in the built output?

anthrotype (approved):
LGTM with one comment/question

Lib/glyphsLib/featureWriters/markFeatureWriter.py
khaledhosny: It does not seem to make a difference, we are getting the same lookup type (8, chained contexts positioning) with and without this PR.

PR #1011, opened 2024-07-30
Port fontc' propagate_anchors.rs to .py; apply it as preflight transformation

anthrotype:
Fixes https://github.com/googlefonts/glyphsLib/issues/368

This PR translates the fontc' [`propagate_anchors.rs`](https://github.com/googlefonts/fontc/blob/main/glyphs-reader/src/propagate_anchors.rs) from Rust to Python and applies the transformation to the GSFont in the 'preflight' step before it gets passed on to the UFOBuilder.
The Rust code was based off the original Glyphs.app's Objective-C code (Georg shared a snippet with us privately) and is as such more 'correct' then the old, reverse-engineered implementation.
The previous `glyphsLib.builder.anchor_propagation` module is now deprecated but is kept around in case users may instantiate the UFOBuilder class directly with propagate_anchors=True.
Also, the existing `propagate_anchors` parameters of `to_designspace`, `to_ufos` and other high-level methods continues to have an equivalent effect but now controls the **new** propagate anchors logic. Clients such as fontmake do not need to do anything besides upgrading glyphsLib to use this.

### DIFF
--- a/Lib/glyphsLib/builder/__init__.py
+++ b/Lib/glyphsLib/builder/__init__.py
@@ -151,7 +151,7 @@ def to_designspace(
     return builder.designspace
 
 
-def preflight_glyphs(font, *, glyph_data=None, **todos):
+def preflight_glyphs(font, *, glyph_data=None, **flags):
     """Run a set of transformations over a GSFont object to make
     it easier to convert to UFO; resolve all the "smart stuff".
 
@@ -169,7 +169,7 @@ def preflight_glyphs(font, *, glyph_data=None, **todos):
         font: a GSFont object
         glyph_data: an optional GlyphData object associating various properties to
             glyph names (e.g. category) that overrides the default one
-        **todos: a set of boolean flags to enable/disable specific transformations,
+        **flags: a set of boolean flags to enable/disable specific transformations,
             named `do_<transformation_name>`, e.g. `do_propagate_all_anchors=False`
             will disable the propagation of anchors.
 
@@ -178,7 +178,7 @@ def preflight_glyphs(font, *, glyph_data=None, **todos):
     """
 
     for transform in TRANSFORMATIONS:
-        do_transform = todos.pop("do_" + transform.__name__, None)
+        do_transform = flags.pop("do_" + transform.__name__, None)
         if do_transform is True:
             pass
         elif do_transform is False:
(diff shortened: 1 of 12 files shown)


anthrotype:
I skimmed the regression tests failures (too many to actually vet each single one), and they all seem as expected and for the good.
The only remaining known difference between this Python (and Rust) implementation, on the one hand, and the built-in Glyphs.app's implementation of anchor propagation, on the other hand, is that the former only handles anchor propagation for "master" layers (i.e. those present by definition which have the same layerId == associatedMasterId in all glyphs), whereas the latter also handles it for non-master (intermediate, alternate, color, etc.) layers.
This limitation is due to the fact that glyphsLib does not currently implement the `componentLayer` property, which would allow to dereference components even when the base glyph does not contain a layer with corresponding layerId (existing non-master layers may get matched by their attributes or new layers get interpolated on the fly): https://github.com/googlefonts/glyphsLib/pull/853#issuecomment-1402257549

However, the same limitation already exists for the current (soon-to-be legacy) anchor_propagation.py, as this is similarly performed for each glyph in each master UFO (`for glyph in ufo:` etc.), that is only on the default layer of such ufo, and not considering additional ufo layers where Glyphs intermediate (brace) layers are stored by glyphsLib.

So I don't consider this limitation blocking. Implementing `componentLayer` (with all the required on-the-fly interpolation of  missing intermediate layers) is complicated and would prefer not to block on that.

Lib/glyphsLib/builder/__init__.py
```diff
@@ -148,12 +151,49 @@ def to_designspace(
...
+    Some transformations may have custom parameters that can be set in the
+    font. For example, the `propagate_all_anchors` transformation can be
+    disabled by setting the custom parameter "Propagate Anchors" to False
+    (see `TRANSFORMATION_CUSTOM_PARAMS`).
+
+    Args:
+        font: a GSFont object
+        glyph_data: an optional GlyphData object associating various properties to
+            glyph names (e.g. category) that overrides the default one
+        **todos: a set of boolean flags to enable/disable specific transformations,
```
rsheeter, line 172: I'm a bit confused about the name; shouldn't they be called something like `flags`?

rsheeter (approved):
I won't claim deep understanding of anchor prop but I don't see any obvious problems and the tests seem to agree, lgtm

Issue #1017, opened 2024-08-08
propagate_anchors should resolve component references in special layers (intermediate, alternate, etc.)


anthrotype:
In both the old anchor_propagation.py and the new one ported from fontc (#1011), anchor propagation is only performed for 'master' layers. To get the anchors from a given component, we search the base glyph for a layer with the same `layerId`, however this only works for master layers. If we can't find a layer in the referenced base glyph that has the same `layerId` as the layer were a component is defined, we skip that component and move on.
Therefore, composite glyphs inside non-master, special layers such as intermediate (brace) or alternate (bracket) layers, or the various color layers, do not currently inherit any anchors from their component glyphs.
Ideally we should match the built-in Glyphs.app implementation and propagate the anchors for special layers as well.

In order to do that, we need to fully implement the logic for resolving component references, which is encapsulated in the `GSComponent.componentLayer` property from Glyphs.app Python scripting API.

As Georg explained in https://github.com/googlefonts/glyphsLib/pull/853, a componentLayer is not the parent layer where a component is defined, but the layer which the component is pointing to in the base glyph, either one of the existing layers or a new one interpolated as needed.

The logic to resolve component references goes as follows (currently only point 1. is implemented):

1) First, search the component's base glyph for a layer with the same `layerId`; if we find one, this must be a 'master' layer for only the latter kind have the same `layerId` in all the glyphs - non-master layers have a unique `layerId`, not useful for matching across glyphs.

2) Else, look for a layer with same `associatedMasterId` _and_ a matching `layerKey`. Wait - what's a layer key now?! It's what the private `GSLayer.layerKey()` method returns in the Glyphs.app python interface: a _string_ that concatenates a layer's optional attributes such as the intermediate coordinates, alternate layer's axis range, whether it's a color layer or "color palette" or bitmap strike or SVG color layer. It may look like `{600}` or `Regular [550‹wght‹700]` or `Color 1` or a combination of these. Also note, the layerKey is always `None` for master layers (they don't need one) and for non-master/non-special (attribute-less) layers like the newly created ones used as draft or backup (they aren't compiled into the font anyway).
[...]

PR #1018, opened 2024-08-09
Implement GSLayer.layerKey() method

anthrotype:
Needed to resolve component references as part of fixing #1017


schriftgestalt:
I implemented this based on the code from Glyphs, yesterday. Can compare with your code as soon as I’m back at my Mac in a bit.

anthrotype:
> I implemented this based on the code from Glyphs, yesterday.

even better

>  Can compare with your code as soon as I’m back at my Mac in a bit.

thanks

PR #1025, opened 2024-08-28
Fix the order of feature writers

khaledhosny:
We want curs feature lookups to come first, then kerning, then mark/mkmk, but we currently are calling CursFeatureWriter last which causes its lookups to be also written last. The lookup order does not matter for HarfBuzz, but it matters for other implementations, see: https://github.com/harfbuzz/harfbuzz/issues/4596#issuecomment-2220558649


khaledhosny:
The regression tests failures are all because we now write the feature writers in a different order in the UFO lib, so they will fix themselves once this PR is merged.

khaledhosny:
Merging this as it matches googlefonts/ufo2ft#864

PR #1029, opened 2024-09-04
Allow GSFont object to be passed to `build_masters` directly (in place of a path)

Hoolean:
This helps avoid a costly and redundant save-and-load for Glyphs sources that are manipulated in memory prior to conversion.

### DIFF
--- a/Lib/glyphsLib/__init__.py
+++ b/Lib/glyphsLib/__init__.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from __future__ import annotations
 
 from io import open
 import collections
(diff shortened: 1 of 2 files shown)


anthrotype:
i'm not necessarily opposed to this, however note that this build_masters function will continue to interact with the filesystem (it writes out the DS and the master UFOs to disk). If you don't want to do that, wouldn't it be better for your code to actually call the inner to_designspace function which this build_masters uses?

anthrotype:
i suppose this is still useful if one wants do to do what build_masters does but also wants to do some extra pre-processing on the GSFont before it is turned into DS+UFO, so LGTM

anthrotype (approved):

Hoolean:
yes, that last comment is our exact use-case 🚀

Hoolean:
Ta for speedy review Cosimo :) I don't have merge rights here, so I'm afraid I'll need to ask a maintainer to hit the final merge button for me too

PR #1030, opened 2024-09-04
markFeatureWriter: Use new ufo2ft MarkFeatureWriter

khaledhosny:
The code here was merged back to ufo2ft’s MarkFeatureWriter. Keeping ContextualMarkFeatureWriter as stub just in case someone is using it directly. Keeping the tests too, to exercise exporting anchor libData to UFO.


khaledhosny:
Should be merged after the new ufo2ft release and requiring it as minimum supported version.

PR #1032, opened 2024-09-28
Rename florin to fhook

moyogo:
Change caterogy=Letter, case=lower as in Unicode.

Keep production name as florin for AGL.


moyogo:
The `test_glyphs_to_designspace[AndadaPro.glyphs]` regression test fails because of fhook.sc having the production name florin.sc in the generated UFOs with this PR’s change, when it had none previously.

AndadaPro.glyphs has a messy setup with regard to Fhook/fhook.
Fhook is as expected, florin is a letter (upright, ascender to descender height), fhook.sc is a smallcap Fhook, florin.sc is a currency symbol (italic, short).
The c2sc feature has the incorrect `sub Fhook by florin.sc` when it should be `sub Fhook by fhook.sc`.
The smcp feature does not have any substitution `sub florin by fhook.sc`.

Glyphs.app 3.3 (3322) doesn’t generate the adequate substitutions automatically if upper and lower case are not named Fhook and fhook.

When florin is renamed fhook, the adequate `sub Fhook by fhook.sc` and `sub fhook by fhook.sc` are added to c2sc and smcp respectively.

/cc @schriftgestalt

anthrotype:
shouldn't any changes to the data xml files first be applied to the upstream repository and then sync'ed back in here? (if that's already the case, ignore me and just merge)

anthrotype:
oh I now see you also filed https://github.com/schriftgestalt/GlyphsInfo/pull/41. let's merge once that one is also merged

khaledhosny:
Included in #1063

PR #1033, opened 2024-09-28
Rename 019B and A7DC

moyogo:
Add '-latin': 019B lambdastroke-latin and A7DC Lambdastroke-latin


khaledhosny:
Shouldn’t this be done upstream first?

khaledhosny:
Included in #1063

Issue #905, opened 2023-05-31
custom parameter is used when present even if disabled


anthrotype:
The .glyphs fonts at this link https://github.com/clauseggers/Playfair/tree/master/sources contain "Axis Mappings" custom parameters whose checkboxews are disabled in the Glyphs UI. The plist for the custom parameter contains a `disabled = 1` attribute. The sources also contain "Axis Location" parameters, so the intention is that the latter be used instead of the disabled Axis Mappings.
When both methods are present, glyphsLib prefers to use the "Axis Mappings" as it's the most explicit. However, if this is marked as disabled glyphsLib should treat it as if it is not set at all.
Currently it is still being read and applied, which may lead to unexpected effects if it is not set the same way as the Axis Location params.

However 
Original issue reported by @clauseggers: https://github.com/googlefonts/fontmake/issues/1002

anthrotype:
this may be a larger issue because, while GSCustomParameter class does have a `disabled` attribute, it doesn't look like we do anything with it... If I search for "disabled" in glyphsLib/builders/custom_params.py or anywhere else really, I don't see it being used anywhere, so it's not just "Axis Mappings". 
Most places where we look up custom parameters we get the value associated with the parameter's name (done in the `ListDictionaryProxy.__getitem__` when the key is not an `int` or a `slice`, so presumably a `str`), but we do not check whether that parameter is disabled. To read that `disabled` attribute you need to have the GSCustomParameter instance, not just its value...

anthrotype:
I could fix this immediate issue by simply doing:

```diff
diff --git a/Lib/glyphsLib/builder/axes.py b/Lib/glyphsLib/builder/axes.py
index cb3b7f2d..af33d7c1 100644
--- a/Lib/glyphsLib/builder/axes.py
+++ b/Lib/glyphsLib/builder/axes.py
@@ -158,7 +158,13 @@ def to_designspace_axes(self):
     regular_master = get_regular_master(self.font)
     assert isinstance(regular_master, classes.GSFontMaster)

-    custom_mapping = self.font.customParameters["Axis Mappings"]
+    # only use 'Axis Mappings' if actually enabled
+    # https://github.com/googlefonts/glyphsLib/issues/905
+    custom_mapping = None
+    for param in self.font.customParameters:
+        if param.name == "Axis Mappings" and not param.disabled:
+            custom_mapping = param.value
+            break

     for axis_def in get_axis_definitions(self.font):
[...]

PR #1036, opened 2024-10-03
skip disabled custom parameters when glyphs=>ufo and minimal=True

anthrotype:
Fixes https://github.com/googlefonts/glyphsLib/issues/905


anthrotype:
(I can add some tests after lunch)

PR #1037, opened 2024-10-05
propagate_anchors: fix handling of contextual anchors

khaledhosny:
- ~~Don’t propagate contextual anchors, GlyphsApp does not propagate them.~~ It actually does, so this part is later reverted.
- When copying anchors, copy also userData, otherwise the context of contextual anchors of any processed glyph will be dropped. This is a regression from https://github.com/googlefonts/glyphsLib/pull/1011

### DIFF
--- a/Lib/glyphsLib/builder/transformations/propagate_anchors.py
+++ b/Lib/glyphsLib/builder/transformations/propagate_anchors.py
@@ -211,9 +211,6 @@ def anchors_traversing_components(
         component_transform = Transform(*component.transform)
         xscale, yscale = get_xy_rotation(component_transform)
         for anchor in anchors:
-            # skip contextual anchors
-            if anchor.name.startswith("*") and "GPOS_Context" in anchor.userData:
-                continue
             new_has_underscore = anchor.name.startswith("_")
             if (component_idx > 0 or has_underscore) and new_has_underscore:
                 continue
--- a/Lib/glyphsLib/builder/transformations/propagate_anchors.py
+++ b/Lib/glyphsLib/builder/transformations/propagate_anchors.py
@@ -211,6 +211,9 @@ def anchors_traversing_components(
         component_transform = Transform(*component.transform)
         xscale, yscale = get_xy_rotation(component_transform)
         for anchor in anchors:
+            # skip contextual anchors
+            if anchor.name.startswith("*") and "GPOS_Context" in anchor.userData:
+                continue
             new_has_underscore = anchor.name.startswith("_")
             if (component_idx > 0 or has_underscore) and new_has_underscore:
                 continue
@@ -299,6 +302,7 @@ def origin_adjusted_anchors(anchors: list[GSAnchor]) -> Iterable[GSAnchor]:
         GSAnchor(
             name=a.name,
             position=Point(a.position.x - origin.x, a.position.y - origin.y),
+            userData=dict(a.userData),
         )
         for a in anchors
         if a.name != "*origin"
@@ -398,7 +402,11 @@ def get_component_layer_anchors(
     if layer_anchors is not None:
         # return a copy as they may be modified in place
         layer_anchors = [
-            GSAnchor(name=a.name, position=Point(a.position.x, a.position.y))
+            GSAnchor(
+                name=a.name,
+                position=Point(a.position.x, a.position.y),
+                userData=dict(a.userData),
+            )
             for a in layer_anchors
         ]
     return layer_anchors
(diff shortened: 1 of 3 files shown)


Lib/glyphsLib/builder/transformations/propagate_anchors.py
```diff
@@ -211,6 +211,9 @@ def anchors_traversing_components(
         component_transform = Transform(*component.transform)
         xscale, yscale = get_xy_rotation(component_transform)
         for anchor in anchors:
+            # skip contextual anchors
+            if anchor.name.startswith("*") and "GPOS_Context" in anchor.userData:
```
khaledhosny, line 215: I’m being a bit conservative here and testing specifically for contextual anchors (`GPOS_Context`), but it might be that GlyphsApp does not propagate any special anchors, but I have no way of checking that.

schriftgestalt: Glyphs traverses all anchors. You can check in Glyphs by running this in the Macro panel:
```
print(Layer.anchorsTraversingComponents())
```

khaledhosny: I was testing by checking the generated feature file (using FEA Export plugin), and it had contextual anchor lookups only for glyphs that explicitly had them. `Layer.anchorsTraversingComponents()` does indeed propagate contextual anchors. Now I checked the GlyphsApp-generated font and it have the anchors propagated as well, so I guess it is the plug that mislead me.

Lib/glyphsLib/classes.py
```diff
@@ -2787,13 +2787,15 @@ def _serialize_to_plist(self, writer):
     _parent = None
     _defaultsForName = {"position": Point(0, 0)}
 
-    def __init__(self, name=None, position=None):
+    def __init__(self, name=None, position=None, userData=None):
         self.name = "" if name is None else name
         self._userData = None
```
anthrotype, line 2792: we are setting `self.userData` immediately below. is this `self._userData` still needed?

Lib/glyphsLib/builder/transformations/propagate_anchors.py
```diff
@@ -299,6 +299,7 @@ def origin_adjusted_anchors(anchors: list[GSAnchor]) -> Iterable[GSAnchor]:
         GSAnchor(
             name=a.name,
             position=Point(a.position.x - origin.x, a.position.y - origin.y),
+            userData=a.userData,
```
anthrotype, line 302: since we want to copy the anchors and this is a mutalbe object (a dict or some custom dict-like object, I don't remember?), it would be better to actually make a copy of the userData as well, so the two anchors are entirely independent from one another, and if one gets modified the other doesn't accidentally

Lib/glyphsLib/classes.py
anthrotype: oh I see. then maybe move them closer together and add a small comment?

khaledhosny: It is a pattern used across the file, so I’m not sure a comment here only is neccassary.

Lib/glyphsLib/classes.py
```diff
@@ -1207,6 +1207,12 @@ def __iter__(self):
         # instead of keys. It matches Glyphs.app though. Urg.
         yield from self._owner._userData.values()
 
+    def __copy__(self):
+        return dict(self)
+
+    def __deepcopy__(self, memo):
+        return {k: copy.copy(v) for k, v in self.items()}
```
anthrotype, line 1214: this isn't recursively "deep" copy, only one level deeper than shallow. Why do we need to customize `__copy__` and `__deepcopy__` for this class?

Lib/glyphsLib/builder/transformations/propagate_anchors.py
```diff
@@ -299,6 +300,7 @@ def origin_adjusted_anchors(anchors: list[GSAnchor]) -> Iterable[GSAnchor]:
         GSAnchor(
             name=a.name,
             position=Point(a.position.x - origin.x, a.position.y - origin.y),
+            userData=deepcopy(a.userData),
```
anthrotype, line 303: in any case, I think just a shallow `dicgt(a.userData)` should be sufficient for this, no need to deep copy or customise deepcopying of the UserDataProxy class

Lib/glyphsLib/classes.py
khaledhosny: So I first used deepcopy, then I found that `Proxy` (the parent class) has implmentations of `__copy__` and `__deepcopy__` that return lists (and `__deepcopy__` even class a non-existent `.copy() of the items), so I went ahead and tried to override in in `UserDataProxy`, missing the forest for the trees...

anthrotype:
thank you Khaled

PR #1038, opened 2024-10-21
[builder] Reorder .notdef and space glyphs in public.glyphOrder

khaledhosny:
Unless "Keep GlyphOrder" (or the older "TrueType Keep GlyphOrder") custom parameter is set to True.

### DIFF
--- a/Lib/glyphsLib/builder/custom_params.py
+++ b/Lib/glyphsLib/builder/custom_params.py
@@ -21,13 +21,13 @@ from glyphsLib.util import bin_to_int_list, int_list_to_bin
 from .filters import parse_glyphs_filter
 from .common import to_ufo_color
 from .constants import (
+    GLYPH_ORDER_KEY,
     GLYPHS_PREFIX,
     UFO2FT_COLOR_PALETTES_KEY,
     UFO2FT_FILTERS_KEY,
     UFO2FT_USE_PROD_NAMES_KEY,
     CODEPAGE_RANGES,
     REVERSE_CODEPAGE_RANGES,
-    PUBLIC_PREFIX,
     UFO_FILENAME_CUSTOM_PARAM,
     UFO2FT_META_TABLE_KEY,
 )
@@ -842,7 +842,7 @@ class GlyphOrderParamHandler(AbstractParamHandler):
 
     def to_glyphs(self, glyphs, ufo):
         if glyphs.is_font():
-            ufo_glyphOrder = ufo.get_lib_value(PUBLIC_PREFIX + "glyphOrder")
+            ufo_glyphOrder = ufo.get_lib_value(GLYPH_ORDER_KEY)
             if ufo_glyphOrder:
                 glyphs.set_custom_value("glyphOrder", ufo_glyphOrder)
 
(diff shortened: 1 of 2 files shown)


khaledhosny:
The regression tests failure is expected, it is all re-ordered `.notdef` and `space`.

khaledhosny:
Any objection, should I merge this PR?

anthrotype:
no objection from me, I just didn't have the time to review, but feel free to merge without me

anthrotype:
@khaledhosny I still cannot even add this "Keep GlyphrOrder" custom parameter, it's not shown in the list when try to add one.. Is one supposed to just add a seemingly unsupported parameter with that name? And what value? The string "1"?

Maybe we rushed a bit in adding support for this in glyphsLib if even the latest beta 3.3 doesn't have it.

khaledhosny:
Any value that glyphsLib considers True should work, or alternatively you can use “TrueType Keep GlyphOrder” which current version of Glyphs support (It does not show up when you type the name, but if you add it anyway, Glyphs recognizes it as a boolean custom parameter).

anthrotype:
they really don't want you to keep the glyph order 😅

khaledhosny:
Apparently

schriftgestalt:
The two "Keep Order" parameters are not in the list but can be still added by typing in the search file and click Add (They are kind of private, I really like to keep people from adding them without a very good reason).
The version from yesterday will display a checkbox instead of the text field.

anthrotype:
Don't you also find a bit weird that now "space" will take the second position even when the gid1 was specified to be the old "CR" glyph and the font's glyphOrder custom parameter is following the old TrueType convention about the first-four glyphs (.notdef, CR, NULL, space)?
Note that CR is also an empty glyph and would work just fine to prevent the old bug in Windows with COLRv0 fonts that prompted this whole space-glyph-must-be-gid1 business.

I really think this is too much hand-holding. If a font developer has bothered setting a glyphOrder custom parameter (which is already optional), why not just simply obeying that?... 

```
$ diff -u /Users/clupo/oss/fontc/build/default/fontc.GlyphOrder.ttx /Users/clupo/oss/fontc/build/default/fontmake.GlyphOrder.ttx
--- /Users/clupo/oss/fontc/build/default/fontc.GlyphOrder.ttx	2024-11-07 12:44:21
+++ /Users/clupo/oss/fontc/build/default/fontmake.GlyphOrder.ttx	2024-11-07 12:44:21
@@ -1,9 +1,9 @@
 <GlyphOrder>
     <!-- The 'id' attribute is only for humans; it is ignored when parsed. -->
     <GlyphID id="0" name=".notdef"/>
-    <GlyphID id="1" name="CR"/>
-    <GlyphID id="2" name="NULL"/>
-    <GlyphID id="3" name="space"/>
+    <GlyphID id="1" name="space"/>
+    <GlyphID id="2" name="CR"/>
+    <GlyphID id="3" name="NULL"/>
     <GlyphID id="4" name="nbspace"/>
     <GlyphID id="5" name="A"/>
     <GlyphID id="6" name="Aacute"/>
\ No newline at end of file
```

I'm reaaaaly tempted to just revert this. I really don't like this.

anthrotype:
actually this particular font (https://github.com/googlefonts/praise-script/blob/master/sources/Praise-Pro.glyphs) actually has them in the "wrong" order, NULL should be second and CR third (not the other way around) according to the old Apple Truetype spec, but it doesn't matter. NULL also doesn't have any contours and would be fine to have a second glyph as far as that Windows COLRv0 bug is concerned.

anthrotype:
wait.. I just tried to export a TTF from https://github.com/googlefonts/praise-script/blob/master/sources/Praise-Pro.glyphs directly from Glyphs.app (3.3 3325, the latest) and it does NOT reorder the space glyph in this particular case.

it has the following glyphOrder custom parameter, and it does not have that `Keep GlyphOrder` paramter for it's too new:

image

so glyphsLib is reordering the space, but Glyphs.app isn't.

@schriftgestalt is there some extra logic that we're missing? Does Glyphs.app only attempts to place space glyph as second when the glyph that is already in second place is not empty?

anthrotype:
the following archive contains two simple glyphs sources, one with the other wihout glyphOrder custom parameter.
When exported with Glyphs.app 3.3 3325, _neither_ of them has 'space' placed in second position.

[space-glyph-order.zip](https://github.com/user-attachments/files/17661679/space-glyph-order.zip)

is it the names I used or the fact the would-be second glyph is empty, I don't know.. But this example already produces a different output from glyphsLib 6.9.3, and I think a revert is in order here.

Issue #745, opened 2021-11-06
Designspace axis definitions have reversed userspace (input) and designspace (output) coordinates


jpt:
Given the following example `Axis Mappings` custom parameter:

```
{
customParameters = (
{
name = "Axis Mappings";
value = {
wght = {
22 = 100;
50 = 400;
72 = 700;
};
};
}
);
}
```

I get this output:

```
    <axis tag="wght" name="Weight" minimum="22" maximum="72" default="22">
      <map input="22" output="100"/>
      <map input="50" output="400"/>
      <map input="72" output="700"/>
    </axis>
```

Expected output:

```
    <axis tag="wght" name="Weight" minimum="100" maximum="700" default="100">
      <map input="100" output="22"/>
      <map input="400" output="50"/>
      <map input="700" output="72"/>
    </axis>
```

The dimension and xvalues elsewhere in the designspace file correctly point to the internal coordinates.

I wonder if this could be caused by the masters themselves also having remapped axis locations? For example on the master with `72` as the internal coordinate, the `Axis Location` custom parameter is defined as such:

```
{
customParameters = (
{
name = "Axis Location";
value = (
{
Axis = Weight;
Location = 700;
}
);
}
);
}
```

jpt:
I've attached a v2 Glyphs file that can reproduce the problem. Note that this fails entirely if saved as v3 (see #746)

[TestFontv2.glyphs.zip](https://github.com/googlefonts/glyphsLib/files/7491269/TestFontv2.glyphs.zip)

anthrotype:
@jpt thanks for filing this issue!

I just stumbled on this while tinkering with Glyphs.app, and I wonder how could this have gone unnoticed for so long. Apologies for that.

Ever since we added support for this "Axis Mappings" custom parameter, we have been interpreting it as a mapping from user-space to internal design-space coordinates, see:

https://github.com/googlefonts/glyphsLib/blob/74aefb387a5a84659902b53fa805b74ff3d8ed0c/Lib/glyphsLib/builder/axes.py#L183-L186

The key part is `ul, dl in sorted(mapping.items())`, where "ul" presumably means "_user_-space location" and "dl" means "_design_-space location" thus => {external: internal}.

However... Glyphs.app interprets it **the other way around**, as a map from {internal: external} coordinates.
[...]

PR #1040, opened 2024-10-31
Fix "Axis Mappings" interpretation

### LINKED ISSUE #993: Axis mappings completely upside down
Let's take [Nunito.glyphs](https://github.com/googlefonts/nunito/blob/main/sources/Nunito.glyphs) which is a Glyphs 2 file and convert it to designspace. The weight axis mappings look like this:

```xml
    <axis tag="wght" name="Weight" minimum="200" maximum="1000" default="200">
      <map input="200" output="42"/>
      <map input="300" output="61"/>
      <map input="400" output="81"/>
      <map input="500" output="91"/>
      <map input="600" output="101"/>
      <map input="700" output="125"/>
      <map input="800" output="151"/>
      <map input="900" output="178"/>
      <map input="1000" output="208"/>
    </axis>
```

In Glyphs, I open the file, change it to file format 3 and save it. When run through glyphs2ufo, the axis mapping now looks like this:

```xml
    <axis tag="wght" name="Weight" minimum="42" maximum="208" default="42">
      <map input="42" output="200"/>
      <map input="61" output="300"/>
      <map input="81" output="400"/>
      <map input="91" output="500"/>
      <map input="101" output="600"/>
      <map input="125" output="700"/>
      <map input="151" output="800"/>
      <map input="178" output="900"/>
      <map input="208" output="1000"/>
    </axis>
```

I think this is because Glyphs changed its understanding of the Axis Mapping custom parameter, so that's a pain but it probably can't be helped.

Let's get rid of the Axis Mapping custom parameter, save the Glyphs file, and try again, relying on the axis coordinates in the instances to provide the mapping.

```xml
 <axis tag="wght" name="Weight" minimum="1" maximum="1000" default="200">
      <map input="1" output="208"/>
      <map input="200" output="42"/>
      <map input="300" output="61"/>
      <map input="400" output="91"/>
      <map input="500" output="91"/>
      <map input="600" output="101"/>
      <map input="700" output="125"/>
      <map input="800" output="151"/>
      <map input="900" output="178"/>
      <map input="1000" output="208"/>
    </axis>
```

Where did that "1" come from?

simoncozens:
This fixes #993. We look at the file format version, app version and instance mapping to determine the correct interpretation of the "Axis Mappings" custom parameter.

### DIFF
--- a/Lib/glyphsLib/builder/axes.py
+++ b/Lib/glyphsLib/builder/axes.py
@@ -187,10 +187,22 @@ def to_designspace_axes(self):
         ):
             axis_wanted = True
 
+        # Build the mapping from the instances because they have both
+        # a user location and a design location; we build this here
+        # because we use it to check if a custom mapping needs to be inverted.
+        instance_mapping = {}
+        update_mapping_from_instances(
+            instance_mapping,
+            self.font.instances,
+            axis_def,
+            minimize_glyphs_diffs=self.minimize_glyphs_diffs,
+        )
+
         # See https://github.com/googlefonts/glyphsLib/issues/568
         if custom_mapping:
             if axis.tag in custom_mapping:
                 mapping = {float(k): v for k, v in custom_mapping[axis.tag].items()}
+                mapping = _potentially_invert_mapping(self, mapping, instance_mapping)
                 regularDesignLoc = axis_def.get_design_loc(regular_master)
                 reverse_mapping = {dl: ul for ul, dl in sorted(mapping.items())}
                 regularUserLoc = piecewiseLinearMap(regularDesignLoc, reverse_mapping)
@@ -230,16 +242,6 @@ def to_designspace_axes(self):
             regularDesignLoc = axis_def.get_design_loc(regular_master)
             regularUserLoc = axis_def.get_user_loc(regular_master)
         else:
-            # Build the mapping from the instances because they have both
-            # a user location and a design location.
-            instance_mapping = {}
-            update_mapping_from_instances(
-                instance_mapping,
-                self.font.instances,
-                axis_def,
-                minimize_glyphs_diffs=self.minimize_glyphs_diffs,
-            )
-
             master_mapping = {}
             for master in self.font.masters:
                 # Glyphs masters don't have a user location
@@ -320,6 +322,51 @@ def font_uses_axis_locations(font):
     return font.axes and all(
         master.customParameters["Axis Location"] for master in font.masters
     )
+
+
+def _potentially_invert_mapping(self, mapping, instance_mapping):
+    # Fontmake introduced an Axis Mapping custom parameter before Glyphs did;
+    # then Glyphs introduced it, but with an inverted interpretation. Then
+    # Glyphs changed the interpretation to match fontmake's. Try and figure out
+    # what's going on.
+    # See https://github.com/googlefonts/glyphsLib/issues/745
+
+    # We want to return a userspace->designspace map for our purposes
+    inverted = {v: k for k, v in mapping.items()}
+    if int(self.font.appVersion) >= 3168:
+        # Mapping is *definitely* designspace->userspace, invert
+        return inverted
+    if self.font.format_version != 3:
+        # Glyphs wasn't using this custom parameter, only we were
+        return mapping
+    # Let's get heuristical
+    if min(mapping.keys()) == min(mapping.values()) or max(mapping.keys()) == max(
+        mapping.values()
+    ):
+        # No way of knowing
+        logger.warning(
+            "Axis %s: The Axis Mappings custom parameter is ambiguous, "
+            "assuming it is in userspace->designspace format. Please check the mapping "
+            "and save in a recent version of Glyphs to avoid this warning."
+        )
+        return mapping
+    if min(mapping.keys()) == min(instance_mapping.keys()) and max(
+        mapping.keys()
+    ) == max(instance_mapping.keys()):
+        # Looks like userspace->designspace
+        return mapping
+    if min(mapping.keys()) == min(instance_mapping.values()) and max(
+        mapping.values()
+    ) == max(instance_mapping.keys()):
+        # Looks like designspace->userspace
+        return inverted
+    # No idea
+    logger.warning(
+        "Axis %s: The Axis Mappings custom parameter is ambiguous, "
+        "assuming it is in userspace->designspace format. Please check the mapping "
+        "and save in a recent version of Glyphs to avoid this warning."
+    )
+    return mapping
 
 
 def to_glyphs_axes(self):
--- a/tests/builder/axes_test.py
+++ b/tests/builder/axes_test.py
@@ -740,3 +740,47 @@ def test_virtual_masters_extend_min_max_for_unmapped_axis(ufo_module, datadir):
     assert [
         cp.value for cp in font2.customParameters if cp.name == "Virtual Master"
     ] == virtual_masters
+
+
+def test_axis_mappings_different_interpretations(ufo_module, datadir):
+    # https://github.com/googlefonts/glyphsLib/issues/859
+    font = GSFont(datadir.join("gf", "Rubik.glyphs"))  # Saved under Glyphs 3079
+    expected_map = [
+        (300.0, 60),
+        (400.0, 90),
+        (500.0, 125),
+        (600.0, 142),
+        (700.0, 160),
+        (800.0, 190),
+        (900.0, 220),
+    ]
+    assert "Axis Mappings" in font.customParameters
+    ds = to_designspace(font, ufo_module=ufo_module)
+    assert ds.axes[0].minimum == 300
+    assert ds.axes[0].maximum == 900
+    assert ds.axes[0].map == expected_map
+
+    font.format_version = 3  # Ambiguous, work out interpretation hueristically
+    ds = to_designspace(font, ufo_module=ufo_module)
+    assert ds.axes[0].minimum == 300
+    assert ds.axes[0].maximum == 900
+    assert ds.axes[0].map == expected_map
+
+    # Pretend we saved it under a new version
+    font.appVersion = 3217
+    font.customParameters["Axis Mappings"] = {
+        "wght": {
+            60: 300,
+            90: 400,
+            125: 500,
+            142: 600,
+            160: 700,
+            190: 800,
+            220: 900,
+        }
+    }
+
+    ds = to_designspace(font, ufo_module=ufo_module)
+    assert ds.axes[0].minimum == 300
+    assert ds.axes[0].maximum == 900
+    assert ds.axes[0].map == expected_map


tests/builder/axes_test.py
```diff
@@ -740,3 +740,47 @@ def test_virtual_masters_extend_min_max_for_unmapped_axis(ufo_module, datadir):
     assert [
         cp.value for cp in font2.customParameters if cp.name == "Virtual Master"
     ] == virtual_masters
+
+
+def test_axis_mappings_different_interpretations(ufo_module, datadir):
+    # https://github.com/googlefonts/glyphsLib/issues/859
+    font = GSFont(datadir.join("gf", "Rubik.glyphs"))  # Saved under Glyphs 3079
```
anthrotype, line 747: this Rubik.glyphs seems to already have a correct user->design mapping even if it was saved with an old version:

https://github.com/googlefonts/glyphsLib/blob/1cb4fc5ae2cf385df95d2b7768e7ab4eb60a5ac3/tests/data/gf/Rubik.glyphs#L49-L59

Lib/glyphsLib/builder/axes.py
```diff
@@ -322,6 +324,51 @@ def font_uses_axis_locations(font):
...
+    # then Glyphs introduced it, but with an inverted interpretation. Then
+    # Glyphs changed the interpretation to match fontmake's. Try and figure out
+    # what's going on.
+    # See https://github.com/googlefonts/glyphsLib/issues/745
+
+    # We want to return a userspace->designspace map for our purposes
+    inverted = {v: k for k, v in mapping.items()}
+    if int(self.font.appVersion) >= 3168:
+        # Mapping is *definitely* designspace->userspace, invert
+        return inverted
```
anthrotype, line 338: i'm confused, I thought older versions (< 3168?) would interpret it as design->user and thus require flip, whereas newer versions have aligned with fontmake's interpretation and alreadu have user->design? This block seems to contradict my understanding, for it's an inverted mapping if the version is more recent.

Lib/glyphsLib/builder/axes.py
```diff
@@ -322,6 +324,51 @@ def font_uses_axis_locations(font):
...
+    if self.font.format_version != 3:
+        # Glyphs wasn't using this custom parameter, only we were
+        return mapping
+    # Let's get heuristical
```
anthrotype, line 342: hm i'm a bit skeptical we need to go "heuristical".. Can't we simply check the version and do accordingly? If the source file doesn't match the interpretation for the version it was saved with, we can say it's not our problem

schriftgestalt: I think there were some fonts that would set it "wrong" in Glyphs to make it work with fontMake.

tests/builder/axes_test.py
simoncozens: Right, because it was a format 2 file.

Lib/glyphsLib/builder/axes.py
simoncozens: "Interpret as" is perhaps not helpful here. :-) I just opened Nunito source, converted to Glyphs 3 file format and saved under Glyphs 3310. The custom parameter is *stored as*:

```
wght = {
42 = 200;
61 = 300;
81 = 400;
91 = 500;
101 = 600;
125 = 700;
151 = 800;
178 = 900;
208 = 1000;
};
```

i.e. designspace to userspace.

Lib/glyphsLib/builder/axes.py
simoncozens: Exactly.

Lib/glyphsLib/builder/axes.py
anthrotype: now I am _extremely_ confused

anthrotype: well, I for one don't understand this whole mess. It would be nice to have a recap.

Lib/glyphsLib/builder/axes.py
anthrotype: how about fixing the actual sources instead?

Lib/glyphsLib/builder/axes.py
anthrotype: > I can compile fonts that were not compiling before

good, what I'm worried about are the other fonts that were compiling before and may not after this.
Right now (tested with 3324) on a newly created glyphs3 file, the Axis Mappings is already stored in the source as user->design (or external->internal, as Glyphs calls them), so it is correct and matches fontmake. Wouldn't this PR invert it? I'm seeing 

if int(self.font.appVersion) >= 3168:
    return inverted

Lib/glyphsLib/builder/axes.py
```diff
@@ -322,6 +324,51 @@ def font_uses_axis_locations(font):
     )
 
 
+def _potentially_invert_mapping(self, mapping, instance_mapping):
+    # Fontmake introduced an Axis Mapping custom parameter before Glyphs did;
+    # then Glyphs introduced it, but with an inverted interpretation. Then
+    # Glyphs changed the interpretation to match fontmake's. Try and figure out
```
anthrotype, line 330: when we say "Glyphs changed the interpretation to match fontmake's", does that mean it simply took the existing parameter's keys are meaning user/external and the values design/internal but keep them as were originally in the source... or does it mean that it suddenly decided to swap them?
Is this PR attempting to un-flip them?!

Lib/glyphsLib/builder/axes.py
simoncozens: You are going to love this.

Glyphs now stores user->design for new fonts, but it "upgrades" old fonts by *simply inverting them*. Take your new font, manually edit "appVersion" to (say) 3110, open it in Glyphs and save it again - the mappings will be flipped.

So I guess this means we can't trust the app version number, and we have to probe against the instance mapping.

khaledhosny: We could have went with my suggestion (back in #745) and made glyphsLib match (old) GlyphsApp behavior and fixed the few fonts that would have been broken by this. Now it is even a bigger mess...

simoncozens: I don't deny that it's stupid and wrong and there have may be been questionable choices on both sides, but in 95% of cases the interpretation is completely obvious. Do you have a weight uservalue of 7? You did something wrong. Did all the master locations suddenly fall outside of the design space and you have no valid masters any more? You got the axis mappings backwards. Compare against the instance locations, because we know which way round they are, and if the ranges are wildly different, switch it. It's a mess, but it's not a hard mess to clean up.

Lib/glyphsLib/builder/axes.py
simoncozens: It literally swaps them.

anthrotype:
we do already check the instances when there's _no_ axis mappings nor axis location parameters to infer the master's user-space coordinates based on the instances, and that inference has already proved a bit brittle when the user reports unexpected "base master not fund" or weird interpolations because of unexpected mappings. 
Now we are saying that Axis Mappings, which was supposed to give a full explicit definitions of the mappings, not tied to the masters or instances, can't be trusted and we'll still try to fix it by looking at the instances.. but the instances in a VF at least do _need_ to be there, nor they need to be defined at all the axes corners. Is there a chance that this new heuristic to fix up a few sources at compile time may end up screwing perfectly valid fonts?

anthrotype:
(i hate when I forget a _not_ and the whole sentence stop making sense)

anthrotype:
besides, the inference of user->design mappings from the instances is only meant to work for weight and width, IIRC. And it's unlikely that these instances have Axis Location custom parameter if there's already an Axis Mappings, which in theory trumps Axis Location as it's more general

simoncozens:
Ok. So how would you like to fix the issue?

schriftgestalt:
I would not add any axis mapping if there is nothing explicitly in the file. Maybe issue a warning.

simoncozens:
That doesn't fix the problem. The problem is we can't build fonts because the Axis Mappings that are in the file aren't interpreted correctly.

schriftgestalt:
There are only very few files that have those faulty axis mappings. That has to be caught in QA and fixed in the files.

simoncozens:
>  faulty axis mappings

OK, one last time. These are *not faulty* for Glyphs.app users. Glyphs.app compiles these things just fine, and produces VFs with correct avar mappings. glyphsLib does not. A good example is [Cormorant.glyphs](https://github.com/CatharsisFonts/Cormorant/blob/master/sources/Cormorant.glyphs).

The whole purpose of glyphsLib is to compile Glyphs files so we don't have to use Glyphs.app in an open source toolchain. I don't think we should decide that we're just not going to bother to compile *valid* Glyphs files that work fine in Glyphs.

schriftgestalt:
Then misunderstood the problem. Sorry. 
But Glyphs doesn’t export a working avar for Cormorant. It needs "Axis Location" parameters in the master settings.

schriftgestalt:
As I understand it (and how Glyphs handles it) is such:
Use the axis range from the masters (either the coordinates or the axis locations; those are called internal and external coordinates in the Glyphs 4 API) for fvar and such. 
Then use the "Axis Location" in the instances or "Axis Mapping" in the font to build a avar table.

simoncozens:
> But Glyphs doesn’t export a working avar for Cormorant. 

But it does.

```
$ ttx -o - -t avar CormorantVF.ttf
Dumping "CormorantVF.ttf" to "<stdout>"...
<?xml version="1.0" encoding="UTF-8"?>
<ttFont sfntVersion="\x00\x01\x00\x00" ttLibVersion="4.54">

Dumping 'avar' table...
  <avar>
    <version major="1" minor="0"/>
    <segment axis="wght">
      <mapping from="-1.0" to="1.0"/>
      <mapping from="0.0" to="0.0"/>
      <mapping from="0.2" to="1.0"/>
      <mapping from="0.45" to="1.0"/>
      <mapping from="0.725" to="1.0"/>
      <mapping from="1.0" to="1.0"/>
    </segment>
  </avar>
```

simoncozens:
Urgh, you're right.

anthrotype:
Cormorant.glyhps contains these incorrect mappings, with keys/values swapped:

```
{
name = "Axis Mappings";
value = {
wght = {
40 = 300;
48 = 400;
58 = 500;
69 = 600;
80 = 700;
};
```

Even Glyphs.app (version 3.3 build 3324) displays the first column (the keys) as "external" and the second column (the values) as "internal" (whereas it should be the other way around, but Glyphs.app is just reading what's in the source).

If I just look at the source, I could say with confidence that the source itself contains a bug. Maybe it was not set up like this by the font developer, and the source originally contained mappings in the order that fontmake expects but then got automatically "fixed" by Glyphs.app, I don't know. But it still remains buggy. And even Glyphs.app itself can't export a correct avar from this, not just fontmake.

The way I'd suggest fixing this is by modifying the custom parameter in the source, swapping the values with keys.

anthrotype:
> It needs "Axis Location" parameters in the master settings.

for fontmake at least, you don't need _both_ Axis Mappings and Axis Locations because the former encompasses the latter and thus instead of risking duplicating the same information twice (or havingt them out of sync), we only look at Axis Mappings if present, otherwise we go find Axis Locations.

PR #1041, opened 2024-10-31
Don't unnecessarily drop LTR kerns when combining with RTL

anthrotype:
Fixes #1039

```diff
--- master_ufo/PlaypenSans-Regular.ufo/kerning.plist	2024-10-31 10:32:42
+++ master_ufo_2/PlaypenSans-Regular.ufo/kerning.plist	2024-10-31 10:32:52
@@ -14,7 +14,11 @@
     </dict>
     <key>public.kern1.quotesingle</key>
     <dict>
+      <key>public.kern2.A</key>
+      <integer>-49</integer>
       <key>public.kern2.bet-hb</key>
+      <integer>-20</integer>
+      <key>public.kern2.space</key>
       <integer>-20</integer>
     </dict>
     <key>public.kern1.space</key>
```

Issue #1039, opened 2024-10-30
Merging of kerningLTR and kerningRTL drops kerns


simoncozens:
The attached Glyphs file ([PlaypenSans-mini.glyphs.gz](https://github.com/user-attachments/files/17578383/PlaypenSans-mini.glyphs.gz)) contains a LTR kern table:

```
kerningLTR = {
"5FC9277C-B92C-4C3C-88AB-70E16AEC7C11" = {
"@MMK_L_A" = {
"@MMK_R_quotesingle" = -49;
};
"@MMK_L_quotesingle" = {
"@MMK_R_A" = -49;
"@MMK_R_space" = -20;
};
"@MMK_L_space" = {
"@MMK_R_quotesingle" = -20;
};
};
};
```

and an RTL kern table:

```
kerningRTL = {
"5FC9277C-B92C-4C3C-88AB-70E16AEC7C11" = {
"@MMK_R_bet-hb" = {
"@MMK_L_quotesingle" = -30;
};
"@MMK_R_quotesingle" = {
"@MMK_L_bet-hb" = -20;
};
"@MMK_R_vav-hb" = {
"@MMK_L_bet-hb" = 16;
};
};
};
```

When processed with `glyphs2ufo` (or fontmake) the resulting `kerning.plist` is:

```
<?xml version='1.0' encoding='UTF-8'?>
<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
<plist version="1.0">
  <dict>
    <key>public.kern1.A</key>
    <dict>
      <key>public.kern2.quotesingle</key>
      <integer>-49</integer>
    </dict>
    <key>public.kern1.bet-hb</key>
    <dict>
      <key>public.kern2.quotesingle</key>
      <integer>-30</integer>
    </dict>
    <key>public.kern1.quotesingle</key>
    <dict>
      <key>public.kern2.bet-hb</key>
      <integer>-20</integer>
    </dict>
    <key>public.kern1.space</key>
    <dict>
      <key>public.kern2.quotesingle</key>
      <integer>-20</integer>
    </dict>
    <key>public.kern1.vav-hb</key>
    <dict>
      <key>public.kern2.bet-hb</key>
      <integer>16</integer>
    </dict>
  </dict>
</plist>
```

Notice that while `A quotesingle -49` appears in the UFO, `quotesingle A -49` has been dropped.

[...]

PR #1042, opened 2024-10-31
modify `test_glyphs3_rtl_kerning` to add a mix of LTR and RTL kerns

anthrotype:
to reproduce #1039

Will be fixed by #1041


anthrotype:
I borrowed the kerning from @simoncozens' [test file](https://github.com/user-attachments/files/17578383/PlaypenSans-mini.glyphs.gz) adapting an existing test for simplicity.
It's expected to fail with `main`

PR #1044, opened 2024-11-07
Revert "Merge pull request #1038 from googlefonts/space-glyphOrder"

anthrotype:
This reverts PR #1038


anthrotype:
Until we understand how this is exactly supposed to work I suggest we revert #1038

When in fontc_crater (a tool we use to compare fontc vs fontmake output against all of googlefonts library) we bumped glyphsLib to 6.9.3, immediately hundreds of fonts that were identical before started showing diffs in the "GlyphOrder" ttx pseudo-table, with the space shifted to second position. I don't think one would expect to see this in a patch version bump.

khaledhosny:
fontc_crater could normalize glyphOrder if it does not wish to follow glyphsLib here. I don’t think that is a reason to revert this change. And even if we learnt more about how this is supposed to work, it will still be a change compared to fontc.

khaledhosny:
On a second thought, it does not really matter that much to me, so feel free to merge this PR.

khaledhosny:
You probably should revert the ufo2ft commit as well.

anthrotype:
ufo2ft would only put space after .notdef if no public.glyphOrder is defined, which is a clear signal the font developer doesn't care about a specific ordering and the compiler can decide on their behalf.

You did ask if anyone had any objections and I too hastily said I didn't, so I'll take the blame :)

PR #1046, opened 2024-11-13
Bugfixes

roberto-arista:
To be honest, the `userLoc` issue might be solved upstream, but this was a quick fix I was able to use to run my code.

### DIFF
--- a/Lib/glyphsLib/builder/axes.py
+++ b/Lib/glyphsLib/builder/axes.py
@@ -210,7 +210,8 @@ def to_designspace_axes(self):
                         f"the mapping for user location {userLoc} "
                         f"from {mapping[userLoc]} to {designLoc}"
                     )
-                mapping[userLoc] = designLoc
+                # sometimes userLoc is a str, so we need to convert it to float 🤷‍♂️
+                mapping[float(userLoc)] = designLoc
 
             update_mapping_from_instances(
                 mapping,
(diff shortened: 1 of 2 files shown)


anthrotype:
ciao Roberto!
i see this PR is based on Glyphs3-merge branch (which I haven't reviewed/used yet) but at first sight both these changes look suspicious to me.

> sometimes userLoc is a str

when exactly and why?

> switch from .disabled to .active

how can an attribute that was named "disabled" be renamed "active" while seemingly keeping the same semantics?

roberto-arista:
`.disabled` seems to have been deprecated:
https://github.com/googlefonts/glyphsLib/blob/ea5d7eb9e0320541ca1ec71299f8bb8b76a6e7da/Lib/glyphsLib/classes.py#L4147-L4153

roberto-arista:
I totally agree about the `userLoc` fix, I just wanted to raise the issue. Probably the data needs some validation when it's parsed. We are using this branch for a QA tool we're working on at Dinamo and while parsing some files I encountered some `userLoc` in string form. I'll try to reproduce it.

schriftgestalt:
I fixed the exception.

schriftgestalt:
But I had fixed the "disabled" > "active" API but forgot to push. So now there is a conflict ;(

roberto-arista:
I can close the pull request if you fixed it on your branch, let me know how you prefer to proceed.

PR #1048, opened 2024-11-14
Add mapping for uniqueID

rsheeter:


anthrotype (approved):
LGTM, I added a test as well

Issue #807, opened 2022-09-08
Can't generate font >> invalid literal for int() with base 10: 'bit 29'


etunni:
Loading Glyphs file failed: invalid literal for int() with base 10: 'bit 29'

etunni:
I was able to generate the font files, finally.
The source files traveled from Glyphs 2 to G3 and back during the design process, perhaps some data in the windows: info / masters / export are incompatible between G2 and G3 and some wrong data still remaining in internal code.
I deleted and entered again the same data and the source was fixed

guidoferreyra:
I get this erorr ver often when glyphs file has unicodeRanges and/or codepageRanges custom parameters

anthrotype:
glyphsLib is trying to create an integer with base 10 from a string like int("bit 29"), which of course fails with ValueError. 

Can you search for the string "bit 29" in your source file and copy/paste it here with a few lines before and after for context?

I'd like to see where it shows up to understand where it's coming from, thanks

guidoferreyra:
```
{
name = codePageRanges;
value = (
"1252",
"1250",
"1254",
"1257",
"bit 29"
);
```
"bit 29" corresponds to 
<img width="293" alt="imagen" src="https://user-images.githubusercontent.com/1755590/197017092-608a2411-b369-42b7-85cf-8be15133fbee.png">

schriftgestalt:
The values in the codePageRanges are not numbers but strings. That are the names of the codepages. But some bits don’t have a names assigned but some people still set them.

anthrotype:
closing as duplicate of https://github.com/googlefonts/glyphsLib/issues/530

Issue #530, opened 2019-06-19
Handle new codepage ranges syntax


madig:
```
name = codePageRanges;
value = (
1252,
bit 29
);
},
```

schriftgestalt:
I fixed the missing quotes around the `bit 29` in the wrong branch. Just fixed that properly.

madig:
What do you mean? Did you change it in Glyphs?

schriftgestalt:
I fixed it in Glyphs right now. Will be in v1245

madig:
Mh. Then I guess we have to support both. One day.

schriftgestalt:
But the bit should not be set in most fonts. At least I don’t know why one would do that.

felipesanches:
I got glyphs files with this and fontmake build fails because of that. Can the parser be made tolerant to consume this corrupt file?

felipesanches:
The affected files are here: https://github.com/MonicaRizzolli/Tomorrow/tree/e3ad9978645caeb563082c1c2f762872685214e5/src

anthrotype:
Sure, PRs are welcome :)
thanks for the link to the file.

anthrotype:
@felipesanches if you're also at atypi, if you like we could work together on this later

felipesanches:
yes, I am. ;-D

anthrotype:
oh wait a second. the issue here is the lack of quotes around `bit 29`? If that's the case I would prefer not to modify the parser (which is already quite complicated), and leave this as wontfix. A user can always open and save this corrupt file with a more recent version of Glyphs.app, which I believe should fix the problem. Am I right?

anthrotype:
Ah now I see, this is a feature request to support a new notation for codePageRanges (`bit XX` where `XX` is a number), which glyphsLib doesn't handle these yet. It's not about the parser handling the invalid unquoted string. That should fail normally.
We need to implement this special `bit XX` notation.
Related: https://github.com/googlefonts/glyphsLib/pull/390

@madig where did you get that `codePageRangesUnsupportedBits` custom parameter from? I can't find any reference to it.

anthrotype:
[...]

PR #1061, opened 2025-01-27
Support new codepage ranges syntax #530

black7375:
Currently the problem is occurring in the parser. #807
```
File "/Users/etunni/venv/lib/python3.7/site-packages/glyphsLib/parser.py", line 81, in _parse_dict
self._parse_dict_into_object(res, text)
File "/Users/etunni/venv/lib/python3.7/site-packages/glyphsLib/parser.py", line 98, in _parse_dict_into_object
res[name] = d[name]
File "/Users/etunni/venv/lib/python3.7/site-packages/glyphsLib/classes.py", line 336, in setitem
setattr(self, key, value)
File "/Users/etunni/venv/lib/python3.7/site-packages/glyphsLib/classes.py", line 1366, in setValue
value = readIntlist(value)
File "/Users/etunni/venv/lib/python3.7/site-packages/glyphsLib/types.py", line 349, in readIntlist
return _mutate_list(int, src)
File "/Users/etunni/venv/lib/python3.7/site-packages/glyphsLib/types.py", line 344, in _mutate_list
l[i] = fn(l[i])
ValueError: invalid literal for int() with base 10: 'bit 29'
```

Therefore, pre-classify into `codePageRanges` and `codePageRangesUnsupportedBits`,
https://github.com/googlefonts/glyphsLib/blob/b0fc3bcc14ba0311a7b9a6e699010aeb01899ac2/Lib/glyphsLib/builder/custom_params.py#L467-L502

For `codePageRangesUnsupportedBits`, just use the number with the `bit ` prefix removed.
https://github.com/googlefonts/glyphsLib/blob/b0fc3bcc14ba0311a7b9a6e699010aeb01899ac2/Lib/glyphsLib/builder/constants.py#L157-L196

Note: The error occurs before handling `custom_params`, so it should be fixed in the parser.

### DIFF
--- a/Lib/glyphsLib/builder/custom_params.py
+++ b/Lib/glyphsLib/builder/custom_params.py
@@ -492,7 +492,7 @@ class OS2CodePageRangesParamHandler(AbstractParamHandler):
             if codepages is None:
                 return
 
-        ufo_codepage_bits = [CODEPAGE_RANGES[v] for v in codepages]
+        ufo_codepage_bits = self._convert_to_bits(codepages)
         unsupported_codepage_bits = glyphs.get_custom_value(
             "codePageRangesUnsupportedBits"
         )
@@ -500,6 +500,34 @@ class OS2CodePageRangesParamHandler(AbstractParamHandler):
             ufo_codepage_bits.extend(unsupported_codepage_bits)
 
         ufo.set_info_value("openTypeOS2CodePageRanges", sorted(ufo_codepage_bits))
+
+    @staticmethod
+    def _convert_to_bits(codepages):
+        ufo_codepage_bits = []
+        for codepage in codepages:
+            if isinstance(codepage, int):
+                ufo_codepage_bits.append(CODEPAGE_RANGES[codepage])
+
+            elif isinstance(codepage, str):
+                if codepage.isdigit():
+                    ufo_codepage_bits.append(CODEPAGE_RANGES[int(codepage)])
+                elif codepage.startswith("bit "):
+                    try:
+                        ufo_codepage_bits.append(int(codepage[4:]))
+                    except ValueError as err:
+                        raise ValueError(
+                            f"'{codepage}' is not in correct format. "
+                            "A number must follow after 'bit '."
+                        ) from err
+                else:
+                    raise ValueError(
+                        f"'{codepage}'is neither a number nor 'bit ' format."
+                    )
+
+            else:
+                raise TypeError(f"Unsupported type: {type(codepage)}")
+
+        return ufo_codepage_bits
 
 
 register(OS2CodePageRangesParamHandler())
--- a/Lib/glyphsLib/classes.py
+++ b/Lib/glyphsLib/classes.py
@@ -1357,14 +1357,12 @@ class GSCustomParameter(GSBase):
     _CUSTOM_INTLIST_PARAMS = frozenset(
         (
             "fsType",
-            "openTypeOS2CodePageRanges",
             "openTypeOS2FamilyClass",
             "openTypeOS2Panose",
             "openTypeOS2Type",
             "openTypeOS2UnicodeRanges",
             "panose",
             "unicodeRanges",
-            "codePageRanges",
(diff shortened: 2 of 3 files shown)


black7375:
Now, I have completed CLA.
image

anthrotype:
thanks. Though I don't think the parser should import glyphsLib.builder or special-case those custom parameters every time it's parsing any object from a dict. I think what I wrote here still stands:

https://github.com/googlefonts/glyphsLib/issues/530#issuecomment-1705048940

> in order to be able to support parsing the official codePageRanges custom parameter properly, we need to not treat it as an integer list but as a List[Union[int, str]] where str is one of those "bit XX".

We could probably just remove `codePageRanges` from the `GSCustomParameter._CUSTOM_INTLIST_PARAMS`:

https://github.com/googlefonts/glyphsLib/blob/b0fc3bcc14ba0311a7b9a6e699010aeb01899ac2/Lib/glyphsLib/classes.py#L1367

and then the parser will not attempt to force all elements into integers.

When when the builder is converting to the equivalent UFO fontinfo property, the string values starting with "bit " prefix could be handlded specially.

black7375:
Thanks for quick replay.

As I tested, if I simply passed the value, a problem occurred in next:
```python
    # glyphsLib/Lib/glyphsLib/builder/custom_params.py
    ufo.set_info_value("openTypeOS2CodePageRanges", sorted(ufo_codepage_bits))
```

Here, the codepage should be converted to a bit number list.

anthrotype:
Indeed, glyphsLib/builder/custom_params.py is a good place where that conversion can take place. The parser itself can simply parse the raw list as it is, with some integers and some strings. Does that make sense?

black7375:
Thank you
Then, I'll process the conversions in `custom_params`, reflect Lint, and then force-push.

black7375:
I fixed it and it's definitely better. 👍

anthrotype:
GSCustomParameter._CUSTOM_INTLIST_PARAMS also contains a "openTypeOS2CodePageRanges", I think it's a synonym of "codePageRanges" custom parameter so should be treated the same, i.e. remove it from that list so it doesn't get forced into a list of ints

Lib/glyphsLib/builder/custom_params.py
```diff
@@ -501,6 +501,32 @@ def to_ufo(self, builder, glyphs, ufo):
 
         ufo.set_info_value("openTypeOS2CodePageRanges", sorted(ufo_codepage_bits))
 
+    def _convert_to_bits(self, codepages):
```
anthrotype, line 504: this could be decorated as a `@staticmethod` perhaps since it is not using `self`

anthrotype (approved):

anthrotype:
maybe don't make other unrelated changes within the same PR. You can do a follow up for those if you like.

Do you mind also adding a test?

black7375:
When I run `flake8 Lib tests`, the following error appears. Is it okay if I don't change the rest?
```
Lib/glyphsLib/builder/constants.py:265:5: B041 Repeated key-value pair in dictionary literal.
Lib/glyphsLib/builder/custom_params.py:518:25: B904 Within an `except` clause, raise exceptions with `raise ... from err` or `raise ... from None` to distinguish them from errors in exception handling.  See https://docs.python.org/3/tutorial/errors.html#exception-chaining for details.
Lib/glyphsLib/builder/custom_params.py:519:105: B950 line too long (104 > 80 characters)
Lib/glyphsLib/builder/glyph.py:63:12: E226 missing whitespace around arithmetic operator
```

Ok, I'll add the test to [`tests/builder/custom_params_test.py`](https://github.com/googlefonts/glyphsLib/blob/76e49f1eb0a9a7ecbf598c7a01530e4c87160cce/tests/builder/custom_params_test.py#L221).

anthrotype:
> When I run flake8 Lib tests, the following error appears

as long as the CI "lint" job passes, I don't really care. You can send a follow up PR to fix those if you like

> I'll add the test

thank you

PR #1074, opened 2025-03-16
Change method of determining glyph height

### LINKED ISSUE #1071: Change glyph height values to reflect vheaVert metrics instead of sTypo metrics (if available)
In order for `ufo2ft` to determine that a font is vertical and needs a `vhea` and `vmtx` table, it must have the `vheaVertAscender` / `vheaVertDescender` / `vheaVertLineGap` values assigned. See [comment here](https://github.com/googlefonts/glyphsLib/blob/c4db6b981d577f456d64ebe9993818770e170454/Lib/glyphsLib/builder/masters.py#L74).

However, the glyph vertical height, if not explicitly declared by the designer for a given glyph, is set [based on the sTypoMetrics](https://github.com/googlefonts/glyphsLib/blob/c4db6b981d577f456d64ebe9993818770e170454/Lib/glyphsLib/builder/glyph.py#L359).

In order to support the new approach for CJK fonts that was [agreed upon](https://github.com/google/fonts/issues/8911), the glyph height generation should be changed to prioritize the vheaAscender / vheaDescender values as the UPM / emBox is the defining characteristic for the standard height of CJK glyphs, leaving sTypo as fallback as necessary. This way, folks building fonts the "old way" (where the sTypo reflects the emBox) can still have that functionality, but also allows people to build fonts the "new way" where the sTypo have been set to achieve better line spacing in horizontal scenarios. 

@schriftgestalt I have also filed a related bug on the Glyphs forum.
- https://forum.glyphsapp.com/t/vmtx-table-determined-via-stypo-values-instead-of-vhea-values/33031

aaronbell:
Closes #1071 

In order to make glyphsLib assign the correct glyph height to every glyph on conversion, it needs a more intelligent approach than simply copying the `sTypo` values. In this PR, I updated the logic to initially attempt to use data from a `BASE` table in the font .fea file to determine the correct height values. Barring that, evaluate if the sum of the sTypo metrics are larger than the font upm (which implies some variance from the 'standard' approach). And if all of the above is not usable, or False, then to fall back to the existing logic of using the sTypo metrics or master.ascender / master.descender values. 

Implementing this logic will help avoid the need for a post-production script to address problems in the `vmtx` and `vhea` tables. 

*Notes* 
1) Glyphs does not currently use this approach to determine glyph heights on output. However, I am proposing such a modification to it as well. 

2) This code is horribly un-optimized and could be greatly improved for speed. I am providing this PR primarily as a proof of concept of the approach, since plumbing it properly would be more involved.

### DIFF
--- a/Lib/glyphsLib/builder/glyph.py
+++ b/Lib/glyphsLib/builder/glyph.py
@@ -361,7 +361,7 @@ def to_ufo_glyph_height_and_vertical_origin(self, ufo_glyph, layer):
     # https://github.com/googlefonts/glyphsLib/issues/557#issuecomment-667074856
     assert self.is_vertical
 
-    ascender, descender = _get_typo_ascender_descender(layer.master)
+    ascender, descender = _get_ascender_descender(layer.master)
 
     if layer.vertWidth is not None:
         ufo_glyph.height = layer.vertWidth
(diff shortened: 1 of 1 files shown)


schriftgestalt:
I didn't look at the code, yet. But it doesn't sound right to use the Base table for the. I would rather define a custom parameter (or even a proper metrics setting) for this. Something like `defaultVertHeigth`

aaronbell:
@schriftgestalt there does need to be both a proper ascender and descender value set since that is used for both the advanceHeight and verOrigin entries, so do keep that in mind.

The BASE table is a useful tool in that it usually provides these values directly, or easily determinable as the values are set by the designer.

aaronbell:
It is also worth noting that even if a new field is added, it is helpful to have a backwards compatible methodology to determine this correctly.

schriftgestalt:
Then we add two values. 
And for fonts that do not have the values, can just as well be produced as is. And if someone is complaining, set the values properly instead of applying some guesswork.

PR #1084, opened 2025-06-10
Update requirements

khaledhosny:
It seems that the auto update service is no longer running.


khaledhosny:
I don’t know how to handle Python version-specific requirements with pip-compile.

PR #1085, opened 2025-06-10
[builder] Use variablePostscriptFontName from instances if present

khaledhosny:
This seems to be the new property specifically for fvar instance postscript name, so try it first:
https://handbook.glyphsapp.com/custom-parameter-descriptions/#custom-parameter/variablePostscriptFontName https://github.com/googlefonts/fontc/pull/1525


anthrotype:
Thanks Khaled. fontc already supports this https://github.com/googlefonts/fontc/pull/1428

PR #1093, opened 2025-07-21
[align_alternate_layers] must not skip non-export glyphs when collecting alternate layers

anthrotype:
I noticed what looks like a regression with glyphsLib v6.11.2, specifically after https://github.com/googlefonts/glyphsLib/pull/1092 got merged

the previous `find_component_use` code in `bracket_layers.py` was considering alternate layers inside glyphs marked as non-exportable whereas in the new align_alternate_layers.py that replaced it, I thought I might skip those but turns out it's possible for composite glyphs to reference component glyphs that are in turn marked as export=False and the code still needs to be able to align alternate layers definitions across these. At least that's my current understanding..

This is reproducible with the following ttx_diff.py command

```
python3 resources/scripts/ttx_diff.py 'https://github.com/googlefonts/Playfair?a49f9f9dc9#sources/Playfair.glyphspackage'
```

### DIFF
--- a/Lib/glyphsLib/builder/transformations/align_alternate_layers.py
+++ b/Lib/glyphsLib/builder/transformations/align_alternate_layers.py
@@ -48,7 +48,7 @@ def align_alternate_layers(font, glyph_data=None):
                         for layer in alternate_layers[master][comp.name]
                     }
                     if my_bracket_layers != components_bracket_layers:
-                        # Find what we need to add, and make them hashable
+                        # Find what we need to add
                         needed = components_bracket_layers - my_bracket_layers
                         if needed:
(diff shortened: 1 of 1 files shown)


anthrotype:
I also noticed another issue which actually predates my changes in #1092 and can be reproduced with the previous glyphsLib v6.11.1.

If you take Playfair.glyphspackage from https://github.com/googlefonts/Playfair, and convert to UFO `fontmake -g sources/Playfair.glyphspackage -o ufo` multiple times, the order of the rules in the generated Playfail.designspace is different each time, because the new bracket layers were appended by iterating over a set which is unsorted and affected by python randomized hash seed. So now I made the order deterministic.

Issue #860, opened 2023-02-16
Glyphs 3 localized font info are ignored


khaledhosny:
Glyphs 3 supports localizing many font info entries, but glyphsLib will export only one (English, default, or the first), also the `GSFont` APIs for accessing them are missing (e.g. there is `copyright` but not `copyrights`).

schriftgestalt:
The localized values are stored in the `.properties`: https://docu.glyphsapp.com/#GSFont.properties

schriftgestalt:
I just had a look at this. This is what I found:

The ufos/fontinfo can have `openTypeNameRecords`. But those are the masters and don’t need much names. 
The instances are stored in the .designspace. The default value are attributes in the `instance` elements and the localized values are child element?? https://fonttools.readthedocs.io/en/latest/designspaceLib/xml.html#id50

khaledhosny:
No idea to be honest. I’m personally more concerned about the missing classes APIs.

schriftgestalt:
> I’m personally more concerned about the missing classes APIs.

Can you explain?

khaledhosny:
> > I’m personally more concerned about the missing classes APIs.
> 
> Can you explain?

I mean the missing `GSFont` etc. API like `copyrights` and any other support for handling localized properties at the level of Python API.

schriftgestalt:
the `copyrights`… properties are only meant as a convenient method. For generating and round tripping purposes it is better to use the `.properties`.

khaledhosny:
I need to refresh mu memory, but I recall there were some inconsistencies in how to access localized properties in general, and the API looked incomplete (or incompatible with the Glyphs app Python API).

PR #1108, opened 2025-09-22
[builder] Support localized font properties

khaledhosny:
Convert them to raw `openTypeNameRecords`.

Fixes https://github.com/googlefonts/glyphsLib/issues/860

### DIFF
--- a/Lib/glyphsLib/builder/font.py
+++ b/Lib/glyphsLib/builder/font.py
@@ -89,8 +89,6 @@ PROPERTIES_FIELDS = (
 def fill_ufo_metadata(master, ufo):
     font = master.font
 
-    # "date" can be missing; Glyphs.app removes it on saving if it's empty:
-    # https://github.com/googlefonts/glyphsLib/issues/134
     for info_key, glyphs_key in INFO_FIELDS:
         value = getattr(font, glyphs_key)
         setattr(ufo.info, info_key, value)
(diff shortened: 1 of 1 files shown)


Lib/glyphsLib/builder/font.py
```diff
@@ -69,27 +70,58 @@ def to_ufo_font_attributes(self, family_name):
...
+)
+
+PROPERTIES_FIELDS = (
+    ("copyright", "copyrights", 0),
+    ("openTypeNameDesigner", "designers", 9),
+    ("openTypeNameDesignerURL", "designerURL", None),
+    ("openTypeNameManufacturer", "manufacturers", 8),
+    ("openTypeNameManufacturerURL", "manufacturerURL", None),
+    ("openTypeNameDescription", "descriptions", 10),
+    ("openTypeNameSampleText", "sampleTexts", 19),
```
anthrotype: @khaledhosny I was working on implementing this in fontc and noticed that this is missing `familyNames` which can also be localized. fontc happens to have one of its test fonts with a localized familyNames:

https://github.com/googlefonts/fontc/blob/9925c92ff9a3b253548f849aab0ddbe29735c5ee/resources/testdata/glyphs3/WghtVar.glyphs#L354-L360

Was this intentional to exclude it? Or the list was never meant to be exhaustive (best effort basis) and may be missing other localizable fields in addition to familyNames?

PR #1114, opened 2025-11-06
[tokens] Support a few more glyph predicates

khaledhosny:


Lib/glyphsLib/classes.py
```diff
@@ -4345,6 +4345,53 @@ def vertWidthMetricsKey(self):
     def vertWidthMetricsKey(self, value):
         self.metricVertWidth = value
 
+    # The following methods are there mainly to support there use in predicate
```
anthrotype:
```suggestion
    # The following methods are there mainly to support their use in predicate
```

khaledhosny:
> while looking at your PR I noticed what looks like a preexisting bug in tokens.py where it attempts to call a method called `_parse_aggregation` but I can't find it defined anywhere (it will probably trigger AttributeError if that line is ever hit):
> 
> https://github.com/googlefonts/glyphsLib/blob/ffa8e042c53e04fa51f979cf33d123f27f1eeb20/Lib/glyphsLib/builder/tokens.py#L185-L192
> 
> ```shell
> $ rg _parse_aggregation
> Lib/glyphsLib/builder/tokens.py
> 190:            expected = self._parse_aggregation()
> ```

Yes. I found about this a while ago and forgot to report it. The `in` and `between` operators don’t actually work at all, ad it is not only this non-existing method.

PR #1117, opened 2025-11-15
Fix duplicate postscript names for color layer glyphs

anthrotype:
Color layer glyphs (e.g., "Acaron.color0", "Acaron.color1") were getting duplicate postscript names instead of unique ones with the .colorN suffix preserved.

The bug occurred because when creating color layer glyphs, to_ufo_glyph() was copying unicode values from the parent glyph. This caused glyphdata.get_glyph() to return the base glyph's production name via unicode lookup, losing the .colorN suffix:

Before:
```
  Acaron          -> uni01CD
  Acaron.color0   -> uni01CD
  Acaron.color1   -> uni01CD
```

After:
```
  Acaron          -> uni01CD
  Acaron.color0   -> uni01CD.color0
  Acaron.color1   -> uni01CD.color1
```

### DIFF
--- a/Lib/glyphsLib/builder/color_layers.py
+++ b/Lib/glyphsLib/builder/color_layers.py
@@ -36,10 +36,7 @@ def _to_ufo_color_palette_layers(builder, master, layerMapping):
                 layerGlyphName = f"{glyph.name}.color{i}"
                 ufo_layer = builder.to_ufo_layer(glyph, masterLayer)
                 ufo_glyph = ufo_layer.newGlyph(layerGlyphName)
-                builder.to_ufo_glyph(ufo_glyph, layer, glyph)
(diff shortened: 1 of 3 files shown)


anthrotype:
you may argue that postscript names are useless in particular for these glyphs which are never going to appear in a run of text, but are simply drawn as COLR layers of other color base glyphs... But the problem is, only _some_ of these color layer glyphs would get the duplicate (un-suffixed) postscript name (the ones with a unicode), whereas others like "caroncomb.alt" which doesn't have a unicode to begin with, would get the suffixed postscript name ""caroncomb.alt.color0" etc.

This is inconsistent and hard/ugly to replicate in fontc. We can either make sure that all color layer glyphs keep their .colorN suffix, or we deliberately strip it from all. I chose the former, because it appeared like an oversight rather than intentional.

PR #1119, opened 2025-11-28
Skip irrelevant layers when converting brackets

cmyr:
This is a funny little bug: when converting brackets to designspace rules, we reuse the 'overlayFeatureVariations' function from varLib to create a set of minimal designspace regions to represent the set of of possible substitutions.

overlayFeatureVariations has logic for clearing out axes that are irrelevant (that is, where a rule would match across an entire axis) but this logic expects that the regions are expressed in normalized coordinates. When it is called form glyphsLib, the regions are expressed in userspace coordinates, and this logic is never triggered.

To work around this, we manually skip axis rules that match the entire axis before calling overlayFeatureVariations.

This also adds a new test file and test for format 3 bracket layers, since these multi-axis bracket rules can't be expressed in format2.


anthrotype (approved):
makes sense. And thanks for the test!

PR #1121, opened 2025-12-22
Tell ufo2ft mark feature writer to use `mark` prefix for mark classes

khaledhosny:
ufo2ft default to `MC` prefix for mark classes, but Glyphs uses `mark`, since feature code can reference auto generated code, we want to match Glyphs here.

Depends on https://github.com/googlefonts/ufo2ft/pull/965

### DIFF
--- a/Lib/glyphsLib/builder/constants.py
+++ b/Lib/glyphsLib/builder/constants.py
@@ -208,7 +208,10 @@ UFO2FT_META_TABLE_KEY = PUBLIC_PREFIX + "openTypeMeta"
 DEFAULT_FEATURE_WRITERS = [
     {"class": "CursFeatureWriter"},
     {"class": "KernFeatureWriter"},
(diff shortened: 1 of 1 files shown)


khaledhosny:
Not needed now that ufo2ft will change the default prefix (https://github.com/googlefonts/ufo2ft/pull/974).

schriftgestalt:
> So I don’t think something like this is going to work with current fontc architecture as there are no generated mark class names to reference. 

The compiler could add the classes to the internal state to allow make them accessible to fea code.

khaledhosny:
> > So I don’t think something like this is going to work with current fontc architecture as there are no generated mark class names to reference.
> 
> The compiler could add the classes to the internal state to allow make them accessible to fea code.

I have an idea that I’ll try later (extending fea-rs FeatureProvider to allow querying the caller for missing class names, then other parts of fontc that call fea-rs would provide the class definition on the fly).

PR #1125, opened 2026-01-13
Refresh requirements.txt and requirements-dev.txt

khaledhosny:
Update requirements so that installing in current Python versions does not fail while building binary packages from source.


anthrotype (approved):

khaledhosny:
> > fail while building binary packages
> 
> which ones, out of curiosity?

openstep-plist

PR #1127, opened 2026-01-13
Make serializing GSInstance.type more future-proof

khaledhosny:
There is currently only two types SINGLE and VARIABLE, but don’t hard-code VARIABLE just in case more types are added in the future.


khaledhosny:
This fixes a thiko from #1126

PR #1129, opened 2026-01-19
[tokens] Support tags in tokens

khaledhosny:


khaledhosny:
https://github.com/googlefonts/glyphsLib/blob/7ca929b15af6cb67acd77039fea0a04854e3dc71/Lib/glyphsLib/builder/tokens.py#L198-L199

PR #1137, opened 2026-02-16
Fix ligature anchor numbering on synthesized bracket layers

### LINKED ISSUE #1136: Anchor propagation misses bracket-layer ligature variants
## Summary

When a ligature composite's component has bracket layers, the compiled font contains synthesized bracket variant glyphs (e.g. `f_h.liga.BRACKET.varAlt01`). These variants don't receive propagated anchors and therefore have no MarkToLig attachment rules.

## Details

In Eczar.glyphs, `f_h.liga` is a composite of `f` + `h`. The glyph `h` has bracket layers `[160]` (weight axis). Both fontmake and fontc produce the bracket variant glyph `f_h.liga.BRACKET.varAlt01` in the compiled font, but only fontc gives it MarkToLig rules.

The reason: `propagate_all_anchors` runs during `preflight_glyphs` on the GSFont, where bracket layers are still layers on the original glyph. `f_h.liga` itself has no bracket layers — only its component `h` does. So propagation never creates anchors for a bracket variant of the ligature. The bracket variant glyph is synthesized later during UFO→binary compilation, after propagation has already run.

fontc handles this because it synthesizes bracket variant glyphs early (during source reading), before anchor propagation runs, so `f_h.liga.BRACKET.varAlt01` exists as a first-class glyph and naturally participates in propagation.

## Reproduction

```
Eczar.glyphs from https://github.com/rosettatype/eczar (commit f248ec9c0c)
```

Compare `f_h.liga.BRACKET.varAlt01` MarkToLig rules in fontc vs fontmake output.

## Impact

Low — bracket ligature variants missing mark attachment is unlikely to cause visible issues in practice, but it's technically incorrect: those glyphs can appear in rendered text (via GSUB ConditionSet) and should support mark positioning.

anthrotype:
When align_alternate_layers synthesizes a bracket layer on a ligature composite (because its component has bracket layers), the synthesized layer has a different layerId than the component's bracket layer.

The `base_glyph_counts` lookup in `anchors_traversing_components` used (component_name, layer.layerId) as key, which missed the component's entry stored under a different bracket layerId. This caused the lookup to return 0, skipping ligature anchor renumbering entirely and producing "top" instead of "top_1" on bracket layers.

The fix is to fall back to associated master's layer when the exact layerId lookup misses.

Fixes https://github.com/googlefonts/glyphsLib/issues/1136

### DIFF
--- a/Lib/glyphsLib/builder/transformations/propagate_anchors.py
+++ b/Lib/glyphsLib/builder/transformations/propagate_anchors.py
@@ -182,6 +182,23 @@ def _interpolate_smart_component_anchors(
             anchor.position = Point(new_coord[0], new_coord[1])
 
 
+def _get_base_glyph_count(
+    base_glyph_counts: dict[(str, str), int],
+    component_name: str,
+    layer: GSLayer,
+) -> int:
+    """Look up the base glyph count for a component, handling bracket layers.
+
+    Synthesized bracket layers on composites have different layerIds than their
+    component's bracket layers. When the exact (name, layerId) lookup misses,
+    fall back to the associated master's layer.
+    """
+    count = base_glyph_counts.get((component_name, layer.layerId))
+    if count is not None:
+        return count
+    return base_glyph_counts.get((component_name, layer.associatedMasterId), 0)
+
+
 def anchors_traversing_components(
     glyph: GSGlyph,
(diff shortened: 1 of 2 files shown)


cmyr (approved):
looks good!

PR #1150, opened 2026-04-28
Sync GlyphData.xml to latest (b2a1b29)

justvanrossum:


anthrotype (approved):

anthrotype:
the test expectations may also need to be adjusted

justvanrossum:
The test failure is because of this commit: https://github.com/schriftgestalt/GlyphsInfo/commit/619cbd90507743edc34dfa92e063651c86759f2e

[This comment](https://github.com/googlefonts/glyphsLib/commit/5e93bc36930885a3555bc820fab839b055382d50#commitcomment-157764375) suggests that the new result (`uni30213021.paren` instead of `one_one.paren`) would be correct.

justvanrossum:
But then again, [this comment](https://github.com/googlefonts/glyphsLib/commit/5e93bc36930885a3555bc820fab839b055382d50#commitcomment-157837868) suggests it should be `uni32203220`, not `uni32203220.paren`.

@schriftgestalt please advise.

tests/glyphdata_test.py
```diff
@@ -404,7 +404,7 @@ def test_glyphdata_no_duplicates(self):
     "ya_uMatra-tamil": "uni0BAF0BC1",
     "ya_uuMatra-tamil": "uni0BAF0BC2",
     "yoYing_underscore-thai": "uni0E0D005F",  # uni0E0D_uni005F
-    "one_one-han.paren": "one_one.paren",  # not 'one_onehan.paren'
+    "one_one-han.paren": "uni30213021.paren",  # not 'one_onehan.paren'
```
anthrotype, line 407:
```
"one_one-han.paren": "uni30213021.paren"
```

this looks correct to me. It's a ligature of two one-han.paren, each ligature component is mapped to 0x3021. It's definitely more correct than one_one.paren (which would be interpreted as two latin 1 digits).

schriftgestalt: It should be "uni32203220" but "uni32203220.paren" is acceptable.

PR #1151, opened 2026-04-30
Fix encoding problem with reading .glyphspackage .glyph file

justvanrossum:
Additionally adjusts some test code.

This fixes a similar bug as https://github.com/fontra/fontra-glyphs/issues/140. Occasional glyphsLib errors were mentioned there by @Eigi, so I went for a hunt and found one.

### DIFF
--- a/Lib/glyphsLib/parser.py
+++ b/Lib/glyphsLib/parser.py
@@ -123,7 +123,7 @@ def load_glyphspackage(package_dir):
 
     data["glyphs"] = []
     for glyphfile in (package / "glyphs").glob("*.glyph"):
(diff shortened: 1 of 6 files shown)


khaledhosny (approved):

justvanrossum:
Quick note about the "일" character added to the annotation test: Windows by default opens files using the cp1252 encoding. This will silently accept many utf-8 sequences. But not "일", as its utf-8 representation contains a code that is undefined in cp1252, which causes an error when decoding. While there are likey many other characters with this property, this is the one I bumped into in the Fontra bug.

Issue #800, opened 2022-07-31
Support (or don't) Glyphs-specific variable feature syntax


simoncozens:
Glyphs has a different variable feature syntax to Adobe/fonttools/FontCreator. Instead of:

```
conditionset bold {
    wght 140 180;
} bold;

variation rlig bold {
   sub dollar by dollar.bold;
   sub cent by cent.bold;
} rlig;
```

Glyphs produces

```
feature rlig {
#ifdef VARIABLE

condition 140 < wght < 180;
sub dollar by dollar.bold;
sub cent by cent.bold;

#endif
} rlig;
```

Do we want to support this? I think probably not. Glyphs should do the right thing here.

anthrotype:
would be nice to be able to automatically translate from one to the other form when converting glyphs=>UFOs, but if it's too much trouble we shouldn't bother.
But I also think Glyphs.app should be updated to support the official form (if it doesn't work already, does it?)

schriftgestalt:
It would be nice if there where a spec that all parties had discussed and had agreed to. Now we have a suggestion that some parties just found good enough and went ahead and implemented it.

simoncozens:
Welcome to OpenType.

vv-monsalve:
> would be nice to be able to automatically translate from one to the other form when converting glyphs=>UFOs, but if it's too much trouble we shouldn't bother.

This would be nice, indeed. Would it be too much trouble?

Glyphs is suggesting, so I can see more cases coming (I have on at the moment).

HugoJourdan:
> > would be nice to be able to automatically translate from one to the other form when converting glyphs=>UFOs, but if it's too much trouble we shouldn't bother.

+1

simoncozens:
The Adobe/fontTools variable fea syntax should work already; the Glyphs 3 variable fea syntax should not. What's happening in your case, Hugo?

BaGsn:
I tried Adobe syntax and got the same error as with Glyphs one, but with _conditionset_ instead of _condition_:
`Expected glyph class definition or statement: got NAME condition`
`Expected glyph class definition or statement: got NAME conditionset`

jpt:
> Do we want to support this? I think probably not. Glyphs should do the right thing here.

I for one think glyphsLib should support this. This issue has been open long enough that there are probably a bunch of Glyphs files in the wild that already do it the Glyphs way

jpt:
@anthrotype 
[...]

PR #1159, opened 2026-07-28
Convert variable feature code when building UFOs from Glyphs

khaledhosny:
Convert from Glyphs-specific conditional features and variable GPOS to the corresponding feaLib syntax:
https://handbook.glyphsapp.com/layout/conditional-feature-code/ https://handbook.glyphsapp.com/layout/variable-gpos/

Feature code already in feaLib syntax is kept unchanged.

The #ifndef VARIABLE blocks are dropped. We don’t have a way to tell if the font is going to be built as variable or static font, so we assume variable build.


justvanrossum:
Does this close #800?

khaledhosny:
> Does this close #800?

Yes.

PR #1160, opened 2026-07-28
Support Glyphs 3 slant transforms

### LINKED ISSUE #1047: Issue with skew attribute in component transform in `Glyphs3-merge` branch
Consider the following code

```python
import pprint

from fontTools.pens.recordingPen import RecordingPen
from glyphsLib import GSFont

if __name__ == "__main__":
    gs_font = GSFont("example.glyphs")
    layer = gs_font.glyphs["bitcoin"].layers[0]
    recording_pen = RecordingPen()
    layer.draw(recording_pen)
    pprint.pp(recording_pen.value)
```

When running the code from the `main` branch, the output is:

```
[('moveTo', ((249, 650),)),
 ('lineTo', ((297, 650),)),
 ('lineTo', ((324, 800),)),
 ('lineTo', ((276, 800),)),
 ('closePath', ()),
 ('moveTo', ((364, 650),)),
 ('lineTo', ((412, 650),)),
 ('lineTo', ((439, 800),)),
 ('lineTo', ((391, 800),)),
 ('closePath', ()),
 ('moveTo', ((115, -110),)),
 ('lineTo', ((163, -110),)),
 ('lineTo', ((191, 50),)),
 ('lineTo', ((143, 50),)),
 ('closePath', ()),
 ('moveTo', ((230, -110),)),
 ('lineTo', ((278, -110),)),
 ('lineTo', ((307, 50),)),
 ('lineTo', ((259, 50),)),
 ('closePath', ()),
 ('addComponent', ('B', <affine transformation 1 0 0.0 0.94 5 21>))]
```

Instead, when running the code from `Glyphs3-merge`, the output is:

```
[('moveTo', ((249, 650),)),
 ('lineTo', ((297, 650),)),
 ('lineTo', ((324, 800),)),
 ('lineTo', ((276, 800),)),
 ('closePath', ()),
 ('moveTo', ((364, 650),)),
 ('lineTo', ((412, 650),)),
 ('lineTo', ((439, 800),)),
 ('lineTo', ((391, 800),)),
 ('closePath', ()),
 ('moveTo', ((115, -110),)),
 ('lineTo', ((163, -110),)),
 ('lineTo', ((191, 50),)),
 ('lineTo', ((143, 50),)),
 ('closePath', ()),
 ('moveTo', ((230, -110),)),
 ('lineTo', ((278, -110),)),
 ('lineTo', ((307, 50),)),
 ('lineTo', ((259, 50),)),
 ('closePath', ()),
 ('addComponent',
  ('B', <affine transformation 1.0 0.0 -0.8422883804630793 0.94 5.0 21.0>))]
```

Attached you can find the `.glyphs` file.

[example.glyphs.zip](https://github.com/user-attachments/files/17736484/example.glyphs.zip)

cc @schriftgestalt

anthrotype:
Fixes #1047

Glyphs 3 does not store component transforms as a single affine matrix. It stores `pos`, `scale`, `angle` and `slant` as separate attributes. glyphsLib previously collapsed the attributes it understood into an internal six-value `Transform`, then attempted to reconstruct the split attributes when writing Glyphs 3 again.

That conversion is not lossless. Affine decomposition is non-unique, so the original scale, rotation and slant values cannot in general be recovered from the matrix.

In addition, glyphsLib did not parse `slant` at all. A slanted component was therefore emitted unsheared when building UFOs, and merely opening and saving a Glyphs 3 source through glyphsLib deleted its `slant` attribute.

This PR keeps the split attributes authoritative for objects originating in Glyphs 3 and composes their affine matrix lazily only when a matrix consumer needs it. Their original values, including nonzero vertical slant, therefore round-trip verbatim. Objects originating from Glyphs 2, UFOs or explicit `.transform` assignment instead treat the matrix form as authoritative. Decomposition into split attributes occurs only when such an object must be written as Glyphs 3, where those attributes are unavoidable.

For matrix-authoritative objects, writing the linear matrix as `M = K × R × S`, where `K` contains the two shear factors, means that removing a candidate `K` must leave a matrix whose columns are orthogonal: rotation followed by independent X/Y scaling. That condition gives a conic equation in the two shear factors.

The implementation generates horizontal-only, vertical-only and closed-form rotation-free candidates. The rotation-free candidate attributes both off-diagonal terms to slant and therefore needs no rotation. Every candidate is recomposed and checked against the original matrix to reject numerical false solutions. The implementation prefers any valid candidate with `slantY = 0`, matching what the Glyphs UI can express; otherwise it selects the valid candidate requiring the least total shear.

Singular matrices have no unique decomposition. We preserve their translation and fall back to the existing legacy scale/rotation approximation with zero slant, allowing collapsed components to remain readable and writable without crashing.

The public `transform` getter now returns a built-in six-tuple instead of glyphsLib’s mutable `Transform`, exactly matching Glyphs.app’s Python scripting API. Builders and pens consequently receive immutable values rather than internal cache storage.

Compatibility note: code that mutates the returned object in place or uses `Transform`-specific attributes must instead assign a complete transform. Normal sequence access and whole-property assignment remain supported.

`GSBackgroundImage` now uses the same transform model. This also fixes two image-related bugs: stale `scale` and `rotation` values after assigning an image transform, and a format 2 image’s linear transform disappearing when written as format 3.

I have never seen a slanted background image in the wild, so I checked this directly: Glyphs 3.4.1 reads, applies and round-trips background-image slant.

Finally, I benchmarked this on both synthetic and real sources and observed no end-to-end performance regression in the tests I ran. On a deliberately component-only 2,000-component source, parse plus `to_ufos` was about 28% faster than main for ordinary components and 20% faster for slanted components, largely because fields are recorded cheaply and matrices composed lazily.

A `to_ufos`-only torture test using 10,000 cloned dirty components makes that deferred cache fill visible in the builder timer, but moves work from parsing rather than adding it end to end. On BigShoulders, with 9,252 components across six masters, tuple-returning boundaries showed no measurable `to_ufos` regression: 1.204 seconds versus 1.223 seconds using raw cached matrices, with the apparent improvement attributable to measurement noise.

I am also working on the equivalent fontc change, which I plan to push in the coming days while working intermittently from holiday.

### DIFF
--- a/Lib/glyphsLib/builder/smart_components.py
+++ b/Lib/glyphsLib/builder/smart_components.py
@@ -202,12 +202,13 @@ def instantiate_smart_component(self, layer, component, pen):
 
     # Don't forget that the GSComponent might also be transformed, so
     # we need to apply that transformation to the new layer as well
-    if component.transform:
+    transform = component._transformMatrix()
+    if transform:
         # We must reverse path direction for flipped components
         # https://github.com/googlefonts/glyphsLib/issues/882
-        should_reverse = component.transform.determinant() < 0
+        should_reverse = transform.determinant() < 0
         for p in new_layer.paths:
-            p.applyTransform(component.transform)
+            p.applyTransform(transform)
             if should_reverse:
                 p.reverse()
 
--- a/Lib/glyphsLib/classes.py
+++ b/Lib/glyphsLib/classes.py
@@ -25,6 +25,7 @@ from enum import IntEnum
 from io import StringIO
 
 # renamed to avoid shadowing glyphsLib.types.Transform imported further below
+from fontTools.misc.bezierTools import solveQuadratic
 from fontTools.misc.transform import Identity, Transform as Affine
 from fontTools.pens.basePen import AbstractPen
 from fontTools.pens.pointPen import (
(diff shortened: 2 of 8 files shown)


anthrotype:
Here's some extra motivation for this fix. It turns out the `slant` attribute is not exotic. About 20 families with .glyphs sources among the fontc crater corpus use slant on exported layers (backgrounds and non-exported glyphs excluded), and the respective shipped binaries on google/fonts are missing the intended shear.

The image below only compares some of the worst offenders: current google/fonts binaries (left) vs the same sources built by fontmake on this branch (right), both rendered with hb-view.

slant-damage-sheet-v3-fontmake

[Teachers](https://github.com/google/fonts/blob/e4572de925a4c3be12f1f9983ee0adbe1eb6e9fe/ofl/teachers/Teachers%5Bwght%5D.ttf) draws the l/Lslash stroke as a horizontal bar instead of a 30-37 degree diagonal; [Figtree Italic](https://github.com/google/fonts/blob/e4572de925a4c3be12f1f9983ee0adbe1eb6e9fe/ofl/figtree/Figtree-Italic%5Bwght%5D.ttf)'s grave doesn't lean with the italic; [Elms Sans Italic](https://github.com/google/fonts/blob/e4572de925a4c3be12f1f9983ee0adbe1eb6e9fe/ofl/elmssans/ElmsSans-Italic%5Bwght%5D.ttf)'s greater and lessequal are mirrored horizontally from less/greaterequal without the compensating slant, so they lean backwards against the italic; similar for [Geist Mono Italic](https://github.com/google/fonts/blob/e4572de925a4c3be12f1f9983ee0adbe1eb6e9fe/ofl/geistmono/GeistMono-Italic%5Bwght%5D.ttf)'s second, [Miranda Sans Italic](https://github.com/google/fonts/blob/e4572de925a4c3be12f1f9983ee0adbe1eb6e9fe/ofl/mirandasans/MirandaSans-Italic%5Bwght%5D.ttf)'s spacing accents, and [Amarna Italic](https://github.com/google/fonts/blob/e4572de925a4c3be12f1f9983ee0adbe1eb6e9fe/ofl/amarna/Amarna-Italic%5Bwght%5D.ttf)'s backslash. I also visually spot-checked the "intended" shapes against Glyphs.app's own rendering.

fontc's implementation of the slant attribute (googlefonts/fontc#2071) produces matching output for all six families.

khaledhosny (approved):

PR #1162, opened 2026-09-02
Expand Glyphs multiple languages syntax when building UFOs

### LINKED ISSUE #1109: Concurrent language tags aren't correctly recognised
I'm not entirely sure where exactly to post this issue, whether in fontmake or glyphsLib. I hope here is correct.

In Glyphs, multiple language tags for a lookup can be written concurrently:
`language ROM MOL;`

However, this leads to an error when using fontmake to compile a ufo:
`Compiling UFO failed: <features>:110:14: Expected ';'`

The UFO compiler expects one language tag per lookup, it appears. Glyphs allows concurrent language tags like `language AZE CRT KAZ TAT TRK;`.

How can this be amended, and where?

tomekthewo:
Fixes #1109.

Glyphs lets a single `language` statement carry several tags and applies everything that follows to each of them:

```fea
language AZE CRT KAZ TAT TRK;
lookup idotaccent {
    sub i by idotaccent;
} idotaccent;
```

The FEA spec allows exactly one tag, so feaLib stops at the second one and the UFOs glyphsLib writes cannot be compiled at all — `<features>:N:14: Expected ';'`, column 14 being where `CRT` starts. The syntax is documented in the [Glyphs Handbook](https://handbook.glyphsapp.com/layout/multiple-languages-syntax/) and written up as a spec proposal in [adobe-type-tools/feature_file_workshops#8](https://github.com/adobe-type-tools/feature_file_workshops/pull/8). Glyphs 4 emits it from *automatic* feature generation, so a source that built yesterday stops building after a save.

This expands the shorthand into one statement per tag in `_to_ufo_features`. What reaches the UFO is then plain spec-compliant FEA that any compiler reads — feaLib, fea-rs and makeotf alike; expanding downstream would leave the UFO itself non-portable.

The scope is emitted once for the first tag and replayed for every remaining one. Named lookups cannot simply be duplicated — that would redefine them — so the replays reference the lookup defined under the first tag:

```fea
language AZE;
lookup idotaccent {
    sub i by idotaccent;
} idotaccent;
language CRT;
lookup idotaccent;
```

Glyph class and `markClass` definitions are likewise emitted only once; bare rules and `lookupflag`s are repeated verbatim, which is what the shorthand means. Trailing keywords (`exclude_dflt` and friends) are kept on every tag. A statement that is already spec-compliant, or that cannot be classified with confidence, is left untouched for feaLib to report rather than guessed at. So is a scope containing an `include()`, whose contents are not visible and therefore cannot be replayed; a statement listing `dflt` beside other tags, which has to be specified alone and which Glyphs rejects as well; one sitting outside any block, where FEA does not allow `language` at all; and one whose scope holds a statement with no closing semicolon.

**Built on feaLib's `Lexer`.** The scan walks the token stream, so a statement ends at the first `;` at brace depth zero wherever it sits on the page, and a brace or semicolon inside a comment or a string is never taken for code. What is left is classifying five statement heads (`lookup NAME {`, `@Class =`, `markClass`, `include(`, everything else): no glyph classes to resolve, no lookup bodies to understand, no name resolution. What is emitted for the first tag is sliced out of the original by offset, so comments, indentation and blank lines survive untouched; the replays are statements only, so a comment inside the scope is not duplicated. A `\blanguage\b` guard keeps the lexer off feature code that cannot contain the shorthand.

**It reproduces what Glyphs 3 wrote.** Taking a `locl` feature from a production source in both its Glyphs 3 and its Glyphs 4 form — the automatic code nobody edited, before and after a save in Glyphs 4 — expanding the shorthand gives back the Glyphs 3 text byte for byte, tabs and blank lines included. Compiled with feaLib, the two produce an identical GSUB table. A second test does the same for the shape where the language statements sit *inside* the lookup block; there the single-tag form is what Glyphs 3 wrote for another production source and the shorthand is that form collapsed by hand, since Glyphs 4 has not been seen to emit it — it pins down that the scope ends at the block's own brace, with the rules repeated rather than referenced.

**Ordering.** The expansion runs after `VariableFeatureConverter`, so `#ifndef VARIABLE` blocks are already resolved. The markers are comments and therefore not statements: expanding first would place the repeats *inside* the block, and the strip that follows would then take the second `language` statement with it. `test_conditional_block_is_resolved_before_the_expansion` pins this down.

**One-way.** The expansion is not reversed on the way back, so `.glyphs` → UFO → `.glyphs` yields the expanded form rather than the original statement. `test_expansion_does_not_round_trip` pins this down. UFO → `.glyphs` → UFO is unaffected, since the original feature text is recovered from `ORIGINAL_FEATURE_CODE_KEY`. Happy to restore the shorthand in `to_glyphs_features` instead if you would rather keep it lossless.

39 tests in `tests/builder/multi_language_test.py`, including the four counterexamples from the review, the two real-source shapes above, every line ending feaLib recognises, each shape left untouched, idempotence, two through `to_ufos` and one through the round trip. Two are invariants over a corpus of awkward sources rather than one example each: the output is either identical to the input or it compiles, and line endings do not change which rules end up under which tag. Full suite: 1376 passed, 64 skipped, 4 xfailed. `black --check` and `flake8` over `Lib tests` are clean.

### DIFF
--- a/tests/builder/multi_language_test.py
+++ b/tests/builder/multi_language_test.py
@@ -18,7 +18,7 @@ from textwrap import dedent
 
 from fontTools.feaLib.parser import Parser
 
-from glyphsLib import classes, to_ufos
+from glyphsLib import classes, to_glyphs, to_ufos
 from glyphsLib.builder.multi_language import expand_multi_language_statements
 
 GLYPH_NAMES = ["i", "idotaccent", "Scedilla", "Scommaaccent", "Lcommaaccent"]
@@ -111,7 +111,7 @@ def test_unparseable_statement_is_untouched():
     assert expand_multi_language_statements(original) == original
 
 
-def test_to_ufos_expands_feature_code(ufo_module):
+def make_font():
     font = classes.GSFont()
     font.masters.append(classes.GSFontMaster())
     for name in ("i", "idotaccent"):
@@ -134,8 +134,11 @@ def test_to_ufos_expands_feature_code(ufo_module):
         } idotaccent;
     """)
     font.features.append(feature)
+    return font
 
-    (ufo,) = to_ufos(font, ufo_module=ufo_module)
+
+def test_to_ufos_expands_feature_code(ufo_module):
+    (ufo,) = to_ufos(make_font(), ufo_module=ufo_module)
 
     lines = [line.strip() for line in ufo.features.text.splitlines()]
     assert "language AZE TRK;" not in lines
@@ -143,3 +146,19 @@ def test_to_ufos_expands_feature_code(ufo_module):
     assert "language TRK;" in lines
     assert lines.count("lookup idotaccent;") == 1
     Parser(io.StringIO(ufo.features.text), glyphNames=GLYPH_NAMES).parse()
+
+
+def test_expansion_does_not_round_trip(ufo_module):
+    """The expansion is one-way: the shorthand is not restored.
+
+    Going back to Glyphs yields the expanded form rather than the original
+    statement. UFO -> Glyphs -> UFO is unaffected, because the original
+    feature text is recovered from ORIGINAL_FEATURE_CODE_KEY.
+    """
+    (ufo,) = to_ufos(make_font(), ufo_module=ufo_module)
+
+    (feature,) = to_glyphs([ufo]).features
+    lines = [line.strip() for line in feature.code.splitlines()]
+    assert "language AZE TRK;" not in lines
+    assert "language AZE;" in lines
+    assert "language TRK;" in lines
--- a/tests/builder/multi_language_test.py
+++ b/tests/builder/multi_language_test.py
@@ -23,128 +23,99 @@ from glyphsLib.builder.multi_language import expand_multi_language_statements
 
 GLYPH_NAMES = ["i", "idotaccent", "Scedilla", "Scommaaccent", "Lcommaaccent"]
 
+LANGUAGE_SYSTEMS = "languagesystem DFLT dflt;\nlanguagesystem latn dflt;\n" + "".join(
+    f"languagesystem latn {tag};\n"
+    for tag in ("AZE", "CRT", "KAZ", "TAT", "TRK", "ROM", "MOL")
+)
 
-def parse(fea, glyph_names=GLYPH_NAMES):
+
+def parse(fea):
     """Parse the feature text, raising on anything feaLib does not accept."""
-    header = "languagesystem DFLT dflt;\nlanguagesystem latn dflt;\n"
-    header += "".join(
-        f"languagesystem latn {tag};\n"
-        for tag in ("AZE", "CRT", "KAZ", "TAT", "TRK", "ROM", "MOL")
-    )
-    return Parser(io.StringIO(header + fea), glyphNames=glyph_names).parse()
+    feature_file = io.StringIO(LANGUAGE_SYSTEMS + fea)
+    return Parser(feature_file, glyphNames=GLYPH_NAMES).parse()
 
 
-def expand(fea):
-    return expand_multi_language_statements(dedent(fea))
-
-
-def stripped(fea):
-    return [line.strip() for line in fea.splitlines()]
-
-
-def test_named_lookup_is_defined_once_and_referenced():
-    fea = expand("""\
-        feature locl {
-        script latn;
-
-        language AZE CRT KAZ TAT TRK;
-        lookup idotaccent {
-            sub i by idotaccent;
-        } idotaccent;
-
-        } locl;
-        """)
-
-    lines = stripped(fea)
-    # Defined under the first tag only ...
-    assert lines.count("lookup idotaccent {") == 1
-    # ... and referenced under each of the remaining four.
-    assert lines.count("lookup idotaccent;") == 4
-    for tag in ("AZE", "CRT", "KAZ", "TAT", "TRK"):
-        assert f"language {tag};" in lines
+def check(source, expected):
+    """Expand the source, assert the whole output, then parse it."""
+    fea = expand_multi_language_statements(dedent(source))
+    assert fea == dedent(expected)
     parse(fea)
 
 
 def test_bare_rules_are_repeated():
-    fea = expand("""\
+    check(
+        """\
         feature locl {
         script latn;
         language ROM MOL;
         sub Scedilla by Scommaaccent;
         } locl;
-        """)
-
-    lines = stripped(fea)
-    assert lines.count("sub Scedilla by Scommaaccent;") == 2
-    assert lines.count("language ROM;") == 1
-    assert lines.count("language MOL;") == 1
-    parse(fea)
-
-
-def test_glyph_class_is_not_redefined():
-    fea = expand("""\
+        """,
+        """\
         feature locl {
-        language ROM MOL;
-        @Cedillas = [Scedilla];
-        sub @Cedillas by Scommaaccent;
+        script latn;
+        language ROM;
+        sub Scedilla by Scommaaccent;
+        language MOL;
+        sub Scedilla by Scommaaccent;
         } locl;
-        """)
-
-    lines = stripped(fea)
-    assert lines.count("@Cedillas = [Scedilla];") == 1
-    assert lines.count("sub @Cedillas by Scommaaccent;") == 2
-    parse(fea)
+        """,
+    )
 
 
-def test_multi_line_glyph_class_is_dropped_whole():
-    fea = expand("""\
+def test_named_lookup_is_defined_once_and_referenced():
+    check(
+        """\
         feature locl {
-        language ROM MOL;
-        @Ced = [Scedilla
-                Lcommaaccent];
-        sub @Ced by Scommaaccent;
+        script latn;
+        language AZE CRT KAZ;
+        lookup idotaccent {
+            sub i by idotaccent;
+        } idotaccent;
         } locl;
-        """)
-
-    lines = stripped(fea)
-    assert lines.count("Lcommaaccent];") == 1
-    assert lines.count("sub @Ced by Scommaaccent;") == 2
-    parse(fea)
-
-
-def test_mark_class_is_not_redefined():
-    fea = expand("""\
-        feature test {
-        language ROM MOL;
-        markClass [Scedilla] <anchor 0 0> @MC;
-        pos base i <anchor 0 0> mark @MC;
-        } test;
-        """)
-
-    lines = stripped(fea)
-    assert lines.count("markClass [Scedilla] <anchor 0 0> @MC;") == 1
-    assert lines.count("pos base i <anchor 0 0> mark @MC;") == 2
-    parse(fea)
+        """,
+        """\
+        feature locl {
+        script latn;
+        language AZE;
+        lookup idotaccent {
+            sub i by idotaccent;
+        } idotaccent;
+        language CRT;
+        lookup idotaccent;
+        language KAZ;
+        lookup idotaccent;
+        } locl;
+        """,
+    )
 
 
 def test_lookup_with_use_extension_is_referenced():
-    fea = expand("""\
+    check(
+        """\
         feature locl {
         language AZE CRT;
         lookup idot useExtension {
             sub i by idotaccent;
         } idot;
         } locl;
-        """)
-
-    lines = stripped(fea)
-    assert lines.count("lookup idot useExtension {") == 1
-    assert lines.count("lookup idot;") == 1
-    parse(fea)
+        """,
+        """\
+        feature locl {
+        language AZE;
+        lookup idot useExtension {
+            sub i by idotaccent;
+        } idot;
+        language CRT;
+        lookup idot;
+        } locl;
+        """,
+    )
 
 
 def test_lookup_with_brace_on_the_next_line_is_referenced():
-    fea = expand("""\
+    check(
+        """\
         feature locl {
         language AZE CRT;
         lookup idot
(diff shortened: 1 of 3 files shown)


tests/builder/multi_language_test.py
```diff
@@ -0,0 +1,297 @@
...
+        script latn;
+        language ROM MOL;
+        sub Scedilla by Scommaaccent;
+        } locl;
+        """)
+
+    lines = stripped(fea)
+    assert lines.count("sub Scedilla by Scommaaccent;") == 2
+    assert lines.count("language ROM;") == 1
+    assert lines.count("language MOL;") == 1
```
khaledhosny: Test the whole output against an expectation. Counting matching lines does not guarantee the output is as expected.

tomekthewo: Good point, done in b354bdb8 — every case now compares the full expanded text, which doubles as documentation of what the expansion produces. It caught nothing new, but it is a much better regression net: the counting assertions would not have noticed the order of the emitted statements. The two round-trip tests assert the whole feature text as well.

anthrotype:
So.. I’m not convinced a physical line is the right unit for this transformation. FEA defines line endings as ordinary whitespace therefore valid statements may span lines or share one.

Unlike with the existing variable_features.py, expanding multiple languages cannot operate on locally delimited syntax because it needs to understand the type and scope of the following FEA statements, which means reconstructing statement boundaries, nesting, definitions and lookup references, all through a bunch of fragile regexes...

E.g. this would be valid Glyphs syntax that the current code leaves unexpanded because another statement follows on the same line:

```fea
language AZE CRT; sub i by idotaccent;
```

Or take this one:

```fea
language AZE CRT;
lookup idot { sub i by idotaccent; } idot;
sub Scedilla by Scommaaccent;
```

here the scanner can't detect a lookup's opening and closing braces that occur on the same line (brace depth is 0) and treats the lookup as continuing to the end of the language section, so `CRT` only receives `lookup idot;` while the inline sub rule gets silently ignored.

There are more of these all stemming from conflating line with statements. E.g. `@Ced = [Scedilla]; sub @Ced by Scommaaccent;` would be classified as one definition and the sub not replayed. Or line break after `lookup` or `markClass` would prevent the definition from being recognized.

Fixing these (or even just rejecting them with a loud error to avoid silent wrong output) risks turning this into a partial second FEA parser inside glyphsLib, needing knowledge of lookup, class/markClass semantics. Besides, this operates on incomplete feature text before includes gets resolved, so there's that too.

So I’m not yet convinced glyphsLib can transform the extension safely without duplicating parts of feaLib. And maybe feaLib would not necessarily be the wrong layer, despite not being official FEA... I honestly don't know, sorry for being inconclusive.

anthrotype:
(and then there is the question of also needing to implement this in fontc/fea-rs which just hit 1.0 this week.. at some point I would rather stop introducing new features in both compilers)

tomekthewo:
You're right, and the line-based scanning isn't salvageable. Thanks a lot for writing the counterexamples out — all four reproduced exactly as you described, including the two that fail silently.

I've rewritten the expansion on feaLib's own `Lexer` instead of on the text:

| input | before | now |
|---|---|---|
| `language AZE CRT; sub i by idotaccent;` | not expanded | expanded |
| `lookup idot { … } idot;` closed on one line | `CRT` lost the rule that followed | both statements replayed |
| `@Ced = [Scedilla]; sub @Ced by Scommaaccent;` | `CRT` got nothing | rule replayed, class defined once |
| line break after `lookup` | block duplicated → redefinition | replayed as `lookup idot;` |

The part I'd gently push back on is the "partial second FEA parser" worry — tokenizing is precisely what removes the need for one. `Lexer` already gives me statement boundaries (a `;` at brace depth zero, wherever it sits on the page), and it already tells a brace inside a comment or a string from a real one. What's left is classifying the head of each statement, and that grammar is five cases:

- `lookup NAME [useExtension] {` → definition: emitted once, replayed as `lookup NAME;`
- `@Class = …` → definition: emitted once
- `markClass …` → definition: emitted once
- `include(…)` → bail out (below)
- anything else, `lookup NAME;` references included → repeated verbatim

with the scope ending at the next `language`/`script` or an unmatched `}`. No glyph classes to resolve, no lookup bodies to understand, no name resolution. The emitted text is sliced out of the original by offset, so comments, indentation and blank lines survive untouched.

`TokenExpander` in `builder/tokens.py` already runs inside `_to_ufo_features` and rewrites Glyphs-only `$[…]` / `${…}` / `$name` syntax into plain FEA — with a hand-rolled predicate subparser and an `eval()` for the arithmetic. In my opinion, it is more Glyphs-specific machinery than the five heads above. I know those tokens are locally delimited and the shorthand isn't — that's a fair distinction. My point is only that lexing moves that line: on a token stream a statement boundary is as local as the `}` that closes a `${`.

On includes: agreed, they can't be replayed, so a multi-tag scope containing one is left alone entirely. The shorthand is invalid FEA, so feaLib then reports it — the failure stays loud instead of becoming a silently partial expansion.

**What this does not address is your second comment** about fontc parity, because it still fails unless the same expansion is implemented there too. The portability argument in my PR description ("the UFO gets plain FEA, so any compiler reads it") doesn't cover fontc, and I shouldn't have framed it that broadly. The parity cost you're describing is real and this patch doesn't remove it.

**One thing that might change the calculus: Glyphs 4 emits it from *automatic* feature generation. I took two of our Glyphs 3 sources (appVersion 3517 and 3526), opened and saved each in Glyphs 4 (appVersion 4004), and in both the automatic `locl` came back rewritten the same way:**

```fea
# before, appVersion 3517
language AZE;
lookup locl_latn_0 {
    sub i by idotaccent;
} locl_latn_0;
language CRT;
lookup locl_latn_0;
language KAZ;
lookup locl_latn_0;
...

# after, appVersion 4004
language AZE CRT KAZ TAT TRK;
lookup idotaccent {
    sub i by idotaccent;
} idotaccent;
```

No edits by the designer, the feature still flagged `automatic` before and after, and both fonts went from ten single-tag statements to two single plus the same two shorthand ones (`AZE CRT KAZ TAT TRK` and `ROM MOL`). So this isn't an opt-in extension that some sources happen to use — it's ordinary Glyphs 4 output, and a source that built yesterday stops building after a save, in feaLib and fea-rs alike. Two fonts against one Glyphs 4 build is still not a survey, but it reproduces rather than looking like a fluke.

As I read it, there are three ways this could go:

1. **Token-based expansion** as described — I can push it to this branch.
2. **Detect and reject, no expansion.** A `language` statement with more than one tag becomes a clear error instead of `Expected ';'` pointing at column 14. Needs no statement semantics at all, needs no fontc counterpart, and is strictly better than today.
3. **Drop it** — it belongs in feaLib/fea-rs, or nowhere.

I can only guess which one you'd prefer. But if Glyphs emits the shorthand on save, everything that reads `.glyphs` sources ends up having to solve it, and for the UFO path this is the one place to do it once — fontc would still need its own, which is the cost you flagged, though I don't think that argues for nobody doing it.

Happy to push (1) here, or cut it back to (2) — whichever you'd merge. For what it's worth on the approach itself: I ran the tokenizer over ~1100 feature code blocks from 35 production `.glyphs` files and all of them lex cleanly on the raw pre-include text, which is what convinced me it holds up outside the test suite.

anthrotype:
> Happy to push (1) here

thanks, since you have done it already and sounds like it's better than the line-base scanner, you may as well push it

> Glyphs 4 emits it from automatic feature generation

oh, great. That means we can't dismiss this and will have to support it in fontc anyway (https://github.com/googlefonts/fontc/issues/2126). And maybe in feaLib as well, which would make the current PR redundant.

schriftgestalt:
We documented our fea additions. There is a branch for each topic: https://github.com/schriftgestalt/feature_file_workshops (This is not meant as a request to add it to the spec as is, just to have written it down properly.)

anthrotype:
Thanks Georg, I briefly checked Adobe's upstream repo the other day but couldn't find your multi-lang extension proposal. I found it now here: https://github.com/adobe-type-tools/feature_file_workshops/pull/8

tomekthewo:
Pushed. The line scanner is gone — `multi_language.py` now walks feaLib's own `Lexer`, so a statement ends at the first `;` at brace depth zero wherever it sits on the page, and a brace inside a comment or a string is never code. Your four counterexamples are in the tests under their own names:

- `test_statement_sharing_the_line_with_the_shorthand_is_expanded`
- `test_lookup_closed_on_one_line_does_not_swallow_what_follows`
- `test_definition_sharing_a_line_with_a_rule_replays_the_rule`
- `test_line_break_after_lookup_keyword_is_still_a_definition`

Georg — thanks, adobe-type-tools/feature_file_workshops#8 is exactly the grammar this implements (`language <tag>+ [exclude_dflt|include_dflt] [required];`, following rules registered under every listed tag for the current script), so nothing to change here. And if the conclusion is that this belongs in feaLib rather than at the UFO boundary, that makes sense to me — expanding here is only worth it for as long as the compilers don't.

anthrotype:
Thanks.

> https://github.com/adobe-type-tools/feature_file_workshops/pull/8 is exactly the grammar this implements

What does Glyphs 4 do with `dflt` inside a multi-tag language statement? The proposal says it must be specified alone (https://github.com/schriftgestalt/feature_file_workshops/blob/2d1d7a058e2642a9bf6ef0f6dd8594ea7d7d40e8/OpenTypeFeatureFileSpecification.md?plain=1#L1219), and Glyphs 3.4.1 refuses to export it.

E.g. with the following feature prefix:

```fea
languagesystem DFLT dflt;
languagesystem latn dflt;
languagesystem latn AZE;
```

and this `locl` feature:

```fea
script latn;
language dflt AZE;
sub i by idotaccent;
```

Glyphs 3.4.1 says `Error: Regular: Expected ";" after language statement, found in feature "locl", line 2`.

Your PR expands it anyway, to `language dflt;` + the original rules and then `language AZE;` + the replayed rules.
But since `language AZE;` implies `include_dflt`, AZE gets the substitution twice, once inherited from dflt and once replayed, as two distinct lookups with identical content both registered under AZE. This is harmless for shaping but superfluous.

Could you check what Glyphs 4 does with it? If it rejects it too, I'd leave such a statement untouched and let feaLib report it, like the other shapes you don't classify. If it accepts it, knowing what lands in the GSUB would settle what the expansion should emit.

tomekthewo:
Glyphs 4.0.1 (4004) rejects it too, with the same error as 3.4.1 — this is the compiler in the Features tab:

```fea
script latn;
language dflt AZE;
sub i by idotaccent;
```

→ `Expected ";" after language statement` on line 2, `locl` marked as failing.

So a multi-tag statement listing `dflt` is now left untouched, with a warning saying why, and feaLib reports it as before. Only `dflt` is guarded, not `DFLT` — expanding a `DFLT` statement still fails loudly in feaLib, while expanding `dflt` compiled and produced the redundant duplicate you described.

Looking for the same class of problem elsewhere turned up three more — a lone `\r` (feaLib ends a line there too, so the offsets went out of step and the replay silently lost its rules), a missing semicolon, and a shorthand outside any block — all now left alone, with two invariants over a corpus of awkward sources in place of one example each. Three existing tests wrap their snippet in a `feature` block now, since the entry point only ever sees the full assembled text, where a top-level `language` is invalid.

anthrotype:
thanks for the follow up, sorry for the late reply.

I'm inclined to merge this as a stopgap even though it still belongs in feaLib and fea-rs (https://github.com/googlefonts/fontc/issues/2126). Most likely I won't get to that soon, and given that Glyphs 4 emits the shorthand from automatic features, waiting means sources stop building. Two small things first.

The `\blanguage\b` guard matches nearly every font, so the lexer runs over code that may not hold the multi-language shorthand at all. We could avoid that by matching that specific shape instead, and returning early when it doesn't:

```python
# whitespace or a # comment running to the end of the line, repeated
_separator = r"(?:\s|#[^\r\n]*)+"
# `language` + separator + tag-like + another separator + one more tag-like character
_shorthand_re = re.compile(
    r"\blanguage\b" + _separator + r"[^;\s#]+" + _separator + r"[^;\s#]"
)
```

A `#` comment separates like whitespace, so a shorthand still matches when the comment contains a semicolon.

```fea
language #;
AZE CRT;
```

It is crude on purpose: `language TRK exclude_dflt;` matches and still gets lexed, which only costs the lexing.

And `_tokenize` should catch `ValueError` and `TypeError` as well as `FeatureLibError`, and return None. The lexer raises those on a bad literal. On text that the guard above let through, those currently abort the build instead of leaving the text alone for feaLib to report.

A shorthand inside an included file stays a known gap, and I'm fine with it for now. This runs before includes are resolved, so it never sees one, and fontmake still fails with Expected ';' where Glyphs compiles it.

schriftgestalt:
https://github.com/schriftgestalt/fonttools/tree/Multi-Languages

I’ll send a PR if you like.

schriftgestalt:
https://github.com/googlefonts/fontc/pull/2169

https://github.com/fonttools/fonttools/pull/4201

anthrotype:
Thank you Georg, I'll take a look today :)

If we then go the fea-rs/feaLib route, then this PR will probably be no longer needed after all

anthrotype:
With @schriftgestalt's fonttools PR https://github.com/fonttools/fonttools/pull/4201 merged, glyphsLib can simply bump its minimum required fonttools to whatever the next release will be and pass these through unchanged. Thanks @tomekthewo

schriftgestalt:
> script latn;
> language dflt AZE;
> sub i by idotaccent;
> → Expected ";" after language statement on line 2, locl marked as failing.

Is that needed? Removing the "AZE" has will apply it to "AZE" anyway?

anthrotype:
where is that error message from?

> Is that needed?

what is not needed?

> Removing the "AZE" has will apply it to "AZE" anyway?

I don't understand what you mean

anthrotype:
oh! I think you mean 

> "Why list AZE next to dflt at all? If you remove AZE, the rule will still apply to AZE anyway."

you're basically saying listing `dflt` with other tags is pointless (which would be an argument for rejecting it, not for allowing it, see https://github.com/googlefonts/fontc/pull/2169#discussion_r4081908422)

But actually AZE is not redundant. With just `language dflt;` in the feature, the compiler registers the rule under latn/dflt only, and `AZE` gets no entry of its own. Then at runtime:
- If the font has no AZE entry at all, the shaper falls back to latn/dflt, so AZE text does get the rule, as you say.
- If however AZE has its own entry, because any feature has AZE-specific rules (which is normal in a font that bothers with Turkic `locl`), the shaper uses that entry and never consults `dflt`. AZE then only gets the rule if it's registered under AZE explicitly.

schriftgestalt:
I tried this: 
```
script latn;

language dflt;
lookup idotaccent {
	sub i by idotaccent;
} idotaccent;

language AZE;
lookup n {
	sub n by m;
} n;

```
and "i" is always replaced and "n" only for AZE.

anthrotype:
Right, in your example `language AZE;` comes after the dflt lookup, so AZE inherits it (that's the implicit `include_dflt`). It only misses it when AZE has its own LangSys but `locl` never mentions AZE. E.g.

```fea
languagesystem DFLT dflt;
languagesystem latn dflt;
languagesystem latn AZE;

feature locl {
    script latn;
    language dflt;
    sub i by idotaccent;
} locl;

feature ccmp {
    sub a by b;
} ccmp;
```

Here AZE gets a LangSys with only `ccmp`, and the shaper never falls back to dflt for locl.

Anyway, this is moot now that you made fea-rs reject dflt in a group, which matches Glyphs and the proposal.

Issue #1082, opened 2025-05-22
Not all tokens are supported


RosaWagner:
I can export my font pretty well with almost all tokens, but one that doesn't work though: `$[category like "Letter" && case==lower && unicode like "*"]`

The problem seems to be the `unicode` one. I tried with `countOfUnicodes == 1` but doesn't work either.

You can find the issue in there: https://github.com/RosaWagner/Borel, first line of the `calt` feature.

khaledhosny:
Do you have the category saved in the glyphs file? IIRC, properties that are taken from `GlyphData.xml` did not work because the code that does token expansion does not currently use GlyphData.

RosaWagner:
I am sorry Khaled I don't understand what you mean 🥲 You mean category as in `category like "letter"` ? How do I save it in the glyphs file?

anthrotype:
command-option-I to open the glyph info panel

RosaWagner:
Yes that I know, but you mean I need to click the thing, and then it is "saved"?

from
<img width="374" alt="Image" src="https://github.com/user-attachments/assets/732c3407-b221-4feb-a5f0-5c559a649b2a" />
to
<img width="373" alt="Image" src="https://github.com/user-attachments/assets/20c40844-4e4e-45a8-9f5a-cf356b6810ec" />

anthrotype:
(but ideally you shouldn't need to do that, the code should look the glyph up in the GlyphData)

RosaWagner:
still doesn't work :(

schriftgestalt:
The cleaner token would be `unicode != nil`, but I don’t know how well this is supported in GlyphsLib.

RosaWagner:
Thnaks @schriftgestalt unfortunately it doesn't work as well with glyphsLib

anthrotype:
I think the problem is not `unicode` property but `case`, see this is commented out in the tokens_test.py:

https://github.com/googlefonts/glyphsLib/blob/bd8276222afad582c030f839d90271bc0ee70d0b/tests/tokens_test.py#L97-L98

anthrotype:
ok it's both category=Letter and case=lower that need to be set directly in the .glyphs source for all the lowercase glyphs with a unicode codepoint, if you want glyphsLib to select them with the tokens

... until this code is updated to support looking up GlyphData.xml

RosaWagner:
It works :) thanks. So it was all of them except the one I thought haha

[...]

PR #1167, opened 2026-09-16
Derive script from Unicode codepoints when not explicitly set

bgermann:
When the script property is not explicitly set on a glyph in Glyphs.app, the application automatically derives the script from the Unicode codepoint. This commit implements the same behavior in glyphsLib by:

1. Adding a helper function _script_from_unicode() that maps Unicode codepoints to Glyphs.app script names (hangul, kana, han, latin, cyrillic, arabic, hebrew, thai, devanagari) based on Unicode ranges.

2. Converting the script attribute on GSGlyph from a simple stored value to a property that:
   - Returns the explicitly set value if one exists
   - Falls back to deriving the script from the first Unicode codepoint
   - Returns None if no script can be determined

This ensures that glyphsLib matches Glyphs.app behavior when serializing and processing fonts where glyphs have Unicode values but no explicit script assignment, which is common in many font projects.

Addresses #1082 partly.


schriftgestalt:
Glyphs uses only the GlyphData.xml file to assigning scripts, except for CJK scripts, those are hard coded by ranges, similar to how it is handled  here.

